PR #4, opened 2015-03-26
Enabled TextField.Hint for Combobox

stefan-schweiger:
Your implementation wasn't working, so I fixed it and also enabled it for the non-editable ComboBox


ButchersBoy:
Cool, thx, I'll try and merge over the weekend.

James

ButchersBoy:
thanks for that, special kudos for the first PR on the repo!

FYI, I will move TextField.Hint to become an object in the future, and provide TextField.HintTemplate, so we can have more advanced hints.

PR #13, opened 2015-04-17
update nuget packages, it's simple, i know ;-)

punker76:
but now with the latest MahApps.Metro and the glow border...

2015-04-17_21h39_28


ButchersBoy:
:-)

PR #23, opened 2015-05-27
Added Hint Opacity Dependency Property to TextField

Falgantil:
Added a Hint Opacity Dependency Property to TextField which allows specifying of opacity on the Hint text.
Standard value is still 0.23 as usual.

The Fruit Combobox is using 0.5 hint opacity to demonstrate the property.


ButchersBoy:
Cool, thanks!  

FYI, at some point I might make a .HintContent, for more complex hints, which can be styled.

PR #51, opened 2015-09-01
Fix 'TemplateRoot' name cannot be found in the name scope of 'System.Windows.Controls.ControlTemplate'

punker76:
If we use the ProgressBar (style) in a TabControl we can get a nasty exception like this:

```'TemplateRoot' name cannot be found in the name scope of 'System.Windows.Controls.ControlTemplate'```

This can happen if the Loaded event comes before ApplyTemplate, so the Storyboard can't find the TargetName.

Found this while playing with OnePlus-One and this issue https://github.com/Droidkit/OnePlus-One/issues/2


ButchersBoy:
Thanks @punker76 ...was literally just about to start looking at this, you saved me an hour or 2.  Thanks.

punker76:
@ButchersBoy an hour... :cake:

itechy21:
woop :)

PR #60, opened 2015-09-12
custom controls for an app bar and a rating bar

spiegelp:
Hi, 

I saw your great port of the Material Design for C# and WPF last weekend. However, I missed an easily reusable control for the app bar and a rating bar in the demo project. I will need them for a project of mine. Therefore I implemented them on my own and ask you to pull them into your original repository. 

Thanks


ButchersBoy:
Hi there...the AppBar looks great, but I don't really want to go down the route of having UserControls in the library.  My feeling is that UserControls typically sit at the application level, whereas this library is a bit more low level than that.  Having said that the XAML and code is clean, and I think you've created a great little example of the library in use.

Also am currently considering how I am going to add a "Drawer" control as this may have some relation with a possible dialogs implementation, as they both grey out underlying content.  Again this could affect the AppBar.

ButchersBoy:
Regarding the RatingBar, I have a few suggestions:

* Turn it into a custom control (derive from Control) with a default theme defined in Generic.xaml. This would bring it inline with the rest of the toolkit.
* Make the number of stars/values a configurable int.  5 may be the most popular number, but we shouldn't limit it, other people will have other requirements.  Then, we might just populate an ItemsControl with sequential numbers, leading us onto:
* Instead of having an Icon or PathData property, expose a RatingItemTemplate (DataTemplate) which would be much more "WPF-ish" and allow the user to use anything for each rating item.  We could set a default value in the Style in generic.xaml to https://materialdesignicons.com/icon/star-outline

If you want any clarification on this give me a shout.

Thanks,

James

spiegelp:
Hi,

the dynamic and more customizable version of the RatingBar is working. It is a strange feeling to write an user interface (partly) in code, if you are used to write XAML and XHTML for years. 

What is about the AppBar and the Drawer? I was already thinking about a Drawer implementation and possible integrations one week ago. Should I remove the AppBar from the branch or are you planning any integrations or migrations? 

Philipp

ButchersBoy:
Hi, yes...can you remove the AppBar please, we already have ToolBar themes, and the ColorZone which are flexible enough to build any menu/app bar the user desires.

ButchersBoy:
I've made a few changes for the final version:

* Used an ItemsControl as the template, this means we can rely on standard WPD layout mechanisms rather that generating Grid layout in the Control code.
* Renamed primary properties to Min/Max/Value to bring in line with ProgressBar
* Exposed templates and templates selector (and orientation) so it's more of a standard WPF mechanism to allow users to customise the control.
* Exposes a routed command so users can really override everything should they desire

Doing the above makes this a more standard WPF control and allows better customisation in the XAML (instead of setting opacity in code for instance), allows us to run animations easier, and also exposes the actual value in each item (which can be seen in the demo).

spiegelp:
I had a look on your changes of the rating bar and I noticed that the ripple effect is cut off by the control's edges. Seems to be an issue with the position of the buttons inside the control. The grid in the previous version filled the control and aligned the buttons in the center of the cells. The paddings might avoided the ripple to be cut off. 

A funny rotation animation on the click event. Remembering the theory back in university, it might distract the user from the content. It's not such a big deal for a simple rating bar. I think that the users won't forget that they wanted to set a specific value. In my opinion, it could be a little bit more decent for the item. However that's just a subjective perception of mine. 

Nevertheless the RatingBar have had a great evolution since the last changes.

ButchersBoy:
I'm actually gonna change the animation, but the effect I want is a bit more complicated so will come back to it.

PR #64, opened 2015-09-24
Add button shadow hover effect

l1pton17:
Looks like button raises up on mouse over.

I don't know should i create a new pull request for each contribution =)
So, there is some improvement for textbox validationerror:
1) It shows only if textbox is in focus (case if show all errors together that they can be mixed with themselves)
2) Use popup for showing error hence it popup over the window

And bug fix: now can select text under a floating hint in textbox

### DIFF
--- a/MaterialDesignThemes.Wpf/CustomValidationiPopup.cs
+++ b/MaterialDesignThemes.Wpf/CustomValidationiPopup.cs
@@ -1,0 +1,120 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows;
+using System.Windows.Controls;
+using System.Windows.Controls.Primitives;
+using System.Windows.Input;
+
+namespace MaterialDesignThemes.Wpf
+{
+    /// <summary>
+    /// MahApp's validation popup copy
+    /// </summary>
+    public sealed class CustomValidationPopup : Popup
+    {
+        private Window _hostWindow;
+
+        public CustomValidationPopup()
+        {
+            this.Loaded += this.CustomValidationPopup_Loaded;
+            this.Opened += this.CustomValidationPopup_Opened;
+        }
+
+        protected override void OnPreviewMouseLeftButtonDown(MouseButtonEventArgs e)
+        {
+            this.IsOpen = false;
+        }
+
+        private void CustomValidationPopup_Loaded(object sender, RoutedEventArgs e)
+        {
+            var target = this.PlacementTarget as FrameworkElement;
+            if (target == null)
+            {
+                return;
+            }
+
+            this._hostWindow = Window.GetWindow(target);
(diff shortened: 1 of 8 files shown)


oniony:
Hmm, Google Material Design spec doesn't show buttons doing this: they merely (and quite subtly too) change colour. Not sure deviating from spec would be a good idea, yo.

l1pton17:
I understand, but material design originally was design for mobile platforms where doesn't exist any hover effect. In case of web, some materialdesign sites introduce a hover effect for theirs controls: http://diskforyou.com/ http://materializecss.com/buttons.html and on.
And i use materialdesign for one internal project at my work, and some users asks me to add an hover behavior, because it confused when u don't see any reactions from UI when cursor over a button.

oniony:
Well, there's already the subtle background colour change I mentioned but I can see on Google's own sites the mouse cursor changes to a hand too (like it has traditionally done for links in a web browser). Not sure if that is part of the spec but it certainly seems to be present in all of the examples I can find (e.g. even in the desktop Google Keep).

ButchersBoy:
Hi...this is a good discussion...I appreciate the fact there isn't any sort of hover effect right now...and also, any kind of guideline from google is coming late to the party as the design language was mobile first.

I think wherever possible we have to work to the Google spec...I have already had people rightly point out where i have deviated, and in those cases I've tried to get back on track.

I think we have to primarily follow two guidelines:

* The official spec (which currently says little on this subject) https://www.google.com/design/spec/material-design/introduction.html
* Material Design Lite (Googles own web implementation) http://www.getmdl.io/

Currently, for both raised and flat buttons MDLite is using a hand cursor:

http://www.getmdl.io/components/index.html#buttons-section

It's not amazing, but if we stick to this we can evolve in line with Google in the future.

Your shadow does look nice, but I think for v1.1 we'll get the hand cursor in and the background hover for the flat buttons (which is also on the MDLite page).

l1pton17:
By the way, Material Design Lite buttons click effect uses shadows. Thanks for answering, i will be waiting for a new version

ButchersBoy:
@l1pton17 interesting point...i hadn't noticed that...only seems to kick in if you hold the mouse...

QuantumDeveloper:
@l1pton17 its pity that you closed your PR. I`d like to see this effect in material design and I agree that google gideline is not 100% correct for desktop apps, especially WPF.

l1pton17:
I don't know should i create a new pull request for each contribution =)
So, i have commit some improvement for textbox validationerror:
1) It shows only if textbox is in focus (case if show all errors together that they can be mixed with themselves)
2) Use popup for showing error hence it popup over the window

And bug fix: now can select text under a floating hint in textbox

ButchersBoy:
hey @l1pton17 , yes please do the validation stuff in a separate PR, easier for me to untangle from the button stuff.  Also, I'd like to propose some changes, it'll be easier for us to work through them on a cleaner branch.

PR #65, opened 2015-09-25
Popup Validationerrortemplate and ShowOnFocus

l1pton17:
Added validationerrortemplate (a copy from MahApps), that can popup over the
window.
Also, added ValidationAssist.ShowOnFocus attached property that allows ValidationError is showed only when textbox in focus (in case when exists a few textbox in error state, they show validation errors that mixed with themselves). 

Before:
before
After:
after

### DIFF
--- a/MaterialDesignThemes.Wpf/Converters/BooleanToVisibilityConverter.cs
+++ b/MaterialDesignThemes.Wpf/Converters/BooleanToVisibilityConverter.cs
@@ -1,0 +1,47 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows;
+using System.Windows.Data;
+
+namespace MaterialDesignThemes.Wpf.Converters
+{
+    public class BooleanToVisibilityConverter : IValueConverter
+    {
+        public Visibility TrueValue { get; set; }
+        public Visibility FalseValue { get; set; }
+
+        public object Convert(object value, Type targetType,
+            object parameter, CultureInfo culture)
+        {
+            if (value is Boolean)
+                return (Boolean)value ? TrueValue : FalseValue;
+
+            if (parameter is Boolean)
+                return (Boolean)parameter ? TrueValue : FalseValue;
+
+            return null;
+        }
+
+        public object ConvertBack(object value, Type targetType,
+            object parameter, CultureInfo culture)
+        {
+            if (Equals(value, TrueValue))
+                return true;
+
+            if (Equals(value, FalseValue))
+                return false;
+
+            if (Equals(parameter, TrueValue))
+                return true;
+
+            if (Equals(parameter, FalseValue))
+                return false;
+
+            return null;
+        }
+    }
+}
--- a/MaterialDesignThemes.Wpf/CustomValidationPopup.cs
+++ b/MaterialDesignThemes.Wpf/CustomValidationPopup.cs
@@ -1,0 +1,120 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows;
+using System.Windows.Controls;
+using System.Windows.Controls.Primitives;
+using System.Windows.Input;
+
+namespace MaterialDesignThemes.Wpf
+{
+    /// <summary>
+    /// MahApp's validation popup copy
+    /// </summary>
+    public sealed class CustomValidationPopup : Popup
+    {
+        private Window _hostWindow;
+
+        public CustomValidationPopup()
+        {
+            this.Loaded += this.CustomValidationPopup_Loaded;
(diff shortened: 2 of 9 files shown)


QuantumDeveloper:
It will be very cool, if this popup could be dragged by mouse) Because often there is a situation when it overlap something important)

ButchersBoy:
OK, here's what I think.  The current non-pop-up template is by design.  And I like it, it encourages white space, which is all too often not there in classic desktop apps.  If anything, looking at the latest spec, in the current design the font is too small.

But I can see the need for your version.

I don't want to change the current design, as non-popups is more in line with the "spec", plus version 1.0 has effectively shipped using that standard.

So let's make your version an optional extra; probably create a new BasedOn TextBox style, or just include leaving the existing style in place.  

Secondly - and would like to hear your opinion on this - you will see there is a MaterialDesignThemes.MahApps project in the solution.  This is to be the home of Material Design styling for MahApps controls, for example https://github.com/ButchersBoy/MaterialDesignInXamlToolkit/issues/40 will live in there.

So we could put this in there.  I think this is my preference, as it means the core Themes.Wpf assembly is more pure Material Design, and then we have the cross over assembly with MahApps.  Also this means we don't duplicate code.  Thoughts?

Thirdly (minor), I've now implemented .Named. swatches in the Colors assembly, so I want to start migrating specific colours (such as validation red) to reference this instead of specifying the hex in the Themes assembly.

ButchersBoy:
@QuantumDeveloper ...dragable tooltips...I'm wondering if this is something suited to Dragablz???

ButchersBoy:
(I think we'll need to update the nuspec for v1 to reference v1.1 of Colors assembly to get the names swatches)

QuantumDeveloper:
@ButchersBoy I think its not dragablz specific, but just separate control that could be dragged. I faced with such need a lot of times earlier.

l1pton17:
I thought that MateriaDesignThemes.MahApps is obsolete because actually it is empty =)
But it is a not total copy from MahApps, actually it is just copy of popup which moves with window. Template of validation is different.
It will be ok, if i will create a second property of VissialAssist like "UsePopup" which allows to swatch between using popup or not in ValidationErrorTemplate?

ButchersBoy:
It's empty because it is the beginning :)

Yes, ```ValidationAssist.UsePopup``` is good.

l1pton17:
Done

ButchersBoy:
Cool...I'll try and bring them in over the weekend...

ButchersBoy:
Hi, looking better, but:

* Can we rename ShowOnFocus to OnlyShowOnFocus? I believe this better represents the fact that you must explicitly set to true to change the default behaviour
* Does ShowOnFocus actually work?  I tried the following on the test project and couldn't see any different behaviour between True/False:

```
<TextBox Grid.Row="0" Grid.Column="1" 
				 x:Name="NameTextBox"
				 wpf:TextFieldAssist.Hint="Name"
                 wpf:ValidationAssist.UsePopup="True"
                 wpf:ValidationAssist.ShowOnFocus="True">
```

l1pton17:
Hi, i will rename. I took a look, there is bug, too many datatriggers makes me crazy =) Just a sec, i will fix it

l1pton17:
Done, and also I have added a slide animation for popup, looks pretty nice.

ButchersBoy:
OK...I have done a little bit of tidying up; there was a bit of dead code and i don't want any dead code on this project.

I have checked my changes in here: 

https://github.com/ButchersBoy/MaterialDesignInXamlToolkit/tree/l1pton17-ValidationErrorTemplate

But now, in the demo Visual Studio isn't happy, and I'd like to find out why:

image

ButchersBoy:
image

Dunno if it's getting confused all around those triggers, but I'd rather not be in a state where people are panicing and reporting errors, when in fact there are none...

l1pton17:
This is strange. I didn't get any warning from xaml in demo project.
everythinsisok

Maybe you should clean and rebuild solution again? Looks like xaml can't find a MaterialDesignValidationErrorTemplate template.
What a build option do you use? I mean debug/release, x64/x86/Any CPU.

QuantumDeveloper:
Looks like VS designer issue.

Davipb:
Can confirm it's just a Visual Studio issue; This happens a lot to me, especially when editing the app.xaml. Closing and re-opening the solution or sometimes just building/debbuging it fixes the problem.

ButchersBoy:
Done, thanks.  

Also, please don't think I'm being pedantic with my feedback...I just want **top** quality for the project :smile:

QuantumDeveloper:
We too :)

ButchersBoy:
Another thought...without looking at code, does OnlyShowOnFocus only apply to popup?  If so we should rename, OnlyShowPopupOnFocus...or preferable, apply the behaviour to the default template.

Otherwise the API is confusing.

l1pton17:
OnlyShowOnFocus applies regardless using popup or not.

ButchersBoy:
Perfect :)

PR #66, opened 2015-09-25
Very small fix of floating hint

l1pton17:
Small fix of selecting text under the floating hint in textbox.

Description of problem:
When you double clicked (or trying to select with mouse) a text under
hint in textbox, nothing happens.


ButchersBoy:
thanks

PR #67, opened 2015-09-25
Button hover effect

l1pton17:
It's clear pull request only with hover effect for buttons.


ButchersBoy:
OK, I've relented, but I've done darkened using a different technique which doesn't involve the extra shadow, and the colouring is more subtle.  I also added hand cursor, and a mouse over to the background of the flat button.

Found this little nugget in the specs:

"On desktop, this elevation may be reserved for hover interactions."

I'm taking "_May_" literally.  So we darken slightly on raised buttons, but the flat button I choose to may _not_ raise on mouse over ;)

PR #69, opened 2015-09-26
Add filling effect for ripple

l1pton17:
It works with a little hack.
Since previewmouseleftbuttonup don't want
to firing so i had to add an IsActive property. It property is set to true on IsPressed trigger.
So if hold button the ripple stands on.
ezgif com-crop

### DIFF
--- a/MainDemo.Wpf/Cards.xaml
+++ b/MainDemo.Wpf/Cards.xaml
@@ -39,7 +39,8 @@
                     <TextBlock TextWrapping="Wrap" VerticalAlignment="Center">A great way to keep fit and forget about the constant grind of IT.</TextBlock>
                 </StackPanel>
                 <StackPanel HorizontalAlignment="Right" Grid.Row="2" Orientation="Horizontal" Margin="8">
-                    <Button Style="{StaticResource MaterialDesignToolButton}" Width="30" Padding="2 0 2 0">
+                    <Button Style="{StaticResource MaterialDesignToolButton}" Width="30" Padding="2 0 2 0"
+                            wpf:RippleAssist.StayOnCenter="True">
                         <Viewbox Width="16" Height="16">
                             <Canvas Width="24" Height="24">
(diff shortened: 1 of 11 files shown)


l1pton17:
ezgif com-crop 3
ezgif com-crop 2

ButchersBoy:
Hey, this looks great...

* Are we 100% sure we can't get the IsActive/IsPressed visual state triggered inside the code?  I don't like the word "Hack" :smile:, and it's a shame the XAML has to call back to the ripple...
* Might be worth slowing down the IsPressed even more?  On my Android phone the Google apps take up to 2 seconds to ripple when you hold.
* Toolbar looks better
* I don't like the switch...it seems odd that the ripple gets left behind.  It's better if it moves with the knob (like the range), and I've seen animations to this effect.  It kind of illustrates it this pic https://www.google.com/design/spec/components/selection-controls.html#selection-controls-switch

l1pton17:
1) I am researching this question but probably there is no better solution, because button hides previewmouseleftbuttonup (and any mouseup) event, even if you subscribe on this event with AddHandler method (which allow subscribe to handled events).
2) What you wish, i use visualstatemanager from your last ripple update, so you or i can change it easy.
4) I have took idea for the switch on http://www.getmdl.io/components/index.html#toggles-section
Your idea is good, i will try to make it, but right now ripple doesn't assume this behaviour =)

ButchersBoy:
4)  maybe not even use the ripple for the toggle?  Just a manual ellipse, like what is on the slider.  It would probably be an easier solution.

l1pton17:
4) Okay, will make it tonight

l1pton17:
Check out ;)
ezgif com-crop 4

l1pton17:
Woohoo! I have been researching 3 days, but now i have made it!
1) It works without any "hack". I have finally found a solution:
Now in OnApplyTemplate ripple checks templatedparent and if it is ButtonBase, bind IsPressed to IsActive. So now we don't need to subscribe on IsPressed trigger in XAML!

2) Add RippleAssist.StayOnCenter and RippleAssist.RippleSizeMultiplier attached properties.
StayOnCenter means two things:
1. Ripple will rise up from center of element
2. If ripple content is an element which has a size, ripple will cover only this element, not whole button.
StayOnCenter is good for buttons with flat style, that have an image as a content.
I thing this behavour is heavy, so maybe it can be splited into 2 properties.

RippleSizeMultiplier:
When ripple rises up, it makes size of ellipse is enough to cover whole button from any place on it, so the correct formula for ellipse size is: 2 * Math.Sqrt(Math.Pow(width, 2) + Math.Pow(height, 2))
But for flatbuttons, who doesn't clip bound and doesn't have any border, this formula give a very large ellipse rather they need. So this property can be used. And it supposed that a customer will decide, need he smaller ripple or not, so i have made it as an attached property.

ButchersBoy:
Double a couple of little tweaks; managed to get it working without IsActive using ```EventManager.RegisterClassHandler```, also I renamed the centre attached prop, and slowed down the anim.

Thanks, for this, much appreciated.

l1pton17:
Thanks :)
A feq questions:
1 If i hold button and then move mouse out of button, ripple is still active. But really button is not in active state and if i release the mouse nothing is happens (button will not be clicked).
2 What do you think about animations on http://www.getmdl.io/components/index.html#buttons-section site? I suppose that site is more "desktop materialdesign" than mobile phones, what is your opinion?

ButchersBoy:
1) good spot. Think I can fix that though.
2) well, I take my inspiration from that site, the actual spec website, and my phone.  And they are all slightly different... One thing though, the slow ripples fill the vertical axis first, then slowly fill out the horizontal. Which is nice.

PR #71, opened 2015-09-27
template for Expander

spiegelp:
Just an adaption of the ListBox template to lay out the items in a grid. The card version is not adapted because cards are a DON'T in grids according to the official specs.


ButchersBoy:
Gonna come back to this because a) we haven't got a proper actual ListView.GridView yet, so there might be some naming conflict and I don't want to have to rename later, b) quickly looking at the markup there might be some duplication we can remove.

spiegelp:
I just added basic templates for the Expander and the TabControl.

ButchersBoy:
@spiegelp thanks for your contributions.

* I purposefully don't want a TabControl template.  I want Dragablz to be the de-facto TabControl, and don't want the extra code to have to support; Dragablz is already far from complete, and taking up a lot of my time.  It does look nice but I know from experience there will soon be requests for vertical tabs, etc and I can't have the duplication.

* Expander: can we tidy this up a little?  Bring all the spacing in-line with the TreeView and use the same animations?  Then we will have some nice consistency.  If you can do this and put the Expander in a separate PR, I'll bring it in.

Thanks.

ButchersBoy:
Thank you, sorry for the delay on this one.

PR #75, opened 2015-09-28
Animate underline in textbox

l1pton17:
Add underline control (i will be agree with renaming of this control) that makes expanding animation for underline in textbox and on like on http://www.getmdl.io/components/index.html#textfields-section

ezgif com-crop 1


ButchersBoy:
.gif looks awesome.  I will code review when I get a chance.

l1pton17:
Okay, it seems i run into with faulty revert merge with Ripple filling effect pull request.
So, if you want, i can make another pull request without this revert merge.

PR #78, opened 2015-09-30
Fix #74

### LINKED ISSUE #74: Clock/TimePicker jumps from PM to AM when dragging hour backward
...should be a pretty easy fix...

Davipb:
A very simple fix for #74


Davipb:
So, the tests failed because of missing nuget packages, but all I changed was a single line of a completly unrelated file...

ButchersBoy:
Cool, thanks! Probably just Appveyor dicking around.

PR #79, opened 2015-09-30
Animate Underline

l1pton17:
Sorry for twice pull request, previous was faulty merge.
All the same, this pull request adding underline animation into all textbox based controls.


ButchersBoy:
nice work :)

PR #81, opened 2015-10-03
Moprhe checkbox

l1pton17:
I have seen this checkbox in some material design frameworks and i have found it is interesting.
It is additional style of checkbox named MaterialDesignMorpheCheckBox, it doesn't alter the existing style, just adds a new.

### DIFF
--- a/MainDemo.Wpf/Buttons.xaml
+++ b/MainDemo.Wpf/Buttons.xaml
@@ -111,8 +111,8 @@
             <RadioButton Style="{StaticResource MaterialDesignRadioButton}" Margin="0 0 8 8" VerticalAlignment="Center">
                 Ga Ga
             </RadioButton>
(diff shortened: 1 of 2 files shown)


Davipb:
Image is giving a 404 error here, so in case anyone has the same problem as me, here's what the checkbox looks like:

ButchersBoy:
I'm not sold on this one...it's a deviation, and I don't necessarily want every single possible animation in the lib, trying to stick - mostly - to the standards...

SuicSoft:
@l1pton17 I just created a repository for material design animation for WPF. There I will have some animations , animated styles (Moprhe checkbox) and more!

ButchersBoy:
@SuicSoft will you be using the standard .Colors.dll colors and brush definitions (what we declare in app.xaml)?  Right from the beginning I extracted the .Colors out in the hope that other people might adopt them...it would be cool if a mini-ecosystem of compatible stuff was to evolve :smile_cat:

SuicSoft:
@ButchersBoy It will use what's in App.xaml

PR #83, opened 2015-10-04
Fix compilation errors with App.xaml

jamesqo:
New user here. I copy-pasted the code from `App.xaml` like the README said, and while the first 2 lines worked fine, I got compiler errors on the others. Decompiling the assemblies with ILSpy shows that there are no `Recommended`, `Accent`, or `Primary` directories in `MaterialDesignColors`:

xamlbug

So now I removed the extra directories and everything's good.

This should save other users a lot of headache.


ButchersBoy:
Hi, sorry, there's been some work on the Colors this week and you just caught it in a state of flux...the docs have got a little ahead of the actual Nuget package for the Colors assembly (which is published manually as  it rarely changes).  I am just updating the Colors assembly now, and the .nuspec depedency for the Themes assembly.  Once this is done the App.xaml should work as per the latest docs.

ButchersBoy:
OK  if you update the Themes package to the latest CI build you'll get the latest Colors package, with the new resources.  I will update the WIKI to reflect that they are in effect pre-release instructions.

PR #84, opened 2015-10-04
24H timerpicker working

quicoli:
added Is24Hours property.
Usage example:

```
<wpf:TimePicker Is24Hours="True" Grid.Row="6" Grid.Column="1"
x:Name="PresetTimePicker24" VerticalAlignment="Top"  Width="100"
HorizontalAlignment="Left" Margin="0 8 0 8" />"
```

### DIFF
--- a/MainDemo.Wpf/MainWindow.xaml
+++ b/MainDemo.Wpf/MainWindow.xaml
@@ -11,7 +11,7 @@
         TextOptions.TextFormattingMode="Ideal" 
         TextOptions.TextRenderingMode="Auto"        
         Background="{DynamicResource MaterialDesignPaper}"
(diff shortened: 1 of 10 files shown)


ButchersBoy:
Cool...my intent is to bring this in after 1.1 release...I want to get 1.1 shipped first because of the fluctuation around the Colors .dll/docs, and my time is a bit constrained.

(looks like it has gone into merge conflict)

ButchersBoy:
Thanks for this, now pulled in.  I've made a few changes, but of a tidy up so see my check-in for feedback: https://github.com/ButchersBoy/MaterialDesignInXamlToolkit/commit/e068a760715f71dd6aafa05c593b150ae67f63af

I think we can still do a couple of little things to make more look like the Google calendars - primarily the inner buttons are slightly bolder, but will come back to this.

Thanks

PR #85, opened 2015-10-04
Fix blurry wpf effect issues

punker76:
Using an Effect on Border with inner content causes sometimes blurry text, so this pr will fix something of this.

- [x] ColorZone
- [x] Card
- [x] ComboBox
- [x] DataGrid ComboBox
- [x] DataGrid Popup Edit TextBox

@ButchersBoy i noticed, that the new Dialogz now a little bit better opens.


ButchersBoy:
:nail_care: polish!

PR #88, opened 2015-10-06
ListBox

l1pton17:
I have changed ListBoxItem template. Now it use ripple and it is transparent. Base color of transparency is depends on foreground, if it is dark, selection in listbox become dark, otherwise white. Hence now listbox can work with any background color. For example, i have changed maindemo navigation listbox, it's background is binded on primaryhuebrush.


ButchersBoy:
Thank you, been wanting to sort this out...looks good.

PR #89, opened 2015-10-08
Animate DatePicker

l1pton17:
Add animation in DatePicker on navigate.


ButchersBoy:
Great job, love the animations!  I was about to moan about the new layout of the top date then I saw they have updated the specs.

PR #92, opened 2015-10-10
Add InputGestureText usage

l1pton17:
Fix https://github.com/ButchersBoy/MaterialDesignInXamlToolkit/issues/91
I have added InputGestureText usage in menuitem. Also style was changed according to material design specs: https://www.google.com/design/spec/components/menus.html#menus-specs


ButchersBoy:
thanks

PR #93, opened 2015-10-10
Use Controlz.PopupEx for Dialogs

punker76:
So this is the next big cool thing...

Dialogs use now the `Controlz.PopupEx` to fix the OS topmost stuff, cause Dialogs should be topmost, but not OS topmost.

To use the custom Popup I use paket to get the control from the `Controlz` [github repository](https://github.com/punker76/Controlz). That's really cool, cause we have no extra dll dependency for material design.

2015-10-11_01h03_05


l1pton17:
Could you change validationerrortemplate to use this popup instead of customvalidationpopup, please?

punker76:
@l1pton17 yepp, want to change next...

punker76:
@l1pton17 done

PR #95, opened 2015-10-11
Fix Menu

l1pton17:
Sorry, i forgot to change some properties in accordance with md spec (right-padding was 16dp, but must be 24)

### DIFF
--- a/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Menu.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Menu.xaml
@@ -43,7 +43,7 @@
     </Style>
 
     <Style TargetType="{x:Type MenuItem}" x:Key="MaterialDesignMenuItem" BasedOn="{x:Null}">
-        <Setter Property="Padding" Value="24 0 16 0"></Setter>
+        <Setter Property="Padding" Value="24 0 24 0"></Setter>
         <Setter Property="OverridesDefaultStyle" Value="True"/>
         <Setter Property="Template">
             <Setter.Value>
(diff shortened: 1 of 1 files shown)


ButchersBoy:
The latest bold text on the menu is making the menu jump around by a pixel when moving the mouse left/rght...

l1pton17:
@ButchersBoy yeah, i have noticed it. Maybe i can resolve this problem in the evening.

ButchersBoy:
Do we even need the bold?

https://www.google.com/design/spec/components/menus.html#menus-usage

l1pton17:
@ButchersBoy wow, on pictures they don't use bold text but on video they do :) So, i suppose there is no so much reason to implement this.

ButchersBoy:
:confused: ...not the first time I have found contradictions...

l1pton17:
Forgot to left a message.
Fixed, i let bold working. Now width of menuitem doesn't change.

ButchersBoy:
So we are sticking with bold?

QuantumDeveloper:
@l1pton17 you are very active)
Maybe you can also do some fixes to the Dragablz ;)

l1pton17:
Why not.
Do you want me to remove bold? I fixed shifting arround a pixel when moving mouse.
According to specs it can be bold.

ButchersBoy:
I only ask cos there is no bold on here: https://www.google.com/design/spec/components/menus.html#
Before you remove it, what is the video you are talking about?

ButchersBoy:
(It does look good though - merged)...but please link the videos so I can see what you are looking at.

l1pton17:
@ButchersBoy here: https://goo.gl/KpXrzq

ButchersBoy:
thanks

PR #96, opened 2015-10-12
Fix ListBox DisplayMemberPath

l1pton17:
Fix ripple to use contenttemplateselector.
And now displaymemberpath in listbox works.


ButchersBoy:
thanks

PR #99, opened 2015-10-17
Palette Helper Changes

SuicSoft:
When changing colors with palette helper they fade quickly (duration can be changed) to make it smooth looking.

### DIFF
--- a/MaterialDesignThemes.Wpf/PaletteHelper.cs
+++ b/MaterialDesignThemes.Wpf/PaletteHelper.cs
@@ -134,8 +134,8 @@ namespace MaterialDesignThemes.Wpf
                 {
                 ColorAnimation animation;
                 animation = new ColorAnimation();
-                animation.From = parentDictionary[entryName]; //The old color
-                animation.To = newValue; //The new color
+                animation.From = parentDictionary[entryName] as Color; //The old color
+                animation.To = newValue as Color; //The new color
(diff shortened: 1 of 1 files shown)


SuicSoft:
Fixed all compile errors

ButchersBoy:
...yes...let's not make anything static.  If anything we should introduce an IPaletteHelper interface to help people with DI, TDD etc

MaterialDesignThemes.Wpf/PaletteHelper.cs
```diff
@@ -123,23 +123,31 @@ public void ReplaceAccentColor(string name)
...
                 return true;
             }
 
             foreach (var dictionary in parentDictionary.MergedDictionaries)
-            {
-                if (ReplaceEntry(entryName, newValue, dictionary))
-                    return true;
-            }
-
+                return ReplaceEntry(entryName, newValue, dictionary);
```
Davipb: The 'if' there was necessary: In your new code, the ReplacyEntry method now returns false if the first merged dictionary fails to have the entry, never looking at the others.

SuicSoft:
Fixed more errors (the if thingy).
Don't know if the iscolor is needed. If all the new values are colors you don't need it. So removed it

SuicSoft:
There should be no problems with the code now. The change is that when a color is changed with PaletteHelper it will fade to the new color. The fade can be customized

PR #102, opened 2015-10-18
Palette helper fade animation

SuicSoft:
Now when changing colors with palette helper it has a fade animation. There is a previous closed pull request of this. The new pull request has a shorter code

### DIFF
--- a/MaterialDesignThemes.Wpf/PaletteHelper.cs
+++ b/MaterialDesignThemes.Wpf/PaletteHelper.cs
@@ -49,8 +49,8 @@ namespace MaterialDesignThemes.Wpf
 
             foreach (var color in swatch.PrimaryHues)
             {
-                ReplaceEntry(color.Name, color.Color, null, false);
-                ReplaceEntry(color.Name + "Foreground", color.Foreground, null, false);
+                ReplaceEntry(color.Name, color.Color);
+                ReplaceEntry(color.Name + "Foreground", color.Foreground);
             }
 
             ReplaceEntry("PrimaryHueLightBrush", new SolidColorBrush(light.Color));
@@ -96,8 +96,8 @@ namespace MaterialDesignThemes.Wpf
 
             foreach (var color in swatch.AccentHues)
             {
-                ReplaceEntry(color.Name, color.Color,null, false);
-                ReplaceEntry(color.Name + "Foreground", color.Foreground, null, false);
+                ReplaceEntry(color.Name, color.Color);
+                ReplaceEntry(color.Name + "Foreground", color.Foreground);
             }
 
             ReplaceEntry("SecondaryAccentBrush", new SolidColorBrush(swatch.AccentExemplarHue.Color));
--- a/MaterialDesignThemes.Wpf/PaletteHelper.cs
+++ b/MaterialDesignThemes.Wpf/PaletteHelper.cs
@@ -153,13 +153,12 @@ namespace MaterialDesignThemes.Wpf
         private static bool ReplaceEntry(object entryName, object newValue, ResourceDictionary parentDictionary = null, bool animate = true)
         {
             const int DURATION_MS = 500; //Change the value if needed
-            int ANIMATION_FPS = GetPowerState().ACLineStatus == Online ? 20 : 60;
             if (parentDictionary == null)
                 parentDictionary = Application.Current.Resources;
             
             if (parentDictionary.Contains(entryName))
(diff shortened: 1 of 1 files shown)


MaterialDesignThemes.Wpf/PaletteHelper.cs
```diff
@@ -123,23 +123,35 @@ public void ReplaceAccentColor(string name)
...
+                if (animate) //Fade animation is enabled
+                {
+                    ColorAnimation animation = new ColorAnimation()
+                    {
+                        From = (Color)parentDictionary[entryName],//The old color
+                        To = (Color)newValue, //The new color
+                        Duration = new Duration(new TimeSpan(0,0,0,0,DURATION_MS))
+                    };
+                    (parentDictionary[entryName] as Brush).BeginAnimation(SolidColorBrush.ColorProperty, animation); //Begin the animation
+                    return true;
```
Davipb: This ```return true``` shouldn't be just inside ```if (animate)```: Right now if we set ```animate``` to ```false``` your code will search the merged dictionaries and may even return ```false``` for the entry even though we already found it.

You should put it after line 149: Inside the ```if (parentDictionary.Contains(entryName))```, but outside the ```if (animate)``` and its ```else```.

ButchersBoy:
Twas a colour not a brush...

image

SuicSoft:
The exception should be fixed now. A timer is used to set the value
I cannot test the code because I don't have Visual Studio installed

SuicSoft:
From now there will be no more changes to this Pull Request

ButchersBoy:
* Is the power state stuff necessary? 
* Instead of try/catch can't we check if we have a Brush beforehand?  This is expensive behaviour.

SuicSoft:
The power state stuff is used to turn off the animation on battery to increase battery life, so it is optional
If you want to avoid the try catch bad performance, make the call to ReplaceEntry on everything that is not a brush like this : ReplaceEntry("resourcename", thenewvaluenotabrush, null, false); thats what the animate argument is for. I have already done this on some (maby all?). Whenever the animate argument is not set to false and the type is not a brush the code will write "The argument : newValue is not a brush." to the debug output (the output window, the project has to be in debug mode to see it)

MaterialDesignThemes.Wpf/PaletteHelper.cs
```diff
@@ -123,24 +151,41 @@ public void ReplaceAccentColor(string name)
...
+                            To = ((SolidColorBrush)newValue).Color, //The new color
+                            Duration = new Duration(new TimeSpan(0,0,0,0,DURATION_MS)) //Set the duration
+                        };
+                        (parentDictionary[entryName] as SolidColorBrush).BeginAnimation(SolidColorBrush.ColorProperty, animation); //Begin the animation
+                    }
+                    catch
+                    {
+                        System.Diagnostics.Debug.WriteLine(@"The argument :" + nameof(newValue) + " is not a brush.");
+                        parentDictionary[entryName] = newValue; //Set the value normally if type is incorrect
+                    }
```
ButchersBoy: What's wrong with making a check instead of an unsafe cast and catching the exception?
```
var solidColorBrush = parentDictionary[entryName] as SolidColorBrush;
if (solidColorBrush == null) return;

var animation = new ColorAnmiation { . . . };
solidColorBrush.BeginAnimation( . . . );
```

MaterialDesignThemes.Wpf/PaletteHelper.cs
```diff
@@ -115,6 +116,33 @@ public void ReplaceAccentColor(string name)
...
+            if (GetSystemPowerStatusRef(state))
+                return state;
+
+            throw new ApplicationException("Unable to get power state");
+        }
+        // Note: Underlying type of byte to match Win32 header
+        public enum ACLineStatus : byte
+        {
+            Offline = 0, Online = 1, Unknown = 255
+        }
```
ButchersBoy: I'm not keen how this is shoe-horned in here.  It makes no consideration of the the rest of the library.

* Is this really an issue?
* Should we be _disabling_ all animations on battery?
  * This is a signification piece of work
  * It should be configurable
  * Maybe if this is a real requirement we should log an issue so I can come up with a more encompassing solution.

SuicSoft:
The null, false stuff is uneeded after removing the try catch.Removing it

SuicSoft:
Merged:smile:

PR #109, opened 2015-10-23
ComboBox

l1pton17:
Make ComboBox according to material design.
If combobox is not editable then selected item isn't displayed in items.

comboboxexample


ButchersBoy:
Hi @l1pton17 , this is heading in the right direction, but I think it looks odd where the underline pushes to the right edge of the popup without padding.  Although I foresee it looking bad if we extend the popup to account for this.

How about we only use the "in place" pop over on non-editable combos?

image

Then we don't need an underline at all.

Thoughts?

l1pton17:
@ButchersBoy yeah, i made it first without underline, give me a sec

l1pton17:
aaaand we run into with differences between pictures and videos again. Here is video of menu where they don't use a right padding: https://material-design.storage.googleapis.com/publish/material_v_4/material_ext_publish/0B6Okdz75tqQsUWtQaDc0MnJOUmc/Components-Menus-Usage-Textfield_Dropdown_xhdpi_006.webm

ButchersBoy:
OK...can we try keeping the existing style for editable combos, and "in-place" for non editable combos?

l1pton17:
@ButchersBoy done. Maybe it will be good if i will add right padding for editable controls?

ButchersBoy:
...maybe...let me download your latest cut and see what it looks like...will have a look later on today.

l1pton17:
@ButchersBoy i add right padding for experiment, it looks not bad for me

ButchersBoy:
OK...will have a look in a bit.

ButchersBoy:
Well done, looks awesome.  One small thing; in editable combos you can't press down to select items from the list, you have to tab first.  Maybe we can get a *Assist behaviour/property silently wired in to restore that behaviour.

PR #127, opened 2015-11-10
Change ComboBox to fix #120

### LINKED ISSUE #120: Editable Combo Textbox Fix
As we have hosting a duplicate text box need to correctly apply focus and respond to keyboard events such as ALT+DOWN etc.

l1pton17:
Fix ComboBox shifting around a pixel while dropdown menu is open.
Also remove a clone of input control in a popup, so we don't need to use comboboxassist.


ButchersBoy:
Problem is, you get this situation if the popup has to move:

image

Also, underline is 1 pixel higher on combo compared to text box (this is not specific to this PR):

image

ButchersBoy:
What are your thoughts on this?  Visually, I really like the style for the "cover-up" but it is problematic.  I am wondering if: 

* Use "cover-up" for non editable combos only
* Provide "cover-up" and "traditional" styles so users can use
* Leave as-is

l1pton17:
@ButchersBoy i think that i could make both styles
combobox

ButchersBoy:
Your animation looks better.  Can I try this *before* you make two styles?  If we can make this work I'd rather a single style than two.

l1pton17:
@ButchersBoy i meant, that will be on combobox style which contains popup with two different template - above and below an input control.

ButchersBoy:
ah OK :-)

ButchersBoy:
Just to be a pain :)  What about this?  I have the combo at the bottom of the screen, straddled across 2 monitors, you can see the popup gets pushed to the side:

image

l1pton17:
@ButchersBoy now we have above and under input control popup template, and also "simple" template if we can use cover up template (in case of 2 monitors for example).

ButchersBoy:
I'll try and have a look this week, and we'll try and get this in for 1.2.

l1pton17:
Sorry, I forgot to fix underline, will be fixed in a few days

l1pton17:
Done

ButchersBoy:
Firstly; this looks awesome - great work.

Secondly, I want to get this bang on now.  I see you've introduced ```ComboBoxPopup``` which gives the extra control; so, can we figure out a way of collapsing the wrapper/clip type row, when the popup displays in an odd position (due to the multiple monitor issue)?  If we could just collapse this under those circumstances then I think we would have it nailed.

I can look at this but you might want to finish the job yourself?  Let me know what you want to do.

l1pton17:
@ButchersBoy if you speak about monitor issue, i have considered it.


default

ButchersBoy:
Ah sorry...I hadn't pulled latest properly, my bad!

Looks great then :smile: :+1:

PR #130, opened 2015-11-13
Added custom xaml namespace

SuicSoft:
Just a small simple addition , a custom xaml namespace like how mahapps has http://metro.mahapps.com/winfx/xaml/controls. 
The custom namespace is http://materialdesigninxaml.net/winfx/xaml/themes for Material Design Themes and http://materialdesigninxaml.net/winfx/xaml/colors for Material Design Colors.

This addition makes code simpler and shorter

Users can still use the old clr-namespace thingy if they like, so it won't break anything emoji

lil painted dog
SuicSoft

### DIFF
--- a/MaterialDesignColors.Wpf/Properties/AssemblyInfo.cs
+++ b/MaterialDesignColors.Wpf/Properties/AssemblyInfo.cs
@@ -16,7 +16,7 @@ using System.Windows.Markup;
 [assembly: AssemblyCopyright("Copyright © 2015")]
 [assembly: AssemblyTrademark("")]
 [assembly: AssemblyCulture("")]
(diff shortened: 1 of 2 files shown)


ButchersBoy:
Hey, yeah, this is good, but, with Resharper I start getting this:

image

How about we go for something a little shorter?
* ```materialDesignTheme```
* ```materialDesignColor``` (do we actually need this one?)

And I say "theme" and not "theme**s**" as it's not plural, there is only one theme.  Then the usage would be quite concise: 

```<materialDesignTheme:Card />```

ButchersBoy:
How about we keep the color one in, just in case we need it in the future, but shorten the main one:

```<materialDesign:Card />```

Sorry to be a pain :frowning: ...but this syntax looks pretty neat :smile: and is shorter still.

ButchersBoy:
Awesome, thanks.

SuicSoft:
Thanks for merge :smile:

PR #134, opened 2015-11-19
Added DWM automatic color selection and few more small changes

SuicSoft:
I have added a feature to automatically generate primary or accent colors based on the DWM (Desktop  Window Manager) color using ```PaletteHelper.AutoPrimary``` or ```PaletteHelper.AutoAccent```. Also I made the palette selector disable the currently selected color like this (circled in green)
image

I also added some useful methods and variables such as ```PaletteHelper.GetClosestSwatch(Color, bool)```

There is a small bug that the Primary or Accent generated from DWM is not disabled in the palette selector. This is a small bug and will not cause any crashes.

Also i had to install the Roslyn compiler NuGet package since I use Visual Studio 2013

PS: Did you see GitHub's new UI


ButchersBoy:
Initial thoughts:

* Please remove Prism.  The toolkit should take no favouritism to any particular MVVM library/framework.  This is a theme-ing library; let's keep it that way and not get dragged into a framework conversation. 
* No public statics on ```PaletteHelper```.
* Getting DWM from the registry: it would be much better to go to the Win32 API. You can see this one one of my other projects: 
  * https://github.com/ButchersBoy/SystemColorChart
  * https://github.com/ButchersBoy/SystemColorChart/blob/master/ColorsModel.cs

SuicSoft:
@ButchersBoy OK , I will do it tommorow since I'm now on Android.

What's the reason for no public static .Static is good for performance

ButchersBoy:
Performance isn't an issue here.  People aren't going to be switching palettes thousands of times a second.  I'd rather something that is a bit more easily swap-able if people are unit testing their view models.  Admittedly this isn't behind an interface; perhaps we should test mocking of this with different mocking frameworks (e.g. we might need to explicitly set a parameter-less constructor etc), but once it's static we would just make it harder for people.

SuicSoft:
I have done everything you said to change. I made the palette selector view model inherit PaletteHelper which is better then having static. Also I removed Prism and used Win32 API instead for the DWM color. Will add for check DWM

ButchersBoy:
I don't care about XP, just 7, 8, 10.  I will have some more comments to work through...,

MahMaterialDragablzMashUp/PaletteSelectorViewModel.cs
```diff
@@ -1,38 +1,42 @@
...
+﻿using System;
+using MaterialDesignColors;
 using MaterialDesignThemes.Wpf;
 using System.Collections.Generic;
 using System.Windows.Input;
 
 namespace MahMaterialDragablzMashUp
 {
-    public class PaletteSelectorViewModel
+    public class PaletteSelectorViewModel : PaletteHelper
```
ButchersBoy: We don't need inheritance here.  If anything we should be illustrating how people can inject in the PaletteHelper, previously we just instantiated it on demand, but if people wanted to TDD a view model to manage palettes they'd most likley inject a mocked version into the constructor.  So maybe just create the PaletteHelper in the constructor.

MahMaterialDragablzMashUp/packages.config
```diff
@@ -2,4 +2,5 @@
 <packages>
   <package id="Dragablz" version="0.0.1.109" targetFramework="net45" />
   <package id="MahApps.Metro" version="1.1.3-ALPHA157" targetFramework="net45" />
+  <package id="Microsoft.Net.Compilers" version="1.1.0-rc1-20151109-01" targetFramework="net45" developmentDependency="true" />
```
ButchersBoy: What's this for?

MainDemo.Wpf/PaletteSelectorViewModel.cs
```diff
@@ -12,34 +12,37 @@
 
 namespace MaterialDesignColors.WpfExample
 {
-    public class PaletteSelectorViewModel
+    public class PaletteSelectorViewModel : PaletteHelper
```
ButchersBoy: Again, no need for inheritance.

MaterialDesignThemes.Wpf/PaletteHelper.cs
```diff
@@ -10,7 +10,95 @@ namespace MaterialDesignThemes.Wpf
...
-        public void SetLightDark(bool isDark)
+        #region Win32
+            [DllImport("dwmapi.dll", PreserveSig = false)]
+            private static extern bool DwmIsCompositionEnabled();
+
+            [DllImport("dwmapi.dll", EntryPoint = "#127")]
+            static extern void DwmGetColorizationParameters(ref DWMCOLORIZATIONPARAMS dp);
+
+            [StructLayout(LayoutKind.Sequential)]
+            public struct DWMCOLORIZATIONPARAMS
```
ButchersBoy: Can this be private?  Don't want this to become part of our API.  Keeping it private will allow internal refactoring without changing our API.

MaterialDesignThemes.Wpf/PaletteHelper.cs
```diff
@@ -10,7 +10,95 @@ namespace MaterialDesignThemes.Wpf
...
+        public Swatch Primary;
+        /// <summary>
+        /// The accent swatch
+        /// </summary>
+        public Swatch Accent;
+
+        /// <summary>
+        /// Primary color generated from DWM color
+        /// </summary>
+        public Swatch AutoPrimary
```
ButchersBoy: Seems like a LOT of activity going on in here for a property?

MaterialDesignThemes.Wpf/PaletteHelper.cs
```diff
@@ -10,7 +10,95 @@ namespace MaterialDesignThemes.Wpf
...
+                    return GetClosestSwatch(ToColor(colorizationParams.ColorizationColor), false);
+                }
+                else
+                    return GetClosestSwatch(Colors.Purple, false); //If you use a slow old computer.
+            }
+        }
+        /// <summary>
+        /// Accent color generated from DWM color
+        /// </summary>
+        public Swatch AutoAccent
```
ButchersBoy: Again, this is a pretty active property.

MaterialDesignThemes.Wpf/PaletteHelper.cs
```diff
@@ -10,7 +10,95 @@ namespace MaterialDesignThemes.Wpf
...
+        }
+        private int GetDiff(Color color, Color baseColor)
+        {
+            int a = color.A - baseColor.A,
+                r = color.R - baseColor.R,
+                g = color.G - baseColor.G,
+                b = color.B - baseColor.B;
+            return a*a + r*r + g*g + b*b;
+        }
+        public void SetLightDark(bool? isDark)
```
ButchersBoy: Not sure I'm keen on making this null-able...diminishing the tightness of the API to handle the fact that a checkbox can be tri state, which is way up in the view.  We shouldn't be bowing to that pressure here.

MahMaterialDragablzMashUp/PaletteSelectorViewModel.cs
```diff
@@ -1,38 +1,42 @@
...
-            Swatches = new SwatchesProvider().Swatches;
-        }
-
-        public ICommand ToggleBaseCommand { get; } = new AnotherCommandImplementation(o => ApplyBase((bool)o));
-
-        private static void ApplyBase(bool isDark)
-        {
-            new PaletteHelper().SetLightDark(isDark);
+            Swatches = new SwatchesProvider().Swatches;  
+            ApplyPrimaryCommand = new DelegateCommand<Swatch>(ApplyPrimary, new Predicate<Swatch>(s => s != Primary));
```
ButchersBoy: We can't automatically apply this in the demo, as it overrides App.xaml, and will confuse people who are creating a specific panel.  We'll need a button for people to explicitly click.

SuicSoft:
@ButchersBoy Why SetLightDark is nullable is so that we can have auto (middle state in toggle button). 
Auto will be light from 7:00am - 5:00pm and will be dark from 5:00pm to 7:00am

ButchersBoy:
If that's the case then that should be managed in your application, not in this library.

SuicSoft:
@ButchersBoy Ok I was thinking of integrating that into the library

SuicSoft:
@ButchersBoy The Microsoft.Net.Compilers is because I use Visual Studio 2013. Also i am a little bit stuck with trying to add an auto item to the swatches list on SwatchesProvider

ButchersBoy:
Have you tried VS2015 Community Edition?

SuicSoft:
I know it's free , but it requires a big download which will waste internet bandwith

SuicSoft:
@ButchersBoy Now i fixed the auto thing. Will fix the diffs

ButchersBoy:
I guess I take internet bandwidth for granted!

SuicSoft:
The reason for inheritance is so that we don't have to create so many palette helper instances

SuicSoft:
I'm getting this error in the AppVeyor build:
xUnit.net Console Runner (32-bit .NET 4.0.30319.42000)
  Discovering: MaterialDesignColors.Wpf.Fixture
  Discovered:  MaterialDesignColors.Wpf.Fixture
  Starting:    MaterialDesignColors.Wpf.Fixture
    MaterialDesignColors.Wpf.Fixture.ResourceProviderFixture.ExcludesBlack [FAIL]
      System.ArgumentNullException : Value cannot be null.
      Parameter name: source
      Stack Trace:
           at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
        MaterialDesignColors.Wpf.Tests\ResourceProviderFixture.cs(18,0): at MaterialDesignColors.Wpf.Fixture.ResourceProviderFixture.ExcludesBlack()
    MaterialDesignColors.Wpf.Fixture.ResourceProviderFixture.IncludesGrey [FAIL]
      System.ArgumentNullException : Value cannot be null.
      Parameter name: source
      Stack Trace:
           at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
        MaterialDesignColors.Wpf.Tests\ResourceProviderFixture.cs(29,0): at MaterialDesignColors.Wpf.Fixture.ResourceProviderFixture.IncludesGrey()
    MaterialDesignColors.Wpf.Fixture.ResourceProviderFixture.BrownHasPrimaries [FAIL]
      System.ArgumentNullException : Value cannot be null.
      Parameter name: source
      Stack Trace:
           at System.Linq.Enumerable.Single[TSource](IEnumerable`1 source, Func`2 predicate)
        MaterialDesignColors.Wpf.Tests\ResourceProviderFixture.cs(53,0): at MaterialDesignColors.Wpf.Fixture.ResourceProviderFixture.BrownHasPrimaries()
    MaterialDesignColors.Wpf.Fixture.ResourceProviderFixture.BrownHasNoAccents [FAIL]
      System.ArgumentNullException : Value cannot be null.
      Parameter name: source
      Stack Trace:
           at System.Linq.Enumerable.Single[TSource](IEnumerable`1 source, Func`2 predicate)
        MaterialDesignColors.Wpf.Tests\ResourceProviderFixture.cs(40,0): at MaterialDesignColors.Wpf.Fixture.ResourceProviderFixture.BrownHasNoAccents()
    MaterialDesignColors.Wpf.Fixture.ResourceProviderFixture.IndigoHasAccents [FAIL]
      System.ArgumentNullException : Value cannot be null.
      Parameter name: source
      Stack Trace:
           at System.Linq.Enumerable.Single[TSource](IEnumerable`1 source, Func`2 predicate)
        MaterialDesignColors.Wpf.Tests\ResourceProviderFixture.cs(66,0): at MaterialDesignColors.Wpf.Fixture.ResourceProviderFixture.IndigoHasAccents()
  Finished:    MaterialDesignColors.Wpf.Fixture
=== TEST EXECUTION SUMMARY ===
   MaterialDesignColors.Wpf.Fixture  Total: 5, Errors: 0, Failed: 5, Skipped: 0, Time: 1.869s
Command exited with code 5

As said previously I do not have VS2015 so maybe you can help with this

ButchersBoy:
The inheritance just isn't necessary.  Why are we exposing the PaletteHelper API through the view model?  The view model should be encapsulating the logic and hiding it away from the outside world, not making a bigger API surface area, and constraining other inheritance chains.  We are setting a bad example for other developers, and then we will have people saying,

"Hi, I am using SomeMVVMLibrary.ViewModelBase and I want to copy PaletteHelperViewModel but you have constricted the inheritance chain, how do I do this?"

Basically, avoid inheritance if it's not necessary. https://en.wikipedia.org/wiki/Composition_over_inheritance

And why do we care how many PaletteHelper instances we create?  It's early, undue optimization which - in this case - would lead to a poorer example of how people could wire up their view models.

We really shouldn't focus too much on performance around this area.  The class is light weight, and the operation is a user-driven action which will not be happening regularly.  The cost of WPF updating the brushes is __way__ more expensive than what we are doing to instigate the action.

SuicSoft:
I will keep the nullable bool since otherwise we will have to cast the toggle button state. Maybe you can use it for something :smile:

ButchersBoy:
I can guarantee the nullable bool won't make it in.

The API is simple

```SetLightDark(bool isDark)```

There is no middle state.  The responsibility of this method on this class is to either set the palette to light or dark.  The fact that check boxes have tri state is not relevant.  That is a view (and view model) concern.  Separation of concerns.  This API looks after its own interests and there is no "just-in-case" requirement.

**Comment by ButchersBoy** [comment]
@SuicSoft You might want to have a look at https://github.com/ButchersBoy/MaterialDesignInXamlToolkit/issues/135

Now we know we have to mark all public methods in this class as ```virtual``` so people can use it when testing their view models.  Losing the performance by marking the methods as virtual is negligible when considering the usage scenarios of this class and making it easier for people when they are using TDD/DI.

**Comment by SuicSoft** [comment]
@ButchersBoy  For some reason i'm getting an error with the Tests in the AppVeyor build

File: MahMaterialDragablzMashUp/PaletteSelectorViewModel.cs [hunk 9]
```diff
@@ -7,32 +8,32 @@ namespace MahMaterialDragablzMashUp
...
+            ApplyPrimaryCommand = new DelegateCommand<Swatch>(ApplyPrimary, new Predicate<Swatch>(s => s != palette.Primary));
+            ApplyAccentCommand = new DelegateCommand<Swatch>(ApplyAccent, new Predicate<Swatch>(s => s != palette.Accent));
+            ToggleBaseCommand = new DelegateCommand<bool>(palette.SetLightDark);
         }
 
+        public ICommand ToggleBaseCommand { get; }
         public IEnumerable<Swatch> Swatches { get; }
 
-        public ICommand ApplyPrimaryCommand { get; } = new AnotherCommandImplementation(o => ApplyPrimary((Swatch)o));
+        public DelegateCommand<Swatch> ApplyPrimaryCommand { get; }
```
ButchersBoy: This should return ICommand; "Depend on abstractions, not concretions"

MahMaterialDragablzMashUp/PaletteSelectorViewModel.cs
```diff
@@ -7,32 +8,32 @@ namespace MahMaterialDragablzMashUp
...
-        private static void ApplyPrimary(Swatch swatch)
+        private void ApplyPrimary(Swatch swatch)
         {
-            new PaletteHelper().ReplacePrimaryColor(swatch);
+            palette.ReplacePrimaryColor(swatch);
+            ApplyPrimaryCommand.RaiseCanExecuteChanged();
         }
 
-        public ICommand ApplyAccentCommand { get; } = new AnotherCommandImplementation(o => ApplyAccent((Swatch)o));
+        public DelegateCommand<Swatch> ApplyAccentCommand { get; }
```
ButchersBoy: This should return ICommand; "Depend on abstractions, not concretions"

MainDemo.Wpf/PaletteSelectorViewModel.cs
```diff
@@ -14,32 +14,32 @@ namespace MaterialDesignColors.WpfExample
...
+            ApplyPrimaryCommand = new DelegateCommand<Swatch>(ApplyPrimary, new Predicate<Swatch>(s => s != palette.Primary));
+            ApplyAccentCommand = new DelegateCommand<Swatch>(ApplyAccent, new Predicate<Swatch>(s => s != palette.Accent));
+            ToggleBaseCommand = new DelegateCommand<bool>(palette.SetLightDark);
         }
 
+        public ICommand ToggleBaseCommand { get; }
         public IEnumerable<Swatch> Swatches { get; }
 
-        public ICommand ApplyPrimaryCommand { get; } = new ActionCommand(o => ApplyPrimary((Swatch)o));
+        public DelegateCommand<Swatch> ApplyPrimaryCommand { get; }
```
ButchersBoy: This should return ICommand; "Depend on abstractions, not concretions"

MainDemo.Wpf/PaletteSelectorViewModel.cs
```diff
@@ -14,32 +14,32 @@ namespace MaterialDesignColors.WpfExample
...
-        private static void ApplyPrimary(Swatch swatch)
+        private void ApplyPrimary(Swatch swatch)
         {
-            new PaletteHelper().ReplacePrimaryColor(swatch);
+            palette.ReplacePrimaryColor(swatch);
+            ApplyPrimaryCommand.RaiseCanExecuteChanged();
         }
 
-        public ICommand ApplyAccentCommand { get; } = new ActionCommand(o => ApplyAccent((Swatch)o));
+        public DelegateCommand<Swatch> ApplyAccentCommand { get; }
```
ButchersBoy: This should return ICommand; "Depend on abstractions, not concretions"

MaterialDesignColors.Wpf/SwatchesProvider.cs
```diff
@@ -36,11 +46,10 @@ public SwatchesProvider()
                     Read(assemblyName, x.SingleOrDefault(y => y.match.Groups["type"].Value == "primary")?.key),
                     Read(assemblyName, x.SingleOrDefault(y => y.match.Groups["type"].Value == "accent")?.key)
                 ))
-                .ToList();
-
+                .ToList());
         }
 
-        public IEnumerable<Swatch> Swatches { get; }
+        public List<Swatch> Swatches { get; }
```
ButchersBoy: This should return IEnumerable<Swatch>; "Depend on abstractions, not concretions"

MaterialDesignThemes.Wpf/PaletteHelper.cs
```diff
@@ -10,6 +10,15 @@ namespace MaterialDesignThemes.Wpf
 {
     public class PaletteHelper
     {
+        /// <summary>
+        /// The primary swatch
+        /// </summary>
+        public Swatch Primary;
```
ButchersBoy: Do we need this?  I am wary of introducing state.  If we are introducing state then we have to look at this a lot more closely to make sure we are handling it correctly.  Which means scanning resources at startup etc which is a risky scenario for us to confidently say to people "this is your current swatch".

So we should not introduce state.

MaterialDesignThemes.Wpf/PaletteHelper.cs
```diff
@@ -10,6 +10,15 @@ namespace MaterialDesignThemes.Wpf
...
+        /// <summary>
+        /// The accent swatch
+        /// </summary>
+        public Swatch Accent;
```
ButchersBoy: See above.  Us trying to manage state in here is unsafe.

SuicSoft:
@ButchersBoy The state is so that we can disable the current swatch from being used. This is null before you fiddle with the PaletteSelector so maybe we should 'hide it' from our users with ```[System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]``` so that no-one will try using it

SuicSoft:
@ButchersBoy Why I have to use ```DelegateCommand<Swatch>``` instead of ```ICommand``` is so that we can update the command state using ```RaiseCanExecuteChanged();```. As you can see I have removed it where possible (the dark light)

SuicSoft:
Still getting the xunit error

SuicSoft:
@ButcherBoy Should i merge some commits into one? This PR has a-lot of commits :smile:

punker76:
@SuicSoft it seems your broke it, cause some tests are failing... :-D

SuicSoft:
@punker76 I don't have Visual Studio 2015 ,do you ?. Then maybe you can debug it and give me the line number and exception message

punker76:
@SuicSoft open your eyes 
https://ci.appveyor.com/project/ButchersBoy/materialdesigninxamltoolkit/build/1.2.1-ci367

SuicSoft:
@punker76 I saw that but it wasn't very helpful

punker76:
@SuicSoft then you should install the VS 2015 community edition and fix the source which fails the tets!

SuicSoft:
As mentioned before it requires a big download

punker76:
@SuicSoft so, I could fix your changes.

1. your changes are creation hell of SwatchesProvider (it's creating really often e.g. by GetClosestSwatch)
2. here my changes that fixes the tests

instead this

```csharp
        public SwatchesProvider ()
	    {
            new SwatchesProvider(true);
	    }
```

do this

```csharp
        public SwatchesProvider()
            : this(true)
        {
        }
```

and instead this

```csharp
        public static Swatch GetClosestSwatch(Color baseColor, bool accent)
        {
            var colors = new SwatchesProvider(false).Swatches.Select(x => new {Value = x, Diff = GetDiff(x.ExemplarHue.Color, baseColor)}).ToList();
            var min = colors.Min(x => x.Diff);
            var color = colors.FindIndex(x => x.Diff == min);
            return accent
                ? new SwatchesProvider(false).Swatches.ElementAtOrDefault(color + 1) != null
                ? new SwatchesProvider(false).Swatches.ElementAtOrDefault(color + 1)
                : new SwatchesProvider().Swatches.ElementAtOrDefault(color - 1)
                : new SwatchesProvider(false).Swatches.ElementAt(color);
        }
```
do this

```csharp
        public static Swatch GetClosestSwatch(Color baseColor, bool accent)
        {
            var colors = new SwatchesProvider(false).Swatches.Select(x => new {Value = x, Diff = GetDiff(x.ExemplarHue.Color, baseColor)}).ToList();
            var min = colors.Min(x => x.Diff);
            var color = colors.FindIndex(x => x.Diff == min);
            return accent
                ? new SwatchesProvider(false).Swatches.ElementAtOrDefault(color + 1) ?? new SwatchesProvider(false).Swatches.ElementAtOrDefault(color - 1)
                : new SwatchesProvider(false).Swatches.ElementAt(color);
        }
```

and instead this

```csharp
        public SwatchesProvider(bool overflow)
        {
            var assembly = Assembly.GetExecutingAssembly();
            var resourcesName = assembly.GetName().Name + ".g";
            var manager = new ResourceManager(resourcesName, assembly);
            var resourceSet = manager.GetResourceSet(CultureInfo.CurrentUICulture, true, true);
            var dictionaryEntries = resourceSet.OfType<DictionaryEntry>().ToList();
            var assemblyName = assembly.GetName().Name;

            var regex = new Regex(@"^themes\/materialdesigncolor\.(?<name>[a-z]+)\.(?<type>primary|accent)\.baml$");

            Swatches = new List<Swatch>();
            if (overflow && DwmIsCompositionEnabled())
                (Swatches as List<Swatch>).Add(new Swatch("Auto", Swatch.AutoPrimary, Swatch.AutoAccent));
            (Swatches as List<Swatch>).AddRange(
                dictionaryEntries
                .Select(x => new { key = x.Key.ToString(), match = regex.Match(x.Key.ToString()) })
                .Where(x => x.match.Success && x.match.Groups["name"].Value != "black")
                .GroupBy(x => x.match.Groups["name"].Value)
                .Select(x =>
                CreateSwatch
                (
                    x.Key,
                    Read(assemblyName, x.SingleOrDefault(y => y.match.Groups["type"].Value == "primary")?.key),
                    Read(assemblyName, x.SingleOrDefault(y => y.match.Groups["type"].Value == "accent")?.key)
                )));
        }
```

i would do this

```csharp
        public SwatchesProvider(bool overflow)
        {
            var assembly = Assembly.GetExecutingAssembly();
            var resourcesName = assembly.GetName().Name + ".g";
            var manager = new ResourceManager(resourcesName, assembly);
            var resourceSet = manager.GetResourceSet(CultureInfo.CurrentUICulture, true, true);
            var dictionaryEntries = resourceSet.OfType<DictionaryEntry>().ToList();
            var assemblyName = assembly.GetName().Name;

            var regex = new Regex(@"^themes\/materialdesigncolor\.(?<name>[a-z]+)\.(?<type>primary|accent)\.baml$");

            var theSwatches = new List<Swatch>();
            if (overflow && DwmIsCompositionEnabled())
                theSwatches.Add(new Swatch("Auto", Swatch.AutoPrimary, Swatch.AutoAccent));
            theSwatches.AddRange(
                                 dictionaryEntries
                                     .Select(x => new { key = x.Key.ToString(), match = regex.Match(x.Key.ToString()) })
                                     .Where(x => x.match.Success && x.match.Groups["name"].Value != "black")
                                     .GroupBy(x => x.match.Groups["name"].Value)
                                     .Select(x =>
                                                 CreateSwatch
                                                 (
                                                  x.Key,
                                                  Read(assemblyName, x.SingleOrDefault(y => y.match.Groups["type"].Value == "primary")?.key),
                                                  Read(assemblyName, x.SingleOrDefault(y => y.match.Groups["type"].Value == "accent")?.key)
                                                 )));
            Swatches = theSwatches;
        }
```

ButchersBoy:
@SuicSoft under further consideration I think you should keep your palette auto detection in your own code/applications:

1) Most people will select a palette for their app up front.
2) You will never get a perfect DWM match any way.
3) Introducing state into ```PaletteHelper``` is not a route I want the library to go down.
4) ```Swatch``` is open for you to create your own palettes without changing the library

PR #136, opened 2015-11-21
a missed WPF control: numeric spinner

spiegelp:
I might add a more decent alternative template later.


ButchersBoy:
Hi @spiegelp ,

Unfortunately I'm going to reject this based on the [Contributing Guidelines](https://github.com/ButchersBoy/MaterialDesignInXamlToolkit/blob/master/CONTRIBUTING.md), which - admittedly - I've only just written.

I just want to keep the scope tight; to existing WPF controls and controls which explicitly support the Material Design components as defined by Google.

So the best place for this control is in your own code.

Thanks for the proposal though; it is appreciated.

PR #142, opened 2015-11-29
Add blackout styles to calendar

### LINKED ISSUE #15: Tidy up error and blackout styles on Calendar/DatePicker.

jdf3:
Fixes #15.

There's a lack of material-styled datepickers on the web that allow disabling dates, but [angular-md](https://material.angularjs.org/latest/demo/datepicker) is one.  Materialize CSS doesn't support disabled dates (yet), though its underlying implementation (`datepicker.js`) does.  Material Design Lite doesn't support a calendar at all.

The angular-md datepicker sets the opacity of the blackout dates to a low (35%-ish) value.

This seems to be the most standard and "material" way of doing things - variations on the base commit in this PR (`9d4147b`) tried a strike-through approach, but this never felt very material to me, especially when a large number of dates were blacked-out.

So, this PR takes the approach of lowering the opacity to indicate disabled/blacked-out dates.

This didn't look nice with the out-of-month dates also having their opacity lowered, so I hid the out-of-month dates from the user.  This feels right: I think having the opacity lowered like that, while Windows-standard, gives the sense that the items are disabled in a material-context, since the official documentation seems pretty rigid regarding which situations get a lower opacity.  [It's also what's done in the spec](https://www.google.com/design/spec/components/pickers.html).

Note that MSFT's implementation of `CalendarItem` explicitly sets `Visibility.Visible` on the `CalendarDayButton` when adding them to the grid, so we set the height to 0 for "inactive" dates.  As a result, not all months appear to have six rows of days (some 4, some 5, some 6).  That means we're saving some vertical space, with the added side effect that scrolling through these months causes the height of calendars styled with these items to vary.  This seems OK - that behavior can be observed in other calendars on the web that take this approach.

### DIFF
--- a/MainDemo.Wpf/Pickers.xaml
+++ b/MainDemo.Wpf/Pickers.xaml
@@ -19,8 +19,10 @@
             <ColumnDefinition Width="Auto" />
         </Grid.ColumnDefinitions>
         <TextBlock Grid.ColumnSpan="3">Classic WPF DatePicker control with Material Design theme, and new TimePicker control:</TextBlock>
-        <DatePicker Grid.Row="1" Grid.Column="0" Width="100" HorizontalAlignment="Left" Margin="0 16 0 0"
-                    materialDesign:TextFieldAssist.Hint="Pick Date"/>
+        <StackPanel Grid.Row="1" Grid.Column="0">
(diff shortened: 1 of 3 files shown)


SuicSoft:
Can you edit this PR to add a screenshot.Nice PRs and issues are nice😃

Issue #145, opened 2015-12-01
Fix Combo clipping on high DPI displays


ButchersBoy:

ButchersBoy:
@l1pton17 I'll guess we'll have to revisit at some point for per-monitor-dpi, but closing for now.  Thanks for great work on drop downs.

PR #146, opened 2015-12-01
Fix ComboBox dpi issue

l1pton17:
Fix issue #145


ButchersBoy:
Just had a quick look on my laptop, this is opening up above.  Haven't  got time to investigate just yet.

image

ButchersBoy:
(Clipping issue looks better though)

l1pton17:
@ButchersBoy i fixed some bugs, but i don't know will it fix your laptop issue (i don't have a laptop).
Also it works bad on two monitos with different dpi, but i can't imagine a man who have two monitors with different dpi.

ButchersBoy:
Well done, works on the laptop!

PR #160, opened 2015-12-13
Add extended example of treeview

l1pton17:
Doesn't fix some bugs, just shows another one =D
Add example of treeview with Add and Remove buttons.
Which also shows that multiactionbutton doesn't adjust PopupBox position
in case of changing it's placement.


ButchersBoy:
Good movie selection! :movie_camera:

PR #162, opened 2015-12-14
Fix buttons inside a PopupBox

jonfunkhouser:
This fixes an issue clicking buttons inside of a PopupBox. The popup is dismissed before any action associated with the button can execute. 

The issue occurs because the content inside a Popup exists in a different visual tree from the Popup, so the button is not considered a visual descendant of the Popup. When the mouse capture is lost to the button, the popup is closed.

The issue is fixed by examining the logical ancestors of the popup content, which includes the Popup. The button click is then properly executed and the popup is closed.

### DIFF
--- a/MaterialDesignThemes.Wpf/Extensions.cs
+++ b/MaterialDesignThemes.Wpf/Extensions.cs
@@ -41,6 +41,28 @@ namespace MaterialDesignThemes.Wpf
                 yield return leaf;
                 leaf = VisualTreeHelper.GetParent(leaf);
             }
-        } 
+        }
+
+        public static IEnumerable<DependencyObject> GetLogicalAncestory(this DependencyObject leaf)
+        {
+            while (leaf != null)
+            {
+                yield return leaf;
+                leaf = LogicalTreeHelper.GetParent(leaf);
+            }
+        }
+
+        public static bool HasAncestor(this DependencyObject leaf, DependencyObject ancestor)
+        {
+            var visualAncestry = leaf.GetVisualAncestory().ToArray();
+            if (visualAncestry.Contains(ancestor))
+            {
+                return true;
+            }
+
+            var lastVisualAncestor = visualAncestry.LastOrDefault();
+            var logicalAncestry = GetLogicalAncestory(lastVisualAncestor);
+            return logicalAncestry.Contains(ancestor);
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


MaterialDesignThemes.Wpf/Extensions.cs
```diff
@@ -41,6 +41,28 @@ public static IEnumerable<DependencyObject> GetVisualAncestory(this DependencyOb
...
+        public static IEnumerable<DependencyObject> GetLogicalAncestory(this DependencyObject leaf)
+        {
+            while (leaf != null)
+            {
+                yield return leaf;
+                leaf = LogicalTreeHelper.GetParent(leaf);
+            }
+        }
+
+        public static bool HasAncestor(this DependencyObject leaf, DependencyObject ancestor)
```
ButchersBoy: I think ```IsDescendantOf``` is a better name.  So usage would be:

```leaf.IsDescendantOf(ancestor) ```

Which makes me think I should rename the original ```IsDescendant``` to ```IsAncestorOf```.

File: MaterialDesignThemes.Wpf/Extensions.cs [hunk 2]
```diff
@@ -41,6 +41,28 @@ public static IEnumerable<DependencyObject> GetVisualAncestory(this DependencyOb
...
+        {
+            var visualAncestry = leaf.GetVisualAncestory().ToArray();
```
ButchersBoy: I don't think we need the ```ToArray()``` here?

jonfunkhouser: I added `ToArray()` so the enumerable isn't iterated twice.

MaterialDesignThemes.Wpf/Extensions.cs
jonfunkhouser: Do you want me to include a commit that renames these two methods?

MaterialDesignThemes.Wpf/Extensions.cs
ButchersBoy: How about this, otherwise we are always going to iterating visual ancestry at least once; for the array, and then possible the whole array again.

```
            DependencyObject parent = null;
            foreach (var node in leaf.GetVisualAncestory())
            {
                if (Equals(node, ancestor))
                    return true;

                parent = node;
            }

            return parent.GetLogicalAncestory().Contains(ancestor);
```

(havent tested that code)

and damn...just realist I am spelling ancestry wrong in the code :worried:

MaterialDesignThemes.Wpf/Extensions.cs
ButchersBoy: Yes please!

ButchersBoy:
Thank you.

Issue #174, opened 2015-12-20
SetLightDark crashes


4ipalino:
My Application crashes always when i call `new PaletteHelper().SetLightDark(isDark);` with an `System.InvalidOperationException`.

StackTrace:

    bei System.Uri.get_AbsolutePath()
    bei MaterialDesignThemes.Wpf.PaletteHelper.<>c.<SetLightDark>b__0_1(ResourceDictionary rd)
    bei System.Linq.Enumerable.SingleOrDefault[TSource](IEnumerable`1 source, Func`2 predicate)
    bei MaterialDesignThemes.Wpf.PaletteHelper.SetLightDark(Boolean isDark)

ButchersBoy:
Please retry on latest.

ButchersBoy:
Merged your PR so closing.

l4km47:
`
namespace MaterialDesignThemes.Wpf
{
    public class PaletteHelper
    {
        public virtual void SetLightDark(bool isDark)
        {
           //[...]
//remove this if fromPaletteHelperClass
    //if (existingResourceDictionary == null)
            //    throw new ApplicationException("Unable to find Light/Dark base theme in Application resources.");
`

PR #175, opened 2015-12-21
Fix SetLightDark fpr Relative Strings

4ipalino:
There was one more Abslout Path. Replacing this also with OriginalString solved the problems on Issue #174 for me.

Sorry I never before created a PullRequest. Now the fix fpr SetLightDark and the FloatingStyle for the PasswordBox are together in this Request. Tried to revert it but that created only more Commits =( 

Next time I will branch for each feature =)


ButchersBoy:
Good job, thank you.

PR #177, opened 2015-12-22
Hyperlink style

FrTerstappen:
I just added styles for the hyperlink control.
They are size and font weight matching with the Textblock styles so that it is easy to use them beside text. Foreground and hover color are based on the current accent (PrimaryHue dark and light).

Using the accent should mirror the style descibed on this page: [Google material design](https://www.google.com/design/spec/style/color.html#color-ui-color-application) under the paragraph accent color.

I also added them to the typography page of the demo application.

This is my first contribution on git-hub i hope that i have not done anything wrong.

### DIFF
--- a/MainDemo.Wpf/Typography.xaml
+++ b/MainDemo.Wpf/Typography.xaml
@@ -1,69 +1,81 @@
 ﻿<UserControl x:Class="MaterialDesignColors.WpfExample.Typography"
              xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
-             xmlns:d="http://schemas.microsoft.com/expression/blend/2008" 
-             mc:Ignorable="d" 
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+             xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+             mc:Ignorable="d"
              d:DesignHeight="300" d:DesignWidth="300">
     <UserControl.Resources>
         <Style TargetType="TextBlock" BasedOn="{StaticResource MaterialDesignCaptionTextBlock}">
             <Setter Property="Opacity" Value=".68"></Setter>
         </Style>
     </UserControl.Resources>
-    
-    <Grid Margin="32">
-        <Grid.ColumnDefinitions>
-            <ColumnDefinition Width="Auto" />
-            <ColumnDefinition Width="64" />
-            <ColumnDefinition Width="Auto" />
-        </Grid.ColumnDefinitions>
-        <Grid.RowDefinitions>
-            <RowDefinition Height="Auto" />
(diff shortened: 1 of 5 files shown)


ButchersBoy:
The mouse over goes from dark -> light, with light being the same as disabled.  Would going from mid -> dark look better (or dark -> mid)?

FrTerstappen:
I just applied your change but i think this is also not very satisfying. I think the problem is that there is no disabled link. A disabled link is just text. Maybe using the Foreground color would be better.

i have also removed the underline style as this is not present in the material design for a link.

ButchersBoy:
I think this looks much better.  For completeness we should hook the hyperlinks into the textblock styles, like this:

```
<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
                    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
        <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Hyperlink.xaml" />
    <Style TargetType="{x:Type TextBlock}" x:Key="MaterialDesignCaptionTextBlock">
        <Style.Resources>
            <Style TargetType="Hyperlink" BasedOn="{StaticResource MaterialDesignCaptionHyperlink}" />
        </Style.Resources>
        <Setter Property="FontSize" Value="12"/>
        <Setter Property="FontWeight" Value="Regular"/>
    </Style>
    . . .
```

This means you can include hyperlinks in a TextBlock without having to manually style the link:

```
<TextBlock Grid.Row="9" Grid.Column="2" Style="{StaticResource MaterialDesignCaptionTextBlock}" Margin="0 8 0 8">
    <Run>Regular 12sp</Run>
    <Hyperlink>link</Hyperlink>
</TextBlock>
```

ButchersBoy:
@FrTerstappen re. the disabled link...like you say...maybe the disabled link should just appear in the body colour (TextElement.Foreground) colour???

FrTerstappen:
Okay the hyperlink style is no available in the text blocks.

Changing:
``` 
<Trigger Property="IsEnabled" Value="false">
    <Setter Property="Foreground"
                Value="{DynamicResource PrimaryHueLightBrush}" />
</Trigger>
```
to:
```
<Trigger Property="IsEnabled" Value="false">
    <Setter Property="Foreground"
                 Value="{DynamicResource TextElement.Foreground}" />
</Trigger>
```
does not change the disabled hyperlinks to black. Instead they look the same as a normal (not mouse over) hyperlink. Sorry if i misunderstood your last comment.

ButchersBoy:
Something like this maybe:

```
{Binding RelativeSource={RelativeSource AncestorType={x:Type FrameworkElement}}, Path=(TextElement.Foreground)}
```

FrTerstappen:
This works. Its even better than what i expected. The disabled link is now a bit lighter than the normal text but i think that it looks good that way.

ButchersBoy:
Thank you.   Made a couple of mods (see commit message).

FrTerstappen:
Okay, thank you. The style inheritance part looks nice. I have to look a bit more into that.
I work mostly backend so this is new for me.

PR #178, opened 2015-12-23
Added ButtonAssist with IconProperty

4ipalino:


ButchersBoy:
Why would we do this when we can put any content, in any layout we want, into a button, using the standard features of WPF?

4ipalino:
I like standard controls that are easy to use and controls that work all nearly the same way. 
In my case i have some menuitems where i can set an icon property, but on my button i had to replace the content. The problem with replacing the content is that the style of this single button is different than all other buttons. Cause the textblockstyle got lost.

ButchersBoy:
OK, but this standard control for you wouldn't be a standard control for everyone else.  What if other uses want icons on the opposite side or above/below etc, or icon/text/icon, or any other form of content.  For that reason I can't accept this change.

If you want the re-useabilty for the button content then in this case I think that the home for some form of (User)Control is in your code-base.

PR #194, opened 2016-01-10
Background support for MaterialDesignCheckBox

xaviergonz:
Added background template support for material design checkbox (before it was fixed to primary mid hue), plus also 3 new styles: MaterialDesignLightCheckBox, MaterialDesignDarkCheckBox, MaterialDesignActionCheckBox

Also consider removing MaterialDesignUserForegroundCheckbox, since it does not allow you to select a different color for the checkbox and the text.

### DIFF
--- a/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.CheckBox.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.CheckBox.xaml
@@ -86,7 +86,7 @@
                             <Canvas Width="24" Height="24">
                                 <Path x:Name="Graphic"
                                       Data="M19,3H5C3.89,3 3,3.89 3,5V19A2,2 0 0,0 5,21H19A2,2 0 0,0 21,19V5C21,3.89 20.1,3 19,3M19,5V19H5V5H19Z" 
-                                      Fill="{TemplateBinding Background}" />
+                                      Fill="{DynamicResource MaterialDesignCheckBoxOff}" />
                                 <Ellipse x:Name="InteractionEllipse" Fill="{TemplateBinding Foreground}" Width="0" Height="0" Canvas.Top="12" Canvas.Left="12" Opacity="0" RenderTransformOrigin="0.5,0.5"
                                          IsHitTestVisible="False">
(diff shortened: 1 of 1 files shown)


MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.CheckBox.xaml
```diff
@@ -114,11 +116,14 @@
                         <Trigger Property="IsPressed" Value="true"/>
                         <Trigger Property="IsChecked" Value="true">
                             <Setter Property="Data" TargetName="Graphic" Value="M10,17L5,12L6.41,10.58L10,14.17L17.59,6.58L19,8M19,3H5C3.89,3 3,3.89 3,5V19A2,2 0 0,0 5,21H19A2,2 0 0,0 21,19V5C21,3.89 20.1,3 19,3Z" />
-                            <Setter Property="Fill" TargetName="Graphic" Value="{DynamicResource PrimaryHueMidBrush}" />
+                        </Trigger>
+                        <Trigger Property="IsChecked" Value="False">
```
ButchersBoy: Isn't this trigger superfluous, considering the default of IsChecked is false?

I think it would stream-lined to revert the Graphic .Fill change and remove the trigger.  Without testing it should have same effect, but with less markup.

xaviergonz:
The check is made because it is not possible (without some hacks) to access a templatebinding from a trigger. If you try to use the templatebinding from the trigger (which would look indeed better) the compiler will work but it will fail at runtime (that was my first approach).

That's why I set the default color to the check state and inversed it on the trigger.

ButchersBoy:
```
{Binding RelativeSource={RelativeSource TemplatedParent}, Path=Background}
```
?

Although I am probably splitting heirs now :)

xaviergonz:
I so didn't now about that binding! I updated the patch accordingly

ButchersBoy:
Looking better...now to be a pain...

Can you do the same for the RadioButton styles, so we have continuity?

xaviergonz:
Done and made a pull request with it

ButchersBoy:
:+1:

PR #195, opened 2016-01-11
Fix for MaterialDesignFlatToggleButton

xaviergonz:
This is a fix for MaterialDesignFlatToggleButton so it can adapt to any kind of size (even non rectangular ones) by using a render scale transform instead of fixed values.
The same could be considered for the action button.

### DIFF
--- a/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.ToggleButton.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.ToggleButton.xaml
@@ -177,94 +177,98 @@
 		<Setter Property="Foreground" Value="{DynamicResource SecondaryAccentForegroundBrush}"/>
 	</Style>
 
-    <Style x:Key="MaterialDesignFlatToggleButton" TargetType="{x:Type ToggleButton}">
-		<Setter Property="Foreground" Value="#616161"/>
-		<Setter Property="Background" Value="#e0e0e0"/>
-		<Setter Property="Width" Value="40"/>
-		<Setter Property="Height" Value="40"/>
-		<Setter Property="FontSize" Value="18"/>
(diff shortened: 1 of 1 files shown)


ButchersBoy:
Hi, this is a good change, but now the markup formatting is different than the rest of the file & project; please stick within the existing standards.

We probably don't need to rename CheckedEllipse to CheckedEllipseScale (nat a biggie), but we just need to change the relative storyboard stuff and keep the layout in-tact.

xaviergonz:
Ok I fixed the formatting with tabs, but it broke a bit the old formatting still because it was wrong-ish.
Out of curiosity, what's wrong with naming a transform and referencing it directly via name? :)

ButchersBoy:
| "Out of curiosity, what's wrong with naming a transform and referencing it directly via name"

Nothing...if you create a storyboard in Blend it typically generates it the long-hand way.  So depends if I create it in Blend, copy-n-paste it, hand-crank it...I use both methods depending what day of the week it is :)

xaviergonz:
Oh, I did it by hand...

ButchersBoy:
:+1: Thankyou, and welcome to the contributors list!

xaviergonz:
Thank you! Btw, I have made a flat toggle button which is rectangular instead of circular. 

Basically it is like a mix between the flatbutton and the flattogglebutton, where it acts exactly like a flat toggle button except that once toggled it enables a fixed background.

Are you interested in it? If so, what should be its name?

ButchersBoy:
Basically, the entry requirement for new controls/styles is that it's either part of Google's spec, or it exists in one of Google's apps/sites and is most likely to make its way into their spec.  I'm just wary about throwing too much stuff in and losing focus.  So it depends if it meets that criteria...

xaviergonz:
Sounds good, to be honest I'm not sure if it matches the criteria. I
basically use it to open and close popups (I guess google would use the
vertical three dots for that).

PR #198, opened 2016-01-12
Added CacheMode to shadow effects for better performance

xaviergonz:
Basically added decorators with the cache mode enabled wrapping elements with associated shadow effects wherever possible in order to speed up performance by a lot. Its effects can be seen on the flyout animations of the demo app for example.


xaviergonz:
While at it fixed the buttons so they are rounded as per google spec. The shadow border was square and on top of the rounded background border, thus hiding it.
Also removed some animations and elements that never were used.

PR #203, opened 2016-01-12
Made ripple effect more according to google spec

xaviergonz:
As said, tweaked the animations so they look more like polymer


xaviergonz:
mixed both pull requests into one

xaviergonz:
Just modified it to be even better looking :+1:

PR #204, opened 2016-01-12
Part of made ripple effect animation more like google

xaviergonz:
This is needed as well as the other previous change to make ripple effects work more like polymer components.


ButchersBoy:
Can you close this and add this as part of your other Ripple PR so I can review as one please.

PR #208, opened 2016-01-13
Fix scroll bar for textbox

xaviergonz:
Fix so when a textbox scrollbar is set to visible it shows on its proper position.


ButchersBoy:
:+1:

PR #209, opened 2016-01-13
Better Ripple FX

xaviergonz:
Tweaked the animations so they look more like polymer


xaviergonz:
Since this is a prerequisite of custom ink color ill put pull requests for those if this patch and cache performance is committed.

ButchersBoy:
Before we had a nice effect where the expansion slows (when you hold the mouse button) which is present on a lot of the  Android apps.  This change seems to have removed that...

xaviergonz:
I was modifying it to mimic google polymer animations as shown here:

however in order to get a slower animation on press it is just needed to modify the *two* lines like this:
```
<EasingDoubleKeyFrame KeyTime="0:0:0.6" Value="1">
```

and more precisely the number 0:0:0.6 and tweak it to a higher value to taste (or even make it a dynamic resource perhaps).

xaviergonz:
Slowed it down from 600msecs to 1.2secs, better?

ButchersBoy:
re. custom ripple colour, somone has asked for this, so you might make their day:
https://github.com/ButchersBoy/MaterialDesignInXamlToolkit/issues/167

PR #215, opened 2016-01-14
set PasswordBox HorizontalContentAlignment and VerticalContentAlignme…

campersau:
…nt to Stretch
which displays the mouse cursor as IBeam over the complete passwordbox

Without this fix:
image

With this fix:
image

This fix also aligns the differences between `MaterialDesignTheme.PasswordBox.xaml` and  `MaterialDesignTheme.TextBox.xaml`.


ButchersBoy:
Thank you :smile:

Issue #219, opened 2016-01-18
DatePicker: Exception using Nullable-DateTime


Matchile:
Whenever I bind a DatePicker to a nullalbe Date I get the following exception.

EDIT: The issue only persist, when turning Validation on.
```
System.Windows.Markup.XamlParseException occurred
  HResult=-2146233087
  LineNumber=0
  LinePosition=0
  Message=Die angegebene Umwandlung ist ungültig.
  Source=PresentationFramework
  StackTrace:
       bei System.Windows.FrameworkTemplate.LoadTemplateXaml(XamlReader templateReader, XamlObjectWriter currentWriter)
  InnerException: 
       HResult=-2147467262
       Message=Die angegebene Umwandlung ist ungültig.
       Source=MaterialDesignThemes.Wpf
       StackTrace:
            bei MaterialDesignThemes.Wpf.Underline.get_IsActive()
            bei MaterialDesignThemes.Wpf.Underline.SelectStateName()
            bei MaterialDesignThemes.Wpf.Underline.IsActivePropertyChangedCallback(DependencyObject dependencyObject, DependencyPropertyChangedEventArgs dependencyPropertyChangedEventArgs)
            bei System.Windows.DependencyObject.OnPropertyChanged(DependencyPropertyChangedEventArgs e)
            bei System.Windows.FrameworkElement.OnPropertyChanged(DependencyPropertyChangedEventArgs e)
            bei System.Windows.DependencyObject.NotifyPropertyChange(DependencyPropertyChangedEventArgs args)
            bei System.Windows.DependencyObject.UpdateEffectiveValue(EntryIndex entryIndex, DependencyProperty dp, PropertyMetadata metadata, EffectiveValueEntry oldEntry, EffectiveValueEntry& newEntry, Boolean coerceWithDeferredReference, Boolean coerceWithCurrentValue, OperationType operationType)
            bei System.Windows.StyleHelper.ApplyTemplatedParentValue(DependencyObject container, FrameworkObject child, Int32 childIndex, FrugalStructList`1& childRecordFromChildIndex, DependencyProperty dp, FrameworkElementFactory templateRoot)
            bei System.Windows.StyleHelper.InvalidatePropertiesOnTemplateNode(DependencyObject container, FrameworkObject child, Int32 childIndex, FrugalStructList`1& childRecordFromChildIndex, Boolean isDetach, FrameworkElementFactory templateRoot)
[...]

PR #228, opened 2016-01-21
Fix crash when DatePicker value is invalid

jonfunkhouser:
An InvalidCastException was thrown on any validation errors


jonfunkhouser:
Fix for #219

ButchersBoy:
Thank you

PR #237, opened 2016-01-30
Add simple search and copy to clipboard function

AuroraDysis:
I don't know do you want to reference interaction, so I did not use it :-P.


ButchersBoy:
Thanks.  Yes, you did the right thing leaving interaction out.  Want the demo to remain as un-opinionated and vanilla as possible.

PR #256, opened 2016-02-11
nameof(...) according to MSDN; usings removed

gobetti:
https://msdn.microsoft.com/library/dn986596.aspx

It makes more sense to use the property and not the DependencyProperty
name as the argument of nameof. Also this is what Microsoft does.

### DIFF
--- a/MaterialDesignThemes.Wpf/ComboBoxPopup.cs
+++ b/MaterialDesignThemes.Wpf/ComboBoxPopup.cs
@@ -1,10 +1,5 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
-using System.Reflection;
-using System.Text;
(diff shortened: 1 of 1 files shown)


ButchersBoy:
Thanks for all these.  I've got a ReSharper template to generate all my dprops...I guess it's time to update it!

gobetti:
Actually, VS2015 by itself has a propdp snippet that still inserts a dependency property without using nameof. And I believe they are right to do so, because using nameof will require developers to update their Visual Studio in order to compile the project... But I noticed your project was already making use of it in other files, and this is not a commercial project where things need to be _really_ retro-compatible, so why not using some C# good news :)

ButchersBoy:
Yep, it's my project, and I choose to use C#6.0  :wink: 

But the compiled version supports < VS2015.

PR #257, opened 2016-02-11
Uses C#6 nameof feature to improve maintainability

gobetti:
This will require VS2015+, but nameof was already being used anyway.
Maintainability is improved because Intellisense does not work for names passed as string literals.


SuicSoft:
:+1:

PR #266, opened 2016-02-19
TimePicker now falls back to previous valid value

chris84948:
Changes made to force behavior of TimePicker to more closely follow the
behavior of the DatePicker control. When there is previously a good
value selected for the time and the user then enters an invalid time,
the time will fall back to the previous good value. The time parsing has
also been opened up to allow more formats.


ButchersBoy:
Hi,  I'll try & pull this soon, just been busy on other stuff (and taking a breather).

James

chris84948:
No big deal, just a minor change I made.

ButchersBoy:
Thank you :smile:

PR #272, opened 2016-02-28
Disable state

l1pton17:
Some controls don't have a disable state.
So i've added it for:
- ComboBox
- ToolButton
- TimePicker
- CheckBox
- HamburgerToggleButton


ButchersBoy:
Thanks...skipped over some of those in the early days when I was racing to get things up and running :+1:

PR #273, opened 2016-02-28
Improve expander animation

l1pton17:
Now expander expands and collapses without using ScaleY animation.
And also this PR fix issue #274


ButchersBoy:
Looks awesome (as always!)   thanks

Issue #167, opened 2015-12-18
Change Ripple effect color


SuplenC:
As described in title. Is it possible for now to change ripple effect color for only one object? In some cases having ripple accent colored would look nice.

jdf3:
Fully addressed by #297

ButchersBoy:
RippleAssist.

PR #227, opened 2016-01-21
Custom ink where possible

xaviergonz:
Using new RippleAssist.Feedback attached prop

PR #297, opened 2016-03-17
Simple ripple feedback via attached property on RippleAssist

### LINKED ISSUE #289: Ripple color (feedback) is often un-modifiable (e.g., for buttons)
Taking a peek at source, it looks like in many styles' templates, `Ripple.Feedback` is usually set to `"{DynamicResource MaterialDesignFlatButtonRipple}"`, with no template binding (e.g., ripple in `ControlTemplate`s of button styles).  I'd really like some control over this.

Here's a simple case: I'm stealing the colors from the albums and putting the album cover on a card that's that "stolen" color (excuse the debugging color bars on the bottom...):

image

So there's some UI logic going on here to make sure we don't put dark text on dark backgrounds.  It's all accommodated quite nicely, aside from ripple - since I'm using the Light theme, I seem to be stuck with Light-themed ripple, even on dark backgrounds (yes, there's ripple feedback there!):

image

I don't believe there's a sane workaround, but there's a possibility I missed something.  I don't think the expectation here should be that I load the light/dark themes as needed, per-card...

jdf3:
Fixes #289.  Implements the spirit of #227, but does not add ripple to calendar navigation buttons as that does.

Added a single example (raised button) to demo - using a dark ripple on a light accent button.

### DIFF
--- a/MainDemo.Wpf/Buttons.xaml
+++ b/MainDemo.Wpf/Buttons.xaml
@@ -122,6 +122,11 @@
                     ToolTip="Resource name: MaterialDesignRaisedAccentButton">
                 ACCENT
             </Button>
+            <Button Style="{StaticResource MaterialDesignRaisedAccentButton}"
+                    Margin="0 0 8 0"
+                    Width="150"
+                    ToolTip="Resource name: MaterialDesignRaisedAccentButton"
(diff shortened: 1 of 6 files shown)


jdf3:
Also fixes #167

ButchersBoy:
Welcome to the contributors list!

:tada: :balloon: :cake:

PR #298, opened 2016-03-17
Change the divider color to 12% opacity per-spec.  Fixes #287.

### LINKED ISSUE #287: Separator color/opacity doesn't match MD guidelines
"Dividers are 1dp thick, with an opacity of either 12% black in light themes or 12% white in dark themes."

At the moment it looks very dark and when I change it to black + 12% opacity it looks very nice and per spec

jdf3:
Refrained from cleanup in case you wanted to park this until a larger milestone, since these constitute visual changes.

Dark:
Before:
image

After:
image

Light:
Before:
image

After:
image


ButchersBoy:
There's a few other things that use this, such as the combo box...I think there were a few areas in Google's early spec' that conflicted or weren't defined fully, and the usage within the toolkit got a bit blurred.

We just need to go back and check the other usages.  It may be that we need to break into a new brush to keep some of the other stuff in tact.  (Can't say for sure without going back and looking at the Google specs)

jdf3:
Will look for other usages.

PR #303, opened 2016-03-18
SmartHint

l1pton17:
According to DRY i don't like current floating hint animation duplicates.
So i have made the SmartHint control which can be used in the both ways: simple or floating hint (the behavior can be switched by UseFloating property).
How to use it:
Just add

	<wpf:SmartHint  x:Name="Hint"
					Hint="{Binding Path=(wpf:TextFieldAssist.Hint), RelativeSource={RelativeSource TemplatedParent}}"
                    HintProxy="{Binding RelativeSource={RelativeSource TemplatedParent}, Converter={x:Static converters:HintProxyFabricConverter.Instance}}"
                    FontSize="{TemplateBinding FontSize}"
                    HintOpacity="{Binding Path=(wpf:TextFieldAssist.HintOpacity), RelativeSource={RelativeSource TemplatedParent}}"
                    UseFloating="{Binding Path=(wpf:HintAssist.UseFloating), RelativeSource={RelativeSource TemplatedParent}}"
                    />
					
Into the controltemplate near the textinput.
And we can obtain a FloatingHint style in the easiest way:

    <Style x:Key="MaterialDesignFloatingHintTextBox" TargetType="{x:Type TextBox}" BasedOn="{StaticResource MaterialDesignTextBox}">
        <Setter Property="wpf:HintAssist.UseFloating" Value="True"/>
    </Style>
	
Just changing of attached property. In general we don't need to use a separate floatinghint style, we can just change the HintAssist.UseFloating property on the control.

How it works:
SmartHint uses a HintProxy property to obtaining information from the target control (TextChanged, IsVisibleChanged, Loaded events, Text property) through the IHintProxy interface.
I made the HintProxyFabric class that can build the IHintProxy from uicontrol (i do it for ComboBox, PasswordBox and TextBox, i suppose this is enough).
This fabric can be extended by means of HintProxyFabric.RegisterBuilder method (just pass predicate and own factory-function).
Also i made the HintProxyFabricConverter.Instance converter that converts a control into IHintProxy.

Now the negative side.
As the SmartHint can be used in both ways, i had to use a RenderTransform.Translate animation instead of Thikness animation.
Because of this, a floating hint can overlaps other controls if the input control doesn't have enough top margin height.
It means that we must take into account the margin of control.
For example:

        <StackPanel Orientation="Horizontal"
                    Margin="0 12 0 0">
            <materialDesign:PackIcon Kind="Key"
                                 Foreground="{Binding ElementName=FloatingPasswordBox, Path=BorderBrush}" />
            <PasswordBox x:Name="FloatingPasswordBox"
                        materialDesign:TextFieldAssist.Hint="Floating Password"
                        Style="{StaticResource MaterialDesignFloatingHintPasswordBox}" />
        </StackPanel>
I had to add Margin="0 12 0 0" to give a floating hint enough space.

So i added Margin="0 12 0 0" as a default margin into text controls (i don't think this is gread idea).

Pros:
- You can get the FloatingHint in the one line.
- SmartHint can be used as a standalone control.
- TextBox and ComboBox with simple or floatinghint style will always have the same base-line.
- Got rid of a lot of code (only one template in ComboBox and TextBox).
- An unified approach to implementing hints.
- Now hint can be easly extended by extending HintAssist and SmartHint classes.
- Can be added into controls in the easiest way. (I added FloatingHint into DatePicker and TimePicker in 5 minutes).

Cons:
- We must notice that FloatingHint can overlaps the other controls, so we have to change the margin.

My next suggestion is to union Hint, HintOpacity, HintVisibility properties into the HintAssist class, but it will be the breaking changes.


ButchersBoy:
Sounds good; let me finalise 1.5, and I will do a code review for 1.5.1.

ButchersBoy:
OK...looking pretty good so far.  By coincidence I've just updated the icon pack and the items have been removed/renamed from the source, so following semver the next version will have to be 2.0.0.  

So now is the opportunity to make breaking changes.  Do the HintAssist consolidation on this PR and I will do another code review.

Can you document all breaking changes in this wiki article: 
https://github.com/ButchersBoy/MaterialDesignInXamlToolkit/wiki/Version-2.0.0---Breaking-Changes

It will be important to communicate these changes.

MaterialDesignThemes.Wpf/SmartHint.cs
```diff
@@ -0,0 +1,179 @@
...
+                var state = string.IsNullOrEmpty(HintProxy.Text)
+                    ? MaterialDesignStateTextEmptyName
+                    : MaterialDesignStateTextNotEmptyName;
+
+                VisualStateManager.GoToState(this, state, useTransitions);
+            }));
+        }
+    }
+
+    public interface IHintProxy : IDisposable
```
ButchersBoy: Can we pop this into it's own file and add a couple of comments to the header...it's a pretty important interface.

l1pton17: @ButchersBoy done

ButchersBoy:
Great work - again!

FYI, I made a couple of changes: renamed a property, and adjusted some code to match ReSharper standards.

Also, I opened up the hint to more complex content; wanted to do this ages ago, but your change has made it easier to make a global change :+1:  , and whilst we are changing the API it seemed sensible.  So now we can do this:

image

punker76:
@ButchersBoy Can I steal this enhancement for MahApps?  ;-)

ButchersBoy:
Stick it in Controlz, and paket it in to both projects? This is mostly @l1pton17 's work; you wanna contribute to SmartHint to Controlz?

PR #331, opened 2016-04-07
Fixes fix of MathConverter

l1pton17:
Double.TryParse(value.ToString(), NumberStyles.Any, ... ) converts input string like "23.066" into the 23066 number, and i got this:
blob

So i've changed Double.TryParse to Convert.ToDouble with try-catch, i hope it will works with german system.

### DIFF
--- a/MaterialDesignThemes.Wpf/Converters/MathConverter.cs
+++ b/MaterialDesignThemes.Wpf/Converters/MathConverter.cs
@@ -22,13 +22,12 @@ namespace MaterialDesignThemes.Wpf.Converters
 
         public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
         {
-            double value1, value2;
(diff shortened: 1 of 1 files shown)


campersau:
Hmm.. just looking at the source and it seams that `value` and `parameter` can either be a `string` or a `double`.

And if it is already a double `value.ToString()` converts it to a string but in the wrong format (e.g. not `CultureInfo.InvariantCulture`).

So wouldn't it be better to check if `value` and `parameter` are already a `double` and the do not the conversion?

l1pton17:
@campersau Convert.ToDouble internally call IConvertable.ToDouble method, which is implemented in Double value type and which returns itself.

campersau:
Ah, I see. Looks good.

Nit: You can remove `System.` from `System.Convert` since `System` it is already imported `using System;`.

l1pton17:
@campersau i can't, because System.Convert is overlapped by IValueConverter.Convert method :)

campersau:
Yeah, `Better go to bed` too :smile:

Issue #267, opened 2016-02-19
Style for GroupBox


Sergey-Terekhin:
Hi. It is seems what style for standard GroupBox is absent. May be it is linked with #57?

mgnslndh:
@ButchersBoy I guess we can close this?

ButchersBoy:
Yes :)

PR #332, opened 2016-04-08
Styles for material group box

mgnslndh:
Made a simple style for the `GroupBox` control. Such a style was asked for in issue #267

This PR also introduce the `ColorZoneAssist` helper which can be used for styles embedding a `ColorZone` and wish to expose the `Mode` property of the `ColorZone`.

image


ButchersBoy:
Excellent, thanks

PR #333, opened 2016-04-08
Fix Ripple can't reset in Popup control.

higankanshi:
Hi!
This PR is fixed a issue about `Ripple` control.
If you use `Ripple` in a `Popup` control, the state of `Ripple` will not reset after you make the mouse up.
Because the `Ripple` only register the mouse event of `Window`.


ButchersBoy:
Thank you, and welcome!

PR #336, opened 2016-04-11
Added SelectedUnfocused

Imhotebkin:
Added a new visual state for unfocused selected objects so that there can be multiple lists on one page and you can clearly tell which one is selected


ButchersBoy:
Thankyou

Issue #343, opened 2016-04-18
The position of the error message is not aligned


lobin-z0x50:
If the Width of the Control is large, the position of the error messages is shifted.
I think there should aligned regardless its width.

My guess is that because MaxWidth=250 of TextBlock of MaterialDesignValidationErrorTemplate.
Is there any way to aligned position of the error messages?

![error_message_positions](https://cloud.githubusercontent.com/assets/11419109/14603191/4f5a5ed6-05a7-11e6-8b17-ff1fde94fd66.png)

ButchersBoy:
(Notes: Probably need to bind the MaxWidth of the validation content to the ActualWidth of the target control, or similar, and align left.)

ButchersBoy:
Hi @lobin-z0x50 , please close if fixed after @mgnslndh PR.

lobin-z0x50:
@ButchersBoy Yes sir!

lobin-z0x50:
:+1: verified, thanx.

PR #346, opened 2016-04-19
Fixes alignment of error message in validation template #343

mgnslndh:
Initially I also had a change for the demo project that demonstrated that this works. But I left it out since it did not feel important for the overall purpose of the demo. 

Anyway, works on my machine. Perhaps @lobin-z0x50 can verify this?


lobin-z0x50:
@mgnslndh Thank you so much!!
I will try it later, please wait.

ButchersBoy:
Fixes #343

PR #353, opened 2016-04-23
Add PopupOpened/PopupClosed events to PopupBox

gen3ric:
I noticed that PopupBox was missing this event, which appears on the WPF Popup control. It doesn't pass through any special information so it utilizes normal `RoutedEventArgs`, and uses `Bubbling` routing strategy.


ButchersBoy:
Hi @gen3ric , to just complete this can you add the standard pattern "On*" event raisers, e.g. DialogHost.OnDialogClosing

gen3ric:
@ButchersBoy Sorry for the delay. I made the change you requested. Let me know if I missed anything.

ButchersBoy:
Hi, sorry, I didn't mean quite like this.  The standard pattern in .net is for a ```protected ``` On' method which is an invocator, _not_ the name of the event itself.  This pattern is still pretty much followed with routed events, and is how other events in this library are raised...

Issue #356, opened 2016-04-24
Clicking outside of an opened PopupBox will not close it


mgnslndh:
1. Open the popup and use the slider.
2. Click outside of the Grid (popup content)
3. Popup does not close as expected

```
<materialDesign:PopupBox StaysOpen="True" Margin="16,16,16,16">
   <Grid Margin="16,16,16,16" Width="200">
      <Grid.RowDefinitions>
         <RowDefinition />
         <RowDefinition />
      </Grid.RowDefinitions>
      <TextBlock Grid.Row="0" materialDesign:HintAssist.Hint="MACHINE SPEED">200</TextBlock>
      <Slider Grid.Row="1" />
   </Grid>
</materialDesign:PopupBox>

```

PR #358, opened 2016-04-25
Close popup on outside click when StaysOpen=True

mgnslndh:
This is a fix for #356. I've also added a demo of a StaysOpen=True PopupBox.


ButchersBoy:
sweet

PR #376, opened 2016-05-12
ComboBox uses VisibleWidth

l1pton17:
I added the GetVisibleWidth method that returns a visible width of an
element. So now ComboBoxPopup correctly covers an input control.


ButchersBoy:
Nice :-)

Issue #328, opened 2016-04-05
Make it possible to disable ripple effects


mgnslndh:
Add a property to RippleAssist that can be used to turn off ripple effects. Controls can query this property to figure out if they should render a ripple or not.

Possible names for the property:
- IsRippleEnabled
- IsRippleDisabled
- UseRippleEffect

Keboo:
Should this issue be closed? It looks like the PR has been merged.

jmwhitmire:
It seems that something has broken this fix. I cannot remove the Ripple from a CheckBox in a ListBoxItem, and the Ripple effect gets applied to the entire ListBoxItem, not just the CheckBox.

PR #387, opened 2016-05-27
Added enums for ripple effect #328 

CarHel:
Added this enum on RippleAssist

`    public enum RippleEffect
    {
        Standard,
        Centered,
        None
    }`
and made according chages for it to work

### DIFF
--- a/MaterialDesignThemes.Wpf/RippleAssist.cs
+++ b/MaterialDesignThemes.Wpf/RippleAssist.cs
@@ -3,12 +3,6 @@ using System.Windows.Media;
 
 namespace MaterialDesignThemes.Wpf
 {
-    public enum RippleEffect
-    {
-        Standard,
-        Centered,
-        None
-    }
     public static class RippleAssist
     {
         #region ClipToBound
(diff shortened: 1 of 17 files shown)


ButchersBoy:
OK, big problem here is this is a breaking API change.  Now, we are following semver, and will soon be moving to ver 2 as we have some breaking changes.  See https://github.com/ButchersBoy/MaterialDesignInXamlToolkit/wiki/Version-2.0.0---Breaking-Changes

However, breaking changes is something I want to avoid unless it is really warranted to push the library forward.

So what are the other options for RippleAssist?

CarHel:
Alright, didn't thought about that. I don't know how to revert changes, so i just copied back older versions and made the new changes.

MaterialDesignThemes.Wpf/RippleAssist.cs
```diff
@@ -54,6 +54,35 @@ public static bool GetIsCentered(DependencyObject element)
...
+        /// </summary>
+        public static readonly DependencyProperty IsRippleDisabledProperty = DependencyProperty.RegisterAttached(
+            "IsRippleDisabled", typeof(bool), typeof(RippleAssist), new FrameworkPropertyMetadata(false, FrameworkPropertyMetadataOptions.Inherits));
+
+        /// <summary>
+        /// Set to <c>True</c> to disable ripple effect
+        /// </summary>
+        /// <param name="element"></param>
+        /// <param name="value"></param>
+        public static void SetIsRippleDisabled(DependencyObject element, bool value)
```
ButchersBoy: Can we drop the "Ripple";  we don't need the noun as we are already in the RippleAssist class, and it  keeps it inline with IsCentered.  So, "IsDisabled".

CarHel:
I made the changes, I'll try to be more careful next time. I don't understand what are the branch conflicts :/

ButchersBoy:
re. merge conflicts, I did some changes on demo Buttons.xaml, so without me pulling your latest change I'm guessing around there.  

So, you need to figure out how to update your fork from master, fix the conflicts, and push the fixes back to GitHub.  Then we should be pretty much ready to pull into master.

CarHel:
Guess i found out :D.

ButchersBoy:
...your didn't need to close the PR...unless you find it easier starting clean

PR #399, opened 2016-06-04
Snackbar

spiegelp:
implementation of a Snackbar according to the specs on https://www.google.com/design/spec/components/snackbars-toasts.html


ButchersBoy:
Hi @spiegelp I've created a new branch, "snackbar".  Can you re-raise this PR against that branch?  I want to suggest/work on some changes so we get the API right before we pull it into master.

PR #400, opened 2016-06-05
ComboBox background

### LINKED ISSUE #369: ComboBox background color should allow transparent
The new version of Material Design XAML Toolkit applies a background color to the ComboBox control which can be an issue when displayed in a ListView which has alternating row colors: 

dr

It would be great to have a way to toggle the background to transparent on ComboBoxes, so the background color is not applied. Thanks.

l1pton17:
Fixes #369 

Now combobox popup's background is dependent on actually background color of combobox.
It means, if you set background of combobox or just place it into border or smthng else with non-null background, it will use the correct background brush.

By default, combobox's background is transparent now.
For example, if you put combobox into a border with red background, combobox's popup will use red background.

BTW, i had to change ScrollBar's background from MaterialDesignPaper to transparent (comboboxpopup with non-default background didn't combine with scroll bars).


MaterialDesignThemes.Wpf/ComboBoxPopup.cs
```diff
@@ -1,14 +1,21 @@
...
 using System.Windows.Controls.Primitives;
+using System.Windows.Data;
 using System.Windows.Media;
 
 namespace MaterialDesignThemes.Wpf
 {
     public class ComboBoxPopup : Popup
     {
+        public const double ContentPresenterExtend = 4;
+        public const double TopBottomMargin = 8;
```
ButchersBoy: Anything we can do to have these 3 constants in XAML?

l1pton17: @ButchersBoy move these back to xaml? lol, i moved them to source code because i thought that i need them in code, but really i didn't. Can move it back if you want.

ButchersBoy: Yes please...if we don't have to make them part of the public API I'd rather that...in case someone becomes dependent on them somewhere down the line.

l1pton17: @ButchersBoy done

l1pton17:
@ButchersBoy done

PR #404, opened 2016-06-08
Snackbar

spiegelp:
implementation of a Snackbar according to the specs on https://www.google.com/design/spec/components/snackbars-toasts.html


ButchersBoy:
OK, initial impressions:

Looks nice, animations are nice, but I think we need to ditch the SnackbarHost.

I know we have those for Drawers and Dialogs but they serve the special purpose of fading and disabling the underlying content.  Also, the dialog host API is very specific to the challenge of dialogs withing WPF, so I think we need to make this more of a "classic" control API which works in code-behind and via binding/MVVM.  As it stands I think the API is a little too prescriptive.

I think the next steps should be something like:

* Drop the 'Host.  We can just style it to appear vertical bottom horizontal center, then a user can put it in any grid.
* It should from ContentControl (I appreciate the guidelines say text only, but we should also follow WPF standards for implementation)
* We need some kind of mode: Open,Closed, Automatic
  * Automatic should auto show and hide every time the content updates
  * This API will make it more "standard", less prescriptive and reduce the extra Host logical and visual tree element
  * I guess this would combine the behaviour of of .State and .Show()
  * We _could_ still keep Show Hide, to cycle through above behaviour...Although I'm tempted to consider moving this to some sort of optional controller...lets look at this afterwards when we see the refreshed API and simplified logical tree.  
* To expose the button behaviour we should add Command, CommandParameter, CommandTarget dep props, which are pretty much standard.  The we can also expose a ButtonClick, so again we have a standard API for both code behind and binding.
  * This would be in place of ActionHandler
* In OnApplyTemplate, make sure you unsubscribe from previous elements, if client code changes templates, we have a memory leak.  Checkout Chip for an example.

spiegelp:
My idea behind the host component was simplicity for the implementation:
1. Snackbar itself was considered for a single use and to throw it away afterwards. Memory leaks cannot occur as there will not be any reference to the Snackbar anymore and the garbage collector will do the rest. 
2. The code in the Show() method is a little less complex if a new Snackbar should replace an already visible one. The state management will be more simple.

I know that the API is restrictive at the moment. It is just my exact understanding of the specs. And if you know that the user can do something wrong, you can also prevent him from doing that ;-)

The idea of a mode is interesting. I was thinking of adding some overloads to add a custom timeout for showing a Snackbar. Something like the manual mode is not mentioned in the specs so I did not consider it in the implementation. On the other hand I like APIs with different overloads. I just had to implement a common use case at work because the API of a third party library did not provide it. 

Will you do the changes or should I implement them?

SCKelemen:
Is there any estimate on when this will be added? 

I agree, it should inherit from contentcontrol, implement commands, implement command parameters, and have an enum for state. I think fading transitions should be included too.

ButchersBoy:
I've scheduled it for 2.1.

Still got a lot of tidying up for 2.0, and I want to incubate the API and have a think about it a bit before it hits master.

ButchersBoy:
@spiegelp if you could do the following then I'll accept the PR into this branch: 

* Inherit from ContentControl
* Delete the 'Host
* Have a basic bool IsOpen which triggers your animations
* Remove the .Show() and timer stuff
* Add the Command, CommandParameter, CommandTarget depProps for the button.

Then we can reassess,  particularly around the Show()/timer stuff when we start looking at 2.1.

spiegelp:
I already started with the refactoring yesterday evening. The triggers already work by changing the Content property. However the code is getting much more complex than the initial implementation with the host. I will see what comes out in the next days.

ButchersBoy:
We should be making it simpler...Not more complex...if the Show method and timer and host is removed, So we just go back to the basic control there should be much less code. Then we can figure out and build up the rest later

spiegelp:
Throwing the visual states did the trick. They blew up the code with complexity to getting the states work. 

The current implementation state is:
- the action raises a well-tried event and additionally calls the optional command from ActionCommand property
- the action button is available by providing content for the ActionLabel property
- the IsOpen property controls the visibility of the Snackbar
- Mode == SnackbarMode.HalfAutomatic enables the automatic show and hide behaviour defined in the Material Design specs by setting the Content property (I'm too lazy for the bad practice of reimplementing the timer in each and every application)

Check it out!

spiegelp:
Four months are enough time to read the API docs and to incubate new ideas. I moved the animations from the XAML into the code. This allows more control on the animations and setting a Completed event handler which will close the Snackbar in HalfAutomatic mode. 

And the best of all: No more need of the timer!

ButchersBoy:
I've got issues over a year old I still haven't tackled, so no, four months isn't enough.

I preferred the animations in XAML.  I think the control needs to be as vanilla and as simple as possible.  You have to consider a busy system where people might start throwing batches of notifications at it, and thus we have to consider threading and queueing and batching.  Which is always risky code.  I'd rather pull this stuff out into some sort of VM/Controller which people can optionally use (if we even supply that as part of the lib at all). In which case I can't see much justification in having the animations in code.  Going back on my previous remarks, I think we simplify things further - at least for now - with just an IsOpen flag. A simple control can make master branch before a more complex and risky controller.  It's my job to safe guard the API (which is just as -if not more- important as the aesthetics), so until I have some clear time and some inspiration to work on this we may have to wait a bit longer.  I'm currently working on something else, which I intend to use the snackbar in, so at this point I will assess the control/timing issue thoroughly.  But at the moment this control is part way down a very long work list of tasks I am working through.

spiegelp:
The four months refer to my experimental phase. I had serveral ideas based on other stuff that I wrote or saw but each one had a drawback, strange side effects or even did fail (exceptions by freezed resources, never ending loops in triggers, broken bindings after one call, many code...). Implementing each detail of the specification pushes WPF to its limits for pretty code. The current Snackbar has probably the smallest drawback. 

I would strongly prefer the animations in the XAML as they are very UI specific. However combining them with the behaviour of the specification caused side effects. The current Snackbar contains some UI logic in the code (I usually refuse to write such code) but it gets the job done. The specified behaviour requires this "ugly" code in any solution. Some async code done by storyboards, own threads or something else will be always needed. Using a view model or a controller might move such code to another place. It is a kind of personal preference in which class such code should go to. Moving to a controller might be an overhead and (drawback's greetings) probably causes that the Snackbar will not be fully configurable and controllable in XAML. 

The intention of also implementing the behaviour is that it will be implemented once instead of each time a Snackbar should be shown in an application. Moreover these small details enhance the user experience. They are the joy of use. It is the user centered thinking of mine. A user does not care about "ugly" code. He/she needs an appealing UI that gets the job done, although it is a tough task for a programmer in some cases. 

The Snackbar is not suitable for batches of notifications by specification. A Snackbar following the specifications just throws old notifications in favour of the newest one. This use case requires another concept or intended control. Maybe some kind of container which creates a new Snackbar for each notification and stacks them. That is an argument for introducing a controller for the basic behaviour and a special one for this use case. 

I will move the animations back to XAML. The behaviour stuff will get a new concept on a different branch for future reference.

ButchersBoy:
Sorry, sorry, sorry, I thought you were being impatient! 😳

OK, get the XAML animations put back in and I'll take another look.  I will will a copy into another of my applications and have a play around with methods of calling it etc.

👍

EdonGashi:
It would be desirable to fire and forget messages without worrying about queuing or timing, so I agree that the control should handle those aspects.

ButchersBoy:
@EdonGashi yes, the final API will support that

EdonGashi:
Hello, I took @spiegelp 's implementation and tweaked it a bit and added queueing support in automatic mode.

[Check it out](https://github.com/EdonGashi/Snackbar) and see if you can salvage anything out of it.

It's in one take so it might have bugs.
Thanks!

ButchersBoy:
Please be aware that the snackbar branch is active and is currently being worked on...

EdonGashi:
Should I send a PR there with separate classes for experimentation?

I needed to use a quick implementation in a project, that's why I put it inside a new library.

ButchersBoy:
You might want to wait a little cos I'm hacking around with it currently.

spiegelp:
Good that I had much work in the garden last weekend. So I did not start a third implementation full of merge conflicts. 

My ideas so far: I was thinking of changing the Snackbar to a control with less logic. Just IsOpen and a trigger for the animations. For timing, queuing or stacking I would introduce a view model and a controller which handle the neccessary logic with bindings and events. As the logic for these features blows up the Snackbar API, new specific controls might be a solution. They will have the specific API for their feature and can use the "stupid" Snackbar internally. 

I hope that these ideas help you a little bit for your hacking.

ButchersBoy:
Ha @spiegelp , I  am the same...gotta get the garden ready for winter :)

SCKelemen:
#404 Snackbar Not Found.

I agree that some of this functionality should be in a ViewModel or Controller.

ButchersBoy:
Locking this conversation as it's a merged and closed PR.  For further updates on snackbar see https://github.com/ButchersBoy/MaterialDesignInXamlToolkit/issues/390

Issue #379, opened 2016-05-18
TextBox border and floating hint colors don't follow Material Design guidelines


redcurry:
If you look at the official [Text fields](https://www.google.com/design/spec/components/text-fields.html) guidelines, the TextBox underline and floating hint are lighter than the foreground color (for the Light theme). Unfortunately, they don't say exactly what the opacity should be, but in all their images they're lighter. I've been able to fix the underline color by overriding the MaterialDesignTextBoxBorder (I set it to #20000000). But I haven't been able to customize the floating hint color since it is set to the body foreground color. Its opacity should probably be set to HintOpacity.

EdonGashi:
Looking at SmartHint.xaml it seems like it already has its opacity set to HintOpacity, I think the visual state animation overrides it to 100% which makes it equal to the default body color:

```XML
<DoubleAnimation Duration="0"
    Storyboard.TargetName="FloatingHintTextBlock"
    Storyboard.TargetProperty="Opacity"
    To="1" />
```

Keboo:
I believe this has been addressed in the latest 4.2.1 release

PR #405, opened 2016-06-09
Floating hints animate to HintOpacity instead of 100%

EdonGashi:
Non empty visual state of SmartHint will animate to the value of HintOpacity property instead of 100%, which will cause floating hints to maintain color while moving upwards.

As noted in #379 and clearly seen in the [spec](https://material-design.storage.googleapis.com/publish/material_v_8/material_ext_publish/0Bzhp5Z4wHba3eC03VnhOekZxWW8/components_textfields_counter3.png) hints have less opacity than the control content.

Because focused controls override the floating hint color to the PrimaryHueMidBrush, an override of HintOpacity to 100% is added on their data triggers.

Additionally, TimePicker was missing such a trigger, which I have added in this PR.

Screenshot comparison below:

compared


ButchersBoy:
Hi @EdonGashi, thanks for this,  looks good; one small bug:

non-editable combos dont pick up the accent colour when they have focus (all  other controls work fine):

image

EdonGashi:
I am looking into it. I will update once I find a solution. Thanks

EdonGashi:
I downloaded the latest master and this bug occurs even without my updates. Set the primary color to yellow or something distinct as purple is hard to notice. It seems that changing the combo's selection once or twice fixes it afterwards...

ButchersBoy:
OK, I'll pull in the PR...if you find a fix in the future another PR would be welcome.

ButchersBoy:
Thank-you, and welcome to the contributors list!

redcurry:
Looking at another theme, however, (see https://material.google.com/components/text-fields.html#text-fields-labels), the hint and underline colors may be set to something else, like an accent color.

PR #418, opened 2016-06-16
TimePicker SelectedTime setting value to null fix

jeffklawiter:
I observed and was able to identify that the **SelectedTime** property on **TimePicker** will fail to display the initial value of a property that is bound to it. Upon initial load of the control WPF will set up the bindings before **OnApplyTemplate** is called. So while the control is being created the bindings for **SelectedTime** are applied. This kicks off a call **SelectedTimePropertyChangedCallback** with the value evaluated from the bindings. This in turn makes a call to change the **TextProperty** value. When **TextProperty** changes it calls **SetSelectedTime()**. **SetSelectedTime()** will in then change the SelectedTime property based on the **Text** of the **_textBox** that is retrieved from the control template. Since this does not yet exist it evaluates to empty. Since it evaluates to empty **SelectedTime** is set to **null**. Which kicks this chain off again. The end result being the **SeletedTime** value and the **TextProperty** value are now null and the _textBox.Text value ends up being null. Thus not displaying the time to the user.

This fix is to only call **SetSelectedTime()** from within the **TextPropertyChangedCallback** if the **_textBox** control is not null. This callback already had this code in there for updating the .Text property of the **_textBox**. 

An alternative possible fix would be to update **SetSelectedTime()** to do an additional check to see if **_textBox** control is not null and only set **SelectedTime** if the control exists.

This PR also includes an example of the binding working in the demo project.

### DIFF
--- a/MainDemo.Wpf/Pickers.xaml
+++ b/MainDemo.Wpf/Pickers.xaml
@@ -45,10 +45,15 @@
             <DatePicker Name="LocaleDatePicker" Width="120" HorizontalAlignment="Left" Margin="0 16 0 0" materialDesign:HintAssist.Hint="Locale Date" />
             <DatePicker Name="LocaleDatePickerRTL" Width="120" HorizontalAlignment="Left" Margin="0 16 0 0" materialDesign:HintAssist.Hint="RTL Locale Date" FlowDirection="RightToLeft" />
         </StackPanel>
-        <materialDesign:TimePicker Grid.Row="1" Grid.Column="1"  VerticalAlignment="Top"  Width="100" HorizontalAlignment="Left" Margin="0 16 0 0"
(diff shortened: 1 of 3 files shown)


jeffklawiter:
I was curious if there was anything else needed for this? This bug is preventing me from using the nuget package, instead I'm having to use my own fork.

ButchersBoy:
Hi @jeffklawiter I had a quick look at it the other day, but wanted to come back and look at it again.  The reason being is that it is changing the API, and the API is modeled on the standard WPF DatePicker to maintain consistency.  So I wanted to have a closer look to see if we can make the fix without making an API change.

ButchersBoy:
Hi @jeffklawiter 

I was able to reproduce your issue.  I've implemented an alternate fix that doesn't change the API.

PR #424, opened 2016-06-25
Add ComboBoxAssist.ShowSelectedItem attached property

l1pton17:
This attached property add ability to change visibility of selected item in combobox


ButchersBoy:
Sorry for delay :/    ...been mega busy

PR #425, opened 2016-06-26
Add Expander.ExpandDirection support

l1pton17:
Now Expander supports all ExpandDirections


spiegelp:
You solved issue #449 before it was asked. +1

I am looking forward for this merge, as I have local changes with an additional style targeting the Expansion panels specification. I was half a year earlier than the Material design team when I started with the first Expander template. Now I have twice the work because of them ;-)

PR #428, opened 2016-06-29
Explicit styling of PoupBox used as a Window Command in MahApps

mgnslndh:
Explicit style for material design `PopupBox` when used in MahApps `MetroWindow` `LeftWindowCommands` and `RightWindowCommands`

Basically makes  the PopupBox look the same (opacity 50%) as other buttons. 

Would be nice if the PR could be improved to style the PopupBox implicitly when used as a window command, just as with buttons that do not need the explicit style.

animation


ButchersBoy:
Thanks.  This package is manually deployed...I'll wait for @punker76 to look at the referencing/paket stuff before deploying the next version to nuget.

PR #430, opened 2016-06-30
Use paket manager instead nuget

punker76:
This PR replaces nuget (incl. restoring) with [paket](https://fsprojects.github.io/Paket/index.html)

- no more version numbers in project files (so other libraries can test pre-releases of third part dlls)
- faster restoring
- makes you happy :-D

Jan


ButchersBoy:
Wow, all  change :)

PR #463, opened 2016-08-11
Stepper

spiegelp:
Implementation of a horizontal and a vertical Stepper. Not all tiny details of the specification are implemented yet but may follow.


gitAllender:
Hi there! Is it some how possible to priories this pull request. Looks like a very useful addition. Thx.

spiegelp:
@gitAllender Feel free to check out the source code and do your own build. I already use the Stepper this way.

MaterialDesignThemes.Wpf/Stepper.cs
```diff
@@ -0,0 +1,345 @@
...
+        }
+
+        private void CancelHandler(object sender, ExecutedRoutedEventArgs args)
+        {
+            if (args.Handled)
+            {
+                return;
+            }
+
+            args.Handled = true;
```
CookiesAndCoffee: I used the CommandManager to listen to the CancelCommand, but the event never reached my code and i think it is because of this line. If the args are set handled = true, it will no longer bubble up the tree.
In case that i am using it wrong, feel free to explain how it should be use and maybe add it to the StepperDemo.

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Stepper.xaml
```diff
@@ -0,0 +1,451 @@
...
+                                                <ColumnDefinition Width="60" />
+                                                <ColumnDefinition Width="100*" />
+                                            </Grid.ColumnDefinitions>
+                                            <Grid.RowDefinitions>
+                                                <RowDefinition Height="auto" />
+                                                <RowDefinition Height="100*" />
+                                                <RowDefinition Height="8" />
+                                            </Grid.RowDefinitions>
+                                            <!-- header as button to switch steps in non-linear steppers -->
+                                            <Button Grid.ColumnSpan="2" Command="{x:Static wpf:Stepper.StepSelectedCommand}" CommandParameter="{Binding}" Style="{StaticResource MaterialDesignStepButton}">
```
CookiesAndCoffee: The hover was a little bit weird for me so i looked it up in the Material Guide Lines. The Icon have to use a 8dp top and bottom padding.

spiegelp: @CookiesAndCoffee I just fixed the headers in vertical steppers to match the specification.

MaterialDesignThemes.Wpf/Stepper.cs
spiegelp: @CookiesAndCoffee Thank you for your feedback. The commands were intended to control the `Stepper` using navigation buttons. I thought that it is not necessary to let them bubble up. The use case of the cancel button is not eleborate so well.

I just committed a change the let the commands bubble up. Furthermore I added some events to get notified if the `Stepper` starts to navigate to another step. These new events may help you out. 

I hope that the changes will fit your requirements. Let me know if you recognise other bugs, strange behaviour or not pixel-perfect UI elements.

EdonGashi:
Hey @spiegelp. I'd love to use this feature. Any chance that you publish this as an independent project? It's also fitting to be separate as it has quite the behavior for a theme library. If you've abandoned it I can start a fork.

wongjiahau:
I want this feature

spiegelp:
@EdonGashi Hi, I never thought of publishing an independent project. I just want to share the Stepper with others by donating it to the original open source library. We are going to get a great library by sharing all our ideas and improvements into one project rathern than doing it all alone by ourselves.

At the time I would consider this version of the Stepper as feature complete. There are many more details in the specification for Material Design, but a big part of them consider mobile interfaces. It is quite enough for a desktop UI. At the moment I am not working on the Stepper as it fits my needs. However I might pick it up in the future and will consider every upcoming feedback. You are welcome to do some improvements and it would be great if you merge your code back into my branch. 

**At everyone who like the Stepper: If you cannot wait anymore, feel free to check out the source code and do your own build!**

EdonGashi:
I understand. We will contribute to your branch in the meantime. Thank you.

wongjiahau:
@spiegelp I had tried out the steppers, however I think there are several issues : 
- first, the transition animation is quite confusing when the steppers is in Vertical layout, my suggestion is perhaps not having the animation is better 
- second, I thought the current way to use this Stepper are not following the common WPF convention, IMHO it should resembles the following : 
```XML
<Stepper Layout="Vertical" IsLinear="True">
    <Step  Header="Step 1 : Login to your account"
                Subheader="You can either login using Google or Facebook"
                Continue="Next">
	    <Page_Login/>
    </Step>
    <Step Header="Step 2 : Enter your password" Back="Login with another email">
	    <Page_EnterPassword/>
    </Step>
</Stepper>
```

ButchersBoy:
Hi guys.

I do think, in it's current state, this would be nice in a separate project/package.

As regards the control implementation itself, it's basically a tab control.  So could either be a style of a TabControl, or a custom control; but it should most likely derive from Selector, to keep in with standard WPF controls.

If it were to wind up in this toolkit I'd like to see it along those lines.  But the difficulty is that it introduces several view models/interfaces.  Typically I've kept view models out of the library.  The library is controls and themes, and agnostic to whether you are using MVVM or not.  I appreciate that people want readily available code they can use, but I dont want this library to become to prescriptive on how it should be used.

EdonGashi:
+1 for the tab control idea. The control should be as dumb as possible. ItemsSource and SelectedIndex-Item is all we need regarding its behavior.

spiegelp:
The idea of a Stepper as specific UI for a TabControl is OK. However if it comes to behaviour, a Stepper is far beyond a TabControl and the specific kinds (i.e. linear Stepper) do not behave like a TabControl is supposed to. 

I understand that the pull request heavily introduces new models and interfaces. However it is a clean solution to realise all the specific and complex behaviour of a Stepper. It is based on WPF's strengths in MVVM and templating. All these small features and the overall behaviour are what Material Design is all about to target the joy of use. As a lazy developer, I would like to write code once and reuse it in every new project. That is why I invested a lot of time to create a generic Stepper with as much behavior features as possible. 

@ButchersBoy I could start a complete new implementation by inheriting TabControl on a new branch. That Stepper would be only a basic UI with much less behavior. The view models will be gone but everyone has to do his/her custom implementation for the behavior. No rules on how to use a dumb control. 

For everyone looking for a fully featured Stepper, I will keep the original Stepper in my fork.

spiegelp:
After thinking about the whole situation during the night, I got an idea how to combine both solutions: 
* Stepper inherits TabControl to expose its API
* the model stuff will be kept as internal helpers, which will be only visible for the assembly itself
* Stepper might add some optional features to TabControl

This way the user gets a TabControl. He/she can use the Stepper as a simple TabControl or optionally use the additional features. What do you think?

spiegelp:
The Stepper now inherits from TabControl: 
* It takes a collection of `TabItem` as steps and converts them internally to the view models to enable the rich UI and behaviour. 
* The behavioural features of semantic data validation are lost. This will require a custom implementation for every application by using the pre-navigation events of the Stepper. 
* It looks like a styled TabControl by setting `IsLinear = false` and `ContentAnimationsEnabled = false`.

wongjiahau:
@spiegelp That's great!

MaterialDesignThemes.Wpf/Stepper.cs
```diff
@@ -0,0 +1,352 @@
...
+            _controller.PropertyChanged -= PropertyChangedHandler;
+        }
+
+        private void InitSteps()
+        {
+            List<IStep> steps = new List<IStep>();
+
+            foreach (TabItem tabItem in Items)
+            {
+                steps.Add(new Step() { Header = tabItem.Header, Content = tabItem.Content });
```
CookiesAndCoffee: Line 190 is preventing my own IStep implementation.
` steps.Add(new Step() { Header = tabItem.Header, Content = tabItem.Content });`

I tried to add `public class MyStep : TabItem, IStep`.

MaterialDesignThemes.Wpf/Stepper.cs
```diff
@@ -0,0 +1,352 @@
...
+    {
+        Horizontal,
+        Vertical
+    }
+
+    /// <summary>
+    /// The argument for the <see cref="Stepper.StepValidation"/> event and the <see cref="Stepper.StepValidationCommand"/> command.
+    /// It holds the <see cref="IStep"/> to validate.
+    /// </summary>
+    public class StepValidationEventArgs : RoutedEventArgs
```
CookiesAndCoffee: Validation has been removed.

spiegelp: @CookiesAndCoffee I recently decided to start a new project for not specified or scope-crashing controls. Have a look at [Material Design Extensions](https://github.com/spiegelp/MaterialDesignExtensions). It contains the old `Stepper` before the refactoring.

CookiesAndCoffee: @spiegelp This is a great solution while this pull request is work in progress.

Keboo:
@spiegelp any objection to closing this PR? I think for now having it as a separate library works well. I also [added a link to your repository in the README](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/commit/af4b50a25041b5c77eca529af88ae4371bc48109#diff-04c6e90faac2675aa89e2176d2eec7d8).

spiegelp:
@Keboo I don't mind. The stepper component disappeared as Google published the new Material Design specification earlier this year. [Material Design Extensions](https://github.com/spiegelp/MaterialDesignExtensions) will be a better now.

PR #464, opened 2016-08-12
treat null values in MathMultipleConverter

spiegelp:
- `Binding.DoNothing` instead of running into a `NullReferenceException`
- optionally use 0.0 if a value is null by setting `TreatNullAsZero == true`


MaterialDesignThemes.Wpf/Converters/MathMultipleConverter.cs
```diff
@@ -12,11 +12,61 @@ namespace MaterialDesignThemes.Wpf.Converters
     public sealed class MathMultipleConverter : IMultiValueConverter
     {
         public MathOperation Operation { get; set; }
-        
+
+        /// <summary>
+        /// if true, null arguments will be treated as 0;
+        /// default: false
+        /// </summary>
+        public bool TreatNullAsZero { get; set; }
```
ButchersBoy: Are we actually using this, anywhere?  It would become part of the public API, so has to justify itself.

ButchersBoy:
Went for  the simpler option in then end.

spiegelp:
I was planning to use this as soon as it is merged. However I had no time to react this morning, because I do not check my private mails at the office. So this will be another changed class in my own build.

PR #469, opened 2016-08-23
Scaling size and translating position of floating hint

mgnslndh:
I've made a spike implementation of the functionality described in PR title. 

Reason behind PR is that I need small floating hints to go with text boxes with large font size. 

Please check it out. I have not made visual comparison between old and new implementation but will do so later.

There might be some code that i've commented out that needs to be fixed as well.


mgnslndh:
# New Implementation

enhanced

# Old Implementation

old

mgnslndh:
One problem with the current implementation is that if you have a hint that does not fit into the visible text box it will also be truncated when rendered as a floating hint above the text box. This behaviour can be seen in the gif above.

By using the LayerTransform for scaling it will render correctly but then we need other other offsets and i'm not sure how LayerTransform affects other elements.

ButchersBoy:
:+1:

PR #474, opened 2016-08-26
Foreground color updates header and icon

jherbert:
Commit 4cf183810f726d1b4da46023adf7a01d462bddc2 updated expander to use PackIcon, pull request so foreground is reflected both on header text and icon.

### DIFF
--- a/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Expander.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Expander.xaml
@@ -70,7 +70,7 @@
                             </VisualStateGroup>
                         </VisualStateManager.VisualStateGroups>
                         <wpf:PackIcon Kind="ChevronDown" Opacity=".38" x:Name="ExpandPath" RenderTransformOrigin="0.5,0.5"
-                                      Height="24" Width="24">
+                                      Height="24" Width="24" Foreground="{Binding RelativeSource={RelativeSource AncestorType={x:Type Expander}}, Path=Foreground}">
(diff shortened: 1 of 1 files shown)


MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Expander.xaml
```diff
@@ -70,7 +70,7 @@
                             </VisualStateGroup>
                         </VisualStateManager.VisualStateGroups>
                         <wpf:PackIcon Kind="ChevronDown" Opacity=".38" x:Name="ExpandPath" RenderTransformOrigin="0.5,0.5"
-                                      Height="24" Width="24">
+                                      Height="24" Width="24" Foreground="{Binding RelativeSource={RelativeSource AncestorType={x:Type Expander}}, Path=Foreground}">
```
ButchersBoy: Is a TemplateBinding not sufficient here?

jherbert:
Updated to use `TemplateBinding` for the foreground on each style.

md_expander_ss

ButchersBoy:
Hi, please see my code review comment, I don't think we need the relative source binding as you are setting the toggle foreground, and the binding is in the toggle template.

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Expander.xaml
ButchersBoy: @jherbert

jherbert:
Sorry you are correct. Had it that way locally but failed to push it up.

ButchersBoy:
Thankyou!

PR #482, opened 2016-08-31
Separator on colored background

mgnslndh:
A light and a dark separator style for use on colored backgrounds. Added cards to show off the new styles.


ButchersBoy:
Looks cool :+1: , I'm just going to move the new styles into the .Defaults.xaml file, as this is the only common, "core" file, and the colours don't have to be specifically used with Cards.

PR #483, opened 2016-08-31
Flat toggle listbox

mgnslndh:
`MaterialDesignFlatToggleListBox` is a new style derived from `MaterialDesignFlatToggleListBox` but with a `ShadowAssist.ShadowDepth` of `Depth0`


MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.ListBox.xaml
```diff
@@ -192,6 +193,10 @@
         </Setter>
     </Style>
 
+    <Style x:Key="MaterialDesignFlatToggleListBox" TargetType="{x:Type ListBox}" BasedOn="{StaticResource MaterialDesignToolToggleListBox}">
```
ButchersBoy: I think we should call this ``` MaterialDesignToolToggleFlatListBox ``` to better indicate it as a variant of the underlying style.

Verbose, I know :)

mgnslndh: Probably a good idea! Fixed.

PR #486, opened 2016-08-31
Expander content alignment

### LINKED ISSUE #478: Expander content does not stretch
The `StackPanel` that displays the content in an Expander has its alignment forced to "Left", so there's no way to make the content stretch; shouldn't this be bound to the Expander's `HorizontalContentAlignment`?

mgnslndh:
Propagates the HorizontalContentAlignment property from the Expander to its internal content panel (StackPanel).

Fixes #478


ButchersBoy:
Thanks for this, did it slightly differently in the end.

PR #501, opened 2016-09-18
Expansion panel details

spiegelp:
- removed divider between header and content (they are not here in the specs)
- inserted dividers in the demo between the expanders (here they are in the specs)
- altered margins to match the specs a little bit better


ButchersBoy:
Thank  you!

PR #506, opened 2016-09-24
Update Flyout style to MahApps v1.3

punker76:
Just noticed an incompatibility with the new v1.3

This should it fix.

matdesign

/cc @ButchersBoy


punker76:
@ButchersBoy done

ButchersBoy:
Thanks!

PR #509, opened 2016-09-30
Allow user to customize Expander Background Color

abhall724:
Passes the background color to the template for the Expander Theme


ButchersBoy:
Thank you!

PR #521, opened 2016-10-21
Changed dark theme background to match guidelines

EdonGashi:
Changed `MaterialDesignPaper` color resource to match the recommended value specified in https://material.google.com/style/color.html#color-themes

`MaterialDesignDarkBackground` has also been changed to #FF303030

Because cards are defined to have a different color than the background, a new resource named `MaterialDesignCardBackground` has been added. Card template now uses this resource for its background.

The guideline says that light theme cards should have a #FFFFFF value, but that needs to be discussed so I didn't set that yet.

We shouldn't merge these changes without first considering its effect on existing controls.

ComboBoxes, Calendars, TimePickers and some other popups would likely need to use the new card background color.

### DIFF
--- a/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Card.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Card.xaml
@@ -35,7 +35,7 @@
     </ControlTemplate>
     <Style TargetType="{x:Type wpf:Card}">
         <Setter Property="Template" Value="{StaticResource CardTemplate}" />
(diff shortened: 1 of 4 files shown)


ButchersBoy:
Thanks...I've updated light card background too.

EdonGashi:
I found a sample of comboboxes using a FFFFFF background in light theme. Couldn't find any example of dark theme though, but most likely it too uses the card background instead of paper. 

combobox

I wanted to apply these changes in a PR but the combobox popup has some weird background inheriting I don't fully understand... Could I have some insight on this? Can we set a color resource for the popup but not the combobox?

PR #528, opened 2016-10-26
Snackbar

spiegelp:
Two little changes: 
- use nameof and constants instead of string literals
- animate Opacity of the Snackbar content


ButchersBoy:
Thanks, think I lost the Opacity animation during refactoring.

PR #535, opened 2016-11-03
Net40

rrs:
Hi James,

I have updated the .net branch so it is up to date with master and then fixed any issues preventing it from working


ButchersBoy:
Wow :)

ButchersBoy:
OK, I'm not gonna review all this, I'll merge it in; you might want to then just get a clean copy and check it's OK.

PR #537, opened 2016-11-04
MaterialDesignCardBackground for ComboBox popups

EdonGashi:
Added a trigger to set combobox background to MaterialDesignCardBackground when IsDropDownOpen.

comboboxpopup

The spec doesn't explicitly specify a color for combobox popups and even google has inconsistencies. In some cases I found a 250 popup and in others a 255. But considering that the idea of material design is utilizing the Z-layer for attention direction, popups, cards, dialogs and other things that are higher in Z usually have lighter color. This effect is clearly noticeable in the dark theme.

This is more based on preference so I advise not to merge without further consideration and research. Also note the 1px bug at the right side... I will look into what's causing it. 

An animation matching the popup fade would also be better than the instant color change.

combobox

If we start this route we need to fix calendars, time pickers and some other controls too.

calendar

### DIFF
--- a/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.ComboBox.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.ComboBox.xaml
@@ -473,9 +473,6 @@
             <Trigger Property="IsEnabled" Value="False">
                 <Setter TargetName="templateRoot" Property="Opacity" Value="0.56"/>
             </Trigger>
-            <Trigger Property="IsDropDownOpen" Value="True">
-                <Setter TargetName="templateRoot" Property="Background" Value="{DynamicResource MaterialDesignCardBackground}"/>
-            </Trigger>
             <Trigger Property="IsEditable" Value="True">
                 <Setter TargetName="PART_EditableTextBox" Property="Visibility" Value="Visible" />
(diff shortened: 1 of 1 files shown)


l1pton17:
You should take into account that when the drop down can't wrap the combobox it shows separatly.
Otherwise you will get this:
image

l1pton17:
And another thing, when a combobox opens a drop down, it take a color from underlying panel in order to keep background color of the combobox equals to background color of the drop down.
But in your case, if i will place a combobox into colorzone with mode = active then set pallet to light and accent to Indigo (some color with white foreground), i will get this:
image
okay, but when i open the combobox i will get this:
image

not so good, yeah?)

and, in the end, what current version of combobox will do:
image

EdonGashi:
Hello, thank you for the observations. My solution is very naive. If the idea behind the proposal is supported we have to consider cases as these you've mentioned. I'm thinking maybe we can add an attached property so this behavior is entirely optional.

Something like `<ComboBox PopupColor="Red">`

Unless we specify popupcolor it should work as always.

EdonGashi:
I don't know why I thought it's necessary put the trigger inside the template, a trigger in an inherited style seems to work just fine. It's a minor and situational behavior that will probably be used rarely.. we can close this PR, or leave it open for discussion about the calendar and popup stuff I mentioned.

PR #542, opened 2016-11-08
Popup fix when selecting value from datepicker inside a popup

stealthcold:
Prevents popup from closing when selecting a value from DatePicker or TimePicker inside the popup


ButchersBoy:
Thank you!

PR #557, opened 2016-11-18
Added slight translucency for better look of flat buttons in dark style

mgnslndh:
### Before

before

### After

after


mgnslndh:
Perhaps the fix is too dark? Here is a second alternative which I think look better. But in this case I have also altered the background color, not only alpha.

alternative

`#19CCCCCC`

ButchersBoy:
What did I do on the snackbar?  I remember making sure it looked good in dark mode...might be good to find parity...

ButchersBoy:
:+1: thanks!

Issue #552, opened 2016-11-15
ComboxPopUp DefaultTemplate not correct


elvince:
Hi,

In your demo project, if you remove the margin "16" in the MainWindow and also in TextFields, then you add a new row in the grid, then add
`<ComboBox Grid.Row="20" Grid.Column="0"  Grid.ColumnSpan="10" materialDesign:HintAssist.Hint="OS">
                <ComboBoxItem>Andoid</ComboBoxItem>
                <ComboBoxItem>iOS</ComboBoxItem>
                <ComboBoxItem>Linux</ComboBoxItem>
                <ComboBoxItem>Windows</ComboBoxItem>
            </ComboBox>`

Launch it full screen and you will see that the combo box is not drawn properly.
In fact, the SelectedItem is not inside the "comboxpopup header". So you think that this element is not belonging to the popup.

![capture](https://cloud.githubusercontent.com/assets/7021898/20307112/3455c994-ab3e-11e6-92f2-178889b6c6ac.PNG)

I hope you can fix this template.

Thanks,

l1pton17:
When combobox doesn't have enough space to show suggestion list right under the selecteditem, it shows list out like standard combobox do.
Maybe the solution of this problem is to somehow mark the selecteditem that this list belongs to it or show selecteditem in the suggestion list when it stands out.
Also, you can use ComboBoxAssist.ShowSelectedItem in order to show selecteditem in the drop down list.

ButchersBoy:
@l1pton17 I think if we have to display in "classic Windows" mode we should probably do the ShowSelectedItem==true behaviour by default.

l1pton17:
@ButchersBoy yeah, i thought about it when i was typing previous message :smile:

elvince:
Excellent Thanks!

Issue #551, opened 2016-11-11
ComboBox popup doesn't show full text of selected item.


jlesp:
When a ComboBox doesn't have enough space to display the full text of the selected item, it is truncated as such:
![closed](https://cloud.githubusercontent.com/assets/20480367/20231624/86dd07f2-a820-11e6-8179-0d050754173f.png)

However, when the ComboBox is opened and the popup is displayed, the selected item text should be displayed in full instead of still being truncated:
![open](https://cloud.githubusercontent.com/assets/20480367/20231649/b66d344c-a820-11e6-9a37-bc7869598a91.png)

elvince:
In fact, the way the template have been done, the selected item that you are seeing is the "current render" of the combobox in your view. It's not part of the list that is rendered in the popup. In fact, in the popup header, there is "hole" so you can see what's under and in this case, It's you combobox.

This design is causing side effect like yours or like my issues: #552

ButchersBoy:
Hi.  Achieving the true look and feel of the Material Design combo, compared to what we see on Android phones, but utilising the WPF ComboBox control, is a pretty tricky task.  There is a mode where it drops below in a more classic sense.  I've wanted to expose this mode as an option for a while, for cases such as yours where display is not working correctly.  So I will try and do this.

Filling in the clipped area is not really an option as delegating keyboard presses becomes a pain in the a$$.

jlesp:
@ButchersBoy I had also noticed that the popup just had a 'window' to the ComboBox behind it. I actually thought this was a clever way to implement the MaterialDesign combo in WPF and avoid any flicker issues, have the fill around, etc.

Thanks!

ButchersBoy:
Well, credit to @l1pton17 for his work on this control.

ButchersBoy:
@l1pton17 in this situation should it not jump into "classic" mode?

l1pton17:
@ButchersBoy it can lead to incomprehension from customers like "why in one sitation i see beatiful wrapper and in another don't?" :smile:

ButchersBoy:
Yes this is a good point.

I think from pure UI design perspective you have to ask yourself if you are going to layout out your combos so you can see the whole field.  Classically, in desktop apps (especially LoB apps) we tend to squash things together, probably a bit like in the above screenshot.  

Personally, I like to give things a lot more white-space.

However, I do think we need to introduce the option to force "classic" mode...then client apps can set a standard look and feel for their whole application.

[...]

PR #558, opened 2016-11-19
ComboBox tweaks

l1pton17:
Solving issues #552 & #551 

* Show selected item when a combobox uses classic mode
* Rename DefaultTemplate to ClassicTemplate
* Make ComboBoxPopup internal, because it isn't a part of public api
* Add ComboBoxAssist.OnlyClassicMode attached property that allow use only classic dropdown view


l1pton17:
@ButchersBoy sorry, english is not my mother tongue, i will do rename this property :smile:

PR #576, opened 2016-12-17
put the content of the demo into ScrollViewer

spiegelp:
put the content of the demo into ScrollViewer because the content was cut in smaller windows


ButchersBoy:
@spiegelp awesome, thankyou!

PR #596, opened 2017-01-19
Default overrides for ListView and ListViewItem

s73obrien:
Set to ListBox and ListBoxItem defaults, respectively


ButchersBoy:
Hi, thanks for this; we will have proper style for 2.3.0

PR #612, opened 2017-02-07
Only allow focus on the ToggleButton in the PopupBox templates

stigrune:
**Before fix**; Focus is set to several of the objects in the template when using tab key.
popupbox focus bug

**After fix**; Focus is only set on the PopupBox togglebutton when tabbing.
popupbox focus bug fixed


ButchersBoy:
Thank you!

PR #615, opened 2017-02-09
Properties for controlling the transition origin

mgnslndh:


mgnslndh:
It is not necessary at all. Just for convenience.

ButchersBoy:
Yes please.  We can always add it in later if it proves to be of common use; I just don't want to add stuff that becomes a permanent part of the public API unnecessarily.   ...maybe it will prove it's worth in the future and we can add it back.

mgnslndh:
You're right, i'll fix it tonight

ButchersBoy:
I've got 5 mins, I'll merge myself and move the Points.

PR #616, opened 2017-02-09
Fix broken card in demo

mgnslndh:
New flat tool style for the ListBox broke a card in the demo. This is a tiny fix for that.


ButchersBoy:
Damn, it's broken more than just that card :/

ButchersBoy:
Gonna revert; and fix underlying issue.

PR #623, opened 2017-02-16
Card no longer Focusable="True" by default

mgnslndh:


ButchersBoy:
Thanks!

mgnslndh:
Im guessing there are more styles where Focusable should not be True so I might go through a couple of controls and find out later.

ButchersBoy:
Yeah, I always forget to tidy them up :)

PR #624, opened 2017-02-16
Fix ComboBoxPopup leak

spdjudd:
Make sure RemoveValueChanged is called on ComboBoxPopup.Child property, from issue https://github.com/ButchersBoy/MaterialDesignInXamlToolkit/issues/622


ButchersBoy:
Awesome, thank you!

PR #631, opened 2017-02-21
Fix content cover remaining visible.

Keboo:
Reproduction steps:
1. Launch the WPF test application
2. Select the EXPLORE button on the home page
3. Make a section from items on the left

Observed: The content cover remains visible after the left drawer is dismissed.

The reason is because the command was not passing any Dock value as its command parameter. This was causing all of the Is*DrawerOpenProperty to be set to true. thus keeping the content cover visible after only the left drawer is dismissed. 

The change here fixes the issue by specifying that only the left drawer should be opened.


ButchersBoy:
Ooops!  Thank-you!

Issue #370, opened 2016-05-06
Add validation error display to RadioButton controls


arc95:
`RadioButton` controls do not support displaying validation error messages underneath the control as is supported by `TextBox`, `ComboBox`, etc. Currently, when a `RadioButton` fails validation, it receives a red border. It should display the validation error underneath as do the other controls. Thanks.

PR #634, opened 2017-02-22
Validation error display for RadioButtons and CheckBoxes

Keboo:
Fixes issue #370.
Added validation template to check boxes and radio button.

Looking through the [material design docs](https://material.io/guidelines/patterns/errors.html) I could not find any guidelines on exactly how to style things like the checkbox and radio button (I suspect because validation on these is a very rare case). IMO I like it when the graphic and the content are both turned red so that is how they are implemented. I am happy to remove this and just have the red text below the control if that is preferred as well. 

I have added validation to a radio button and check box on the buttons demo pages so you can see the result.


ButchersBoy:
Yeah, personally I find it wierd UX, however, it doesn't do any harm to support the validation template.

PR #645, opened 2017-03-03
Add SlideWipe and FadeWipe

Tyrrrz:
Based on SlideOutWipe


ButchersBoy:
Awesome, thank you!

One thought...could we get a single slide wipe, with enum; up/down/left/right?

Also, I just want to push 2.3.0 out first, so this wont make the cut in time for that.

Tyrrrz:
Unified them into SlideWipe. Didn't delete SlideOut because it functions a bit differently.

Tyrrrz:
If you like it, I can also add a few more like FadeWipe (with enum for In and Out).

Do you think we could also maybe unify `TransitionEffects` with `TransitionWipe`? I know they are used for different purposes but it somehow looks like they can be generalized and it's also slightly confusing having to choose between the two.

ButchersBoy:
Yes, a fade wipe would be awesome...I always intended to add more wipes but haven't got around to it so these are helpful.

The effects will stay; I use them personally, plus it is part of the API now so I don't want to remove without a strong case for doing so.

Tyrrrz:
Added the FadeWipe too

ButchersBoy:
Hey, I tested this by changing TransitionsDemoHome.xaml to:

```
            <!-- you can control (and create your own) wipes -->
            <materialDesign:TransitionerSlide>
                <materialDesign:TransitionerSlide.BackwardWipe>
                    <materialDesign:FadeWipe />
                </materialDesign:TransitionerSlide.BackwardWipe>
                <local:Slide3_Intro />
            </materialDesign:TransitionerSlide>
```

Unfortunately, when I jumped back and forth, the slide was no longer visible.  I think a few extra lines on the Completed handler (such as in CircleWipe) are required for a full reset).

Tyrrrz:
Fixed it

PR #650, opened 2017-03-08
Adding support for handling SpellCheck.IsEnabled on TextBoxes.

### LINKED ISSUE #497: Textbox Context Menu and Spell Check
I have come accross a problem with the context menu changes to TextBoxes. In the code (MaterialDesignTheme.TextBox.xaml) on line 27 is a comment "cludge the default context menu". It seems this "cludge" is preventing the normal spell check menu from displaying.

Sorry for not adding code examples; but I don't have access at the moment and I wanted to submit the bug before I forgot!

Keboo:
Fixes #497

Though I understand that the intent it for this library to be a theme library, not a control library, it seems like a loss in functionality for someone moving from standard WPF. 
The intent of this was to simply handle the cases where someone merely sets `SpellCheck.IsEnabled="True"` and then includes this theme library.

I have attempted to handle all possible cases, with a few things in particular worth pointing out.

- The feature was implemented with an attached property in `TextFieldAssist.IncludeSpellingSuggestions`, and bound its default value to `SpellCheck.IsEnabled`. This should allow for people wanting to roll their own UI for the spell check (and thus have `SpellCheck.IsEnabled="True"`), and simply want to opt-out of the behavior for including these in the context menu.
- I created derived classes from MenuItem for the various spell check options, and styled the text to roughly default WPF. This should allow anyone wishing to override the default look to do so easily.
- There are two cases where I hard coded text, "Ignore All" and "(no spelling suggestions)". I put this text in the default style for their corresponding menu items. I can see how this might be a bit of a localization issue, and am open to suggestions.

Thoughts?


ButchersBoy:
Yes, I definitely want the fix...I will review the code after 2.3.0 is out the door though. :+1:

ButchersBoy:
image

Adding image for reference.  Looks awesome!

ButchersBoy:
This looks good.  I'm happy with the implementation apart from the derived MenuItem classes.  I'm not a big fan of inheritance and wonder if we can get away with this by using normal MenuItems...I guess the down side is we have to look up the style...which might have a certain fragility.  But anything increasing the surface API really has to prove its worth to me...

Keboo:
I was just trying to make it easy to style those items. However, I suppose I could put those styles as attached properties on the ContextMenu itself that way they could be swapped out if someone wanted to change the styles.

ButchersBoy:
Could we try giving them style keys, a bit like toolbars?  This would be more in line with standard WPF and lose the inheritance.

<Style x:Key="{x:Static ToolBar.ButtonStyleKey}" TargetType="Button">

Keboo:
Sure, I will update the PR in a few.

ButchersBoy:
Thank-you, nice work!

Issue #646, opened 2017-03-04
Android-like scrollbar style


Tyrrrz:
It would be nice to have a style for scrollbar that mimics how it usually looks on mobile devices - hidden until you start scrolling, doesn't take up space, is overlayed on top of content.

A lot of desktop and web apps also use some variation of that style. For example, here is how it looks on Twitch.tv (the scrollbar is the greyish round rectangle on the right).

![2017-03-04 19_37_27-lirik - twitch](https://cloud.githubusercontent.com/assets/1935960/23580849/0b24732a-0112-11e7-831a-40adbb5bb551.png)

maciz84:
+1

Tyrrrz:
Also this looks like a good base to start off:

http://stackoverflow.com/questions/10969292/apple-style-scrollbar-wpf

maciz84:
I used the same stack overflow xaml code and it works well. I still would like to have an extension on the scroll bar so it increases in width when hovered and not visible at all until control is focused

ButchersBoy:
We need to remember we are on the desktop.  Scrollbars on mobiles are indicators.  On the desktop they are tactile.  The user actually has to get their pointer on to them to use them.  4 pixels to aim at makes that harder.

I can't agree with this: "not visible at all until control is focused".  It is possible to make a mouse gesture on a control (or even Window) that doesn't have focus.  This will just incur the user an extra click to take focus, before they can scroll.  Maybe a mouse over to make visible has merit, but also you've got a lifetime of desktop convention to work with: users see scrollbars on a monitor; they know they can scroll.

We are on a different form factor than mobile, and it's probably because of this there is no real agreed convention, and nothing in the Google Material Design spec.

I've been looking at the inbox.google.com scrollbars...they are much more "traditional".

Anywayz...post your thoughts below...

bebenins:
Have to agree with James.  Some elements simply cannot carry over from mobile sized touchscreens to full desktop with keyboard/mouse.

bebenins:
Perhaps there could be some design "improvements"/changes.  For example, Hangouts desktop app and web Gmail have something closer to mobile scroll bar design but still usable on the desktop.

Keboo:
[...]

PR #651, opened 2017-03-08
removed repeat buttons from scrollbar for a more minimalistic style

Insire:
Read through #646 and figured i give you something to play with.

Is this something you'd like to pursue, or does this not quite hit the spot? @ButchersBoy


Tyrrrz:
Personally, I like where this is going, but would prefer if it was slightly thinner.

Insire:
The style uses the default windows size for scrollbars. I feel the same though. Making the scrollbar thinner would improve the look and feel.

metherul:
I gotta agree, this looks really good.

uPloraX:
Maybe I would add a bit of border radius to both ends too, otherwise it looks more like Metro and not in place when Material Design has these slightly details at whole.

maciz84:
@Insire Do you have the style template for this scrollbar for sharing by any chance?

maciz84:
@Insire oops my bad found it....I am fairly new to Github :)

Keboo:
Putting this in the 2.6.0 release as I would like to see this issue revisited and brought forward.

jespersh:
Would it be possible to add a property that shows/hides the RepeatButtons? Outright removing them seems a little too much, at least to me :-)

Keboo:
Thank you for this work. I referenced it while getting the current styles in and done. See the work on PR #1270. Let me know if there is anything I have missed.

PR #664, opened 2017-03-25
rename class to fix 404 error

thimmy687:
rename class to fix 404 error
update class with current c# features
change namespace


ButchersBoy:
Hi...which link was actually failing?

thimmy687:
The listsandgrids link was failing

PR #665, opened 2017-03-28
Fix pops remaining visible.

### LINKED ISSUE #658: PopupEx Stay active when close
Hi,

Following my request #655 I saw that there is an issue with the PopupEx.

When it's closed, the popup is "still" there. 
To hightlight the issue,  you can set the AllowsTransparency to "False" instead of true. (in generic.xaml, line 602)
you will see a black background instead of the transparency one.
When you press "cancel" button (on the add fruit sample by example), you will see that the content is fading properly but that the Popuphost is still there.

image

Thanks

Keboo:
Fixes #658 

Fixes the issue with the popup remaining visible (fixed IsOpen in transition from Open to Closed)

Fixes the issue with the popups showing initially then hiding (removed initial state where IsOpen was being set to true only to be set to false).


elvince:
Working as expected now. The 2 issues are solved. Good job!

PR #668, opened 2017-04-01
Close dialog when window is hidden.

### LINKED ISSUE #644: DialogHost freezing
[**Source**](https://github.com/Carlj28/MaterialDesignDialogHostFreezing)

Steps to reproduce:
1. Click "Open sub window"
2. Click "Open DH"
3. Close "Sub window" (close is overriden, it actualy hides)
```
        private void Window_Closing(object sender, System.ComponentModel.CancelEventArgs e)
        {
            this.Hide();
            e.Cancel = true;
        }
```
4. DialogHost is still open after closing (hiding) parent window

5. Click "Open sub window"
6. Click "Open DH" and DialogHost is now frozen

This occurs when parent window of DialogHost is Hidden instead of closing.

Keboo:
Fixes #644 

When the parent window is hidden, the DialogHost closes its popup.

### DIFF
--- a/MaterialDesignThemes.Wpf/DialogHost.cs
+++ b/MaterialDesignThemes.Wpf/DialogHost.cs
@@ -696,9 +696,16 @@ namespace MaterialDesignThemes.Wpf
 
         private void WindowIsVisibleChanged(object sender, DependencyPropertyChangedEventArgs e)
         {
-            if (!(bool) e.NewValue)
+            if (IsOpen)
             {
-                Close(null);
+                if (!(bool) e.NewValue)
+                {
+                    VisualStateManager.GoToState(this, ClosedStateName, !TransitionAssist.GetDisableTransitions(this));
+                }
+                else if ((bool) e.NewValue)
+                {
+                    VisualStateManager.GoToState(this, OpenStateName, !TransitionAssist.GetDisableTransitions(this));
+                }
             }
         }
 
--- a/MaterialDesignThemes.Wpf/DialogHost.cs
+++ b/MaterialDesignThemes.Wpf/DialogHost.cs
@@ -217,6 +217,7 @@ namespace MaterialDesignThemes.Wpf
             Loaded += OnLoaded;
             Unloaded += OnUnloaded;
 
+
             CommandBindings.Add(new CommandBinding(CloseDialogCommand, CloseDialogHandler, CloseDialogCanExecute));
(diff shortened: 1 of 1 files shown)


ButchersBoy:
I'm wondering if we should hide the popup (temporarily) and not close it.  Because 1) the dialog could stuill be valid if a window is re-shown, and 2) if we start calling ``` Close{null) ``` then there maybe unwanted effects in user code where people may coded for specific return parameters.

We actually have a ``` CloseOnClickAwayParameter ``` for similar circimstances...but I'm wondering if adding a switch is just going to complicate the API...

Thoughts?

Keboo:
This actually touches on something I have been thinking about. Have you ever considered changing out the DialogHost to simply be a content control, rather than utilizing the popup (similar to the way MahApps does their dialogs)? It just seems like this approach would be much simpler to manage as the dialog would then behave the way it looks (as an element within the visual tree; rather than a light weight window outside of it).

As for this particular issue. What about a flagged enum (perhaps obsolete the existing CloseOnClickAway)?

```C#
[Flags]
public enum DialogCloseEvents
{
    None = 0,
    ClickAway = 1,
    ParentWindowHidden = 2
}

```

ButchersBoy:
Yes, I did consider this in the beginning, and decided against it.  The problem with that is it's size is constrained to the parent Window.

Keboo:
Fair enough. In my own applications, I actually want the size constrained to the parent window so I am just using relative bindings on MaxWidth/MaxHeight to enforce it.

Keboo:
@ButchersBoy see the updated changes. I did both the flagged enum and well as only hiding/showing the dialog when the parent window is hidden/shown. 
I am happy to change this up more, if this is not in line with what you were thinking.

ButchersBoy:
I would prefer we take a step back a little and simplify this.

The underlying UX issue is that we just need to hide the dialog whilst a window is minimized (right?).

So the extra "trigger" is unnecessary.  Once a window is restored, the dialog should still be there.  The user hasn't dismissed it and and the code hasn't called a .Session.Close().

Plus I'd rather avoid obsoleting stuff unless necessary.

Keboo:
@ButchersBoy more like this?

Keboo:
Closing this for now. After reviewing my solution I don't believe it is the best possible option. A better solution is to use the embedded dialog host style instead.

PR #669, opened 2017-04-02
Clipped shadows

Keboo:
Adding the ability to control which edges the shadows are rendered on. Typically this is not an issue as you can simply lay another control over the top, but there are cases where it is simpler to simply clip the shadow to only the desired edges.

These changes primary apply to ColorZone and Card, however these are also used in the control templates of other controls so I tried to add in support for setting the attached property where appropriate.

The ShadowEdges enum is flagged so you can easily specify a single or group of edges for the shadow to render on.

Example usage:
```XAML
<materialDesign:ColorZone Mode="Accent" Padding="16" materialDesign:ShadowAssist.ShadowDepth="Depth3" materialDesign:ShadowAssist.ShadowEdges="Bottom,Right">
...
</materialDesign:ColorZone>
```

### DIFF
--- a/MaterialDesignThemes.Wpf/Converters/ShadowConverter.cs
+++ b/MaterialDesignThemes.Wpf/Converters/ShadowConverter.cs
@@ -1,10 +1,5 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.Globalization;
-using System.Linq;
-using System.Text;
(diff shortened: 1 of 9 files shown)


ButchersBoy:
Nice work, thank you

MrMancunian:
When will this be in the NuGet package?

PR #683, opened 2017-04-14
Specified colors for themes

Sergey-Terekhin:
Added theme colors keys - for example, I had to theme AvalonDock control and the easiest way was to override existing brushes with Material colors, not to use existing brushes


mgnslndh:
@Sergey-Terekhin Did you make a material theme for AvalonDock? If so, are you going to release it?

Sergey-Terekhin:
@mgnslndh I can release it, but I'm only replaced colors with Material palette and icons are taken from that package not more -no animations, transitions etc

ButchersBoy:
This PR seems to break the appearance of cards, shadows are gone:

image

jigneshBadrakhiya:
any way to create a new color scheme from code?

Keboo:
Moving this into 2.6.0 as there will be quite a few changes going in around colors and theming.

Keboo:
Superseded by #1296

PR #685, opened 2017-04-14
ComboBox to handle multiple monitor setups

### LINKED ISSUE #635: ComboBox popup defaulting to ClassicMode when used on monitor to the left of the primary display
I've got a 3 monitor setup, with the middle monitor as my primary display.  When I open a popup from the demo on the middle or right monitor, the popup opens on top of the combobox, using ComboBoxPopupPlacement.Down as expected.  When I open it on the monitor to the left of the primary display, it opens using ComboBoxPopupPlacement.Classic.

When looking at the GetPositioningData of the ComboBoxPopup class, I see that the locationFromScreen variable is getting a negative X value when opened on the monitor to the left of the primary display.  This seems to be what's causing the ComboBoxCustomPopupPlacementCallback method to default to the classic mode.

Keboo:
Fixes #635 

This allows for the combo box to function correctly when the app is on a different monitor.

Fixed issue with the combo box popup not being displayed correctly multiple monitor setup.
The screen class is a port from WinForms and updated to C# 7 and be WPF friendly.


Keboo:
@ButchersBoy this appears to be failing to due to my C# 7 usage. Do you want to update appveyor or would you prefer I change the code?

Insire:
You could include [Microsoft.Net.Compilers](https://www.nuget.org/packages/Microsoft.Net.Compilers/), which will make use of the integrated compiler of that package during build, instead of the local one of the machine its being built on.

Keboo:
@Insire thank you I am aware of that. However, since the project is migrating to VS2017 it seems that the build server (AppVeyor) should also build using the newer compiler without needing to take a nuget dependency.

Keboo:
Latest AppVeyor changes for C# 7 fixed the compile for this. :)

ButchersBoy:
Hey, this is almost there...I have three monitors...on my right monitor it still shows in classic mode.  *However*, my right monitor is vertical which may add to complication.

If you can't look at this/reproduce it, let me know, I will do some debugging on my setup.

* Can we make Screen class internal please, so it doesn't become part of our API.

Keboo:
I was able to reproduce it. 

My original monitor setup that I tested with looked like this (worked):
image
But I was able to reproduce your issue when I moved my second monitor's position below the primary.
image

PR #688, opened 2017-04-17
Fixing attached property signatures

Keboo:
Fixing the signature of the ComboBoxAssist.Get* attached property methods.


ButchersBoy:
Yikes...I  am going to be flexible and not follow semver for this...

PR #660, opened 2017-03-16
Fix possible Badged name scope bug

punker76:
This PR fixes a nasty possible bug if a Badged control is inside a TabItem (or other non loaded control) found at this issue at MahApps https://github.com/MahApps/MahApps.Metro/issues/2885

_"Exception = 'BadgeScaleTransform' name cannot be found in the name scope of 'System.Windows.Controls.ControlTemplate'."_

This PR adds a new dependency property too where you can change the Storyboard (if you want).

PR #701, opened 2017-05-01
Fix exception when badge has not been loaded

### LINKED ISSUE #700: Badged control Badge property throws Exception with PropertyChanged
When I bind a property from a ViewModel to the `Badge` property of the `Badged` Control, it will throw an expection on the `PropertyChanged` Event.

```
<materialDesign:Badged Badge="{Binding EnqueuedDownloads}" />
```
Example binding property with Prism `BindableBase`
```
public int EnqueuedDownloads
{
    get { return _enqueuedDownloads; }
    set { SetProperty(ref _enqueuedDownloads, value); } <-- Crashes here
}
```
```
EnqueuedDownloads++; or EnqueuedDownloads += 1;
```
```
System.InvalidOperationException:
The name "BadgeScaleTransform" cannot be found in the namespace of "System.Windows.Controls.ControlTemplate"
```

Keboo:
Fixes #700 

Preventing the event from firing until the control has actually been loaded.

### DIFF
--- a/paket-files/ControlzEx/ControlzEx/src/ControlzEx/BadgedEx.cs
+++ b/paket-files/ControlzEx/ControlzEx/src/ControlzEx/BadgedEx.cs
@@ -50,7 +50,7 @@ namespace ControlzEx
         }
 
         public static readonly DependencyProperty BadgePlacementModeProperty = DependencyProperty.Register(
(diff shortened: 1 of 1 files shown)


punker76:
@Keboo This is the wrong place to fix this. It should go to the ControlzEx lib. And what's with #660 /cc @ButchersBoy

Keboo:
Thanks @punker76

Issue #703, opened 2017-05-02
PopupBox is not open if layout is changed


Sergey-Terekhin:
Hello!
I've encountered bug with PopupBox control:
If it placed on control which layout is changed in runtime (for example, grid's row become visible or hidden), it show content only while layout is unchanged. If it's changed, PopupBox seems to be inactive
Sample repository below:
https://github.com/Sergey-Terekhin/MaterialDesignPopupBoxTest

Keboo:
@ButchersBoy I believe this issue can be closed now.

PR #704, opened 2017-05-02
Update popup box popup when the anchor control moves.

Keboo:
Proposed fix for #703.

The popup is not properly repositioning itself when its anchor control is moved. This updates the popup's position whenever the position between the two has changed.


ButchersBoy:
:+1: :+1:

PR #711, opened 2017-05-19
Progress bars were pictured twice. Removed one.

Elthan:


ButchersBoy:
I have been meaning to fix this for so long!  :+1:

Issue #670, opened 2017-04-03
RatingBar should have an 'IsReadonly' Property or at least adapt to OneWay binding in the 'Value' property


bigworld12:

ButchersBoy:
Read only sounds reasonable

bigworld12:
Hasn't this issue been solved yet ?

bigworld12:
a Workaround for this, is to set IsEnabled to false and opacity to 1

Keboo:
This was resolved by #714

PR #714, opened 2017-05-21
Added IsReadOnly Property to RatingBar

Insire:
Added IsReadOnly Property to RatingBar, which disables updating the Value-Property of the RatingBar, when any of the child RatingBarButtons is clicked. The OnClick spinning animation is disabled by this aswell.

#670


Insire:
I'm sorry about the extra changes, i have an extension that removes them for me when saving documents. Is this still acceptable?

ButchersBoy:
Hey...can't we just use existing IsEnabled property (and stop it greying out)?

Insire:
Sure, thats an option. The tradeoff is predictability of the default behavior vs more code.

I prefer the more code approach, so that its easier to use for the dev using the ratingbar, but if you like the to use the IsEnabled property instead, i can do that.

Keboo:
@Insire are you able to address the merge conflicts?

Insire:
Pulled and rebased onto master.

Also added a sample for a readonly ratingbar to the demo app.

PR #717, opened 2017-05-24
Fixing combo box issue on right screen.

Keboo:
Requested changes after PR #685 was merged.

Fixing issue where combo box would revert back to classic mode on right monitor to the right of the primary screen.

Made Screen class internal. Initially I just took all of the code for this class from Windows Forms. However you don't want this as part of the API, there are many methods in there that are not being used and could be stripped out if desired.


ButchersBoy:
:+1:

PR #721, opened 2017-05-30
PasswordBox hint alignment

### LINKED ISSUE #718: PasswordTextbox has unsincronized caret position with hint position
When setting padding property in a PasswordBox, the hint and caret positions, become different.
Looking in the source code of the component (PasswordBox.xaml), in the hint section, is possible to see the absence of binding the padding property.

Adding the following code to the SmartHint section 
`Padding="{TemplateBinding Padding}"`
should solve the problem.

Sorry for bad english.

Keboo:
Fixes #718 

Adding template binding to Password Text Box so the hint properly aligns.


ButchersBoy:
Thanks!

PR #732, opened 2017-06-21
Additional styles for TextBox

spiegelp:
Additional TextBox styles for text field boxes and text areas shown in the guidelines.

see "Field types" -> "Text area" and "Text field boxes" on [https://material.io/guidelines/components/text-fields.html](https://material.io/guidelines/components/text-fields.html)


ButchersBoy:
Nice work! :+1:

PR #741, opened 2017-07-05
Adding template bindings for border on the flat button style.

Keboo:
Previously these were unused. On my application it would be nice to apply a border to a flat button, however these values were previously being ignored.

The border brush obviously still gets toggled when the mouse hovers over it. The only small side effect to this change is that flat buttons are now 1 pixel bigger on all sides. This is because the style declares a uniform BorderThickness of 1, but was not using it anywhere. If this is not desirable I can change the style to set a border thickness of 0.
Other than that the look and behavior remain unchanged.


ButchersBoy:
Yeah...can we default the border thickness to zero...then we wont freak anyone out when all their apps are suddenly 1 pixel out of alignment...

Keboo:
Done

ButchersBoy:
:+1:

PR #747, opened 2017-07-17
Fix a bug in Demo and added new control : PopupBoxItem

wongjiahau:
-Clicking the 'Copy to clipboard' button will cause the application to crash.
-Now the bug is fixed
-Also, now snackbar will be shown to notify user that the code is copied to clipboard

### DIFF
--- a/MainDemo.Wpf/MainWindow.xaml
+++ b/MainDemo.Wpf/MainWindow.xaml
@@ -60,8 +60,10 @@
                     <DockPanel>
                         <ToggleButton Style="{StaticResource MaterialDesignHamburgerToggleButton}" IsChecked="False"
                                         x:Name="MenuToggleButton"/>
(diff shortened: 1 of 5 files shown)


wongjiahau:
untitled

ButchersBoy:
Hi,

If we are going to add ```PopupBoxItem``` then it's style needs to be set in XAML, not in it's constructor.

Can we have these two things as two individual pull requests, so I can review each change in isolation please.

wongjiahau:
Ok sure, I'll set the style into XAML, and will make these 2 commits and individual pull requests. 
Thank you for the reply. 
I'm very glad that you replied because this is my first attempt to try to contribute to an open source project :)

Keboo:
Looks like this PR can be close now that #749 and #750 contain the relevant pieces.

PR #749, opened 2017-07-17
Fix a minor bug in MainDemo.wpf

wongjiahau:
Bug : Clicking the "Copy to clipboard" button in IconPackView will cause an exception to be thrown.
Now the bug is fixed. 
Solution is based on StackOverflow answers : 
[https://stackoverflow.com/questions/12769264/openclipboard-failed-when-copy-pasting-data-from-wpf-datagrid](url)

PR #750, opened 2017-07-17
Added new Control : PopupBoxItem 

wongjiahau:
This control allow user to place icon easily in PopupBox menus. 
//Sorry that I've accidentally included the previous "Fix minor bug" commit 
untitled

### DIFF
--- a/MainDemo.Wpf/IconPackViewModel.cs
+++ b/MainDemo.Wpf/IconPackViewModel.cs
@@ -17,9 +17,12 @@ namespace MaterialDesignDemo
     public class IconPackViewModel : INotifyPropertyChanged
     {
         private readonly Lazy<IEnumerable<PackIconKind>> _packIconKinds;
+        private readonly ISnackbarMessageQueue _snackbarMessageQueue;
 
-        public IconPackViewModel()
+        public IconPackViewModel(ISnackbarMessageQueue snackbarMessageQueue)
         {
+            _snackbarMessageQueue = snackbarMessageQueue ?? throw new ArgumentNullException(nameof(snackbarMessageQueue));
+
             OpenDotComCommand = new AnotherCommandImplementation(OpenDotCom);
             SearchCommand = new AnotherCommandImplementation(Search);
             CopyToClipboardCommand = new AnotherCommandImplementation(CopyToClipboard);
@@ -27,7 +30,6 @@ namespace MaterialDesignDemo
                 Enum.GetValues(typeof (PackIconKind)).OfType<PackIconKind>()
                     .OrderBy(k => k.ToString(), StringComparer.InvariantCultureIgnoreCase).ToList()
                 );
-
         }
 
         public ICommand OpenDotComCommand { get; }
@@ -66,7 +68,7 @@ namespace MaterialDesignDemo
             var kind = (PackIconKind?)obj;
             string toBeCopied = $"<materialDesign:PackIcon Kind=\"{kind}\" />";
             Clipboard.SetDataObject(toBeCopied);
-            MainWindow.Snackbar.MessageQueue.Enqueue(toBeCopied + " is copied to clipboard!");
+            _snackbarMessageQueue.Enqueue(toBeCopied + " copied to clipboard");
         }
 
         public event PropertyChangedEventHandler PropertyChanged;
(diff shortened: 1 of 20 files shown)


wongjiahau:
How to use PopupBoxItem in XAML?
codebehind

MainDemo.Wpf/IconPackViewModel.cs
```diff
@@ -63,7 +63,7 @@ private void Search(object obj)
         private void CopyToClipboard(object obj)
         {
             var kind = (PackIconKind?)obj;
-            Clipboard.SetText($"<materialDesign:PackIcon Kind=\"{kind}\" />");
+            Clipboard.SetDataObject($"<materialDesign:PackIcon Kind=\"{kind}\" />");
```
Keboo: We probably do want to take this change out of this PR since it is already included in #749

MaterialDesignThemes.Wpf/PopupBoxItem.cs
```diff
@@ -0,0 +1,77 @@
...
+            DependencyProperty.Register(nameof(Text) , typeof(string) , typeof(PopupBoxItem) , new FrameworkPropertyMetadata("Example" , OnTextPropertyChanged));
+
+        private static void OnTextPropertyChanged(DependencyObject d , DependencyPropertyChangedEventArgs e) {
+            var control = d as PopupBoxItem;
+            if (control._textBlock == null) return;
+            control._textBlock.Text = (string)e.NewValue;
+        }
+        #endregion
+    }
+}
```
Keboo: Just a minor styling thing. The rest of the code base typically puts of the opening brace "{" on a new line rather than at the end of the statement.

MainDemo.Wpf/IconPackViewModel.cs
wongjiahau: Yes, please take that commit out, that was accidentally included by me

Keboo: You will need to take it out since the PR is from your branch. The "easiest" thing to do is to simply rebase your New-control branch from master rather than from the commit with this change. 

TL;DR 
From your New-control branch Just run this command and do a force push:
git rebase --onto f1ed10c3eedd6b23f490a9f07d102c4f8cd4670b abc3228f6788c318329607
5bb033f04e6ab45c3b

For these screenshots I am using [GitKraken](https://www.gitkraken.com/) (which is free for projects like this) to show the commit tree but I am running the git commands from the command line.
Here is a view of the current commit tree:
image
Because we want to basically pick up the commit and change its parent we use the [Git rebase](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) command.
The command above does just that. The `--onto <hash>` is the hash of the commit that we want to make the new parent (ie. the parent we want to be based on). The second hash is the hash of the current parent.
After running the command on the command line the tree will look like this:
image
GitKraken shows local commits with a little computer symbol, and remote commits with a little merge symbol. The remote commit will not change until you push. Because you are changing the parent you will need to [force the push](https://stackoverflow.com/questions/5509543/how-do-i-properly-force-a-git-push) to go through.
`git push origin New-control --force`
After this you should see the local and remote commits in the same spot on the commit tree.

Hope that helps.

wongjiahau: Thanks for the help! I will try to do it.

MaterialDesignThemes.Wpf/PopupBoxItem.cs
```diff
@@ -0,0 +1,77 @@
...
+            var popupBoxItem = (PopupBoxItem)d;
+            var newValue = (PackIconKind)e.NewValue;
+            if (popupBoxItem._icon == null) return;
+            popupBoxItem._icon.Kind = newValue;
+        }
+
+        #endregion
+
+        #region TextProperty
+        public string Text {
```
ButchersBoy: We should follow the standard WPF paradigms.

If you look at MenuItem, you will see that ```.Icon``` and ```.Content``` are of type ```object```.

Even if 90% of use cases are icons and text, we shouldn't limit them, and by using ```object``` standard WPF concepts such as templates come into play.

So we don't actually need a ```Text``` as this is effectively the ```Content``` as this ultimately derives from ```ContentControl```. 

You will see the framework's adherence to core WPF patterns in this manner in the ```Chip``` control.

wongjiahau: Ok sure! Thanks for the advice.

wongjiahau:
I had already updated it by using content control

wongjiahau:
I closed this pull request because I accidentally included irrelevant commits by other contributors.

PR #752, opened 2017-07-23
Tweaked Typography page in MainDemo

wongjiahau:
- Now user can copy the style for each typography easily
typography


Keboo:
This functionality has been handled with the addition of ShowMeTheXAML.

image

PR #754, opened 2017-07-23
Added guideline source for each type of control

wongjiahau:
- so that user can easily refer to the original guideline posted by Google


wongjiahau:
@Keboo I would try but I won't be able to test the app because I switch to Linux :rofl:

Keboo:
I like this option but since Google has changed the URLs they are no longer accurate. Going to close for now. If anyone wants to revisit this and adjust the links I am happy to consider an updated PR.

PR #755, opened 2017-07-23
New control : popup box item

wongjiahau:
- now the internal part is using content control

### DIFF
--- a/MainDemo.Wpf/MainWindow.xaml
+++ b/MainDemo.Wpf/MainWindow.xaml
@@ -62,8 +62,16 @@
                                         x:Name="MenuToggleButton"/>
                         <materialDesign:PopupBox DockPanel.Dock="Right" PlacementMode="BottomAndAlignRightEdges" StaysOpen="False">
                             <StackPanel>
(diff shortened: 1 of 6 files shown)


ButchersBoy:
Hi...i'm just holding this back until after the next release (2.3.1).

Reasoning:

For a while I've wanted to smarten up the drawer content in the demo, and I think we can use this as part of that work.  So it might warrant renaming to something a bit more generic, and we could introduce a few extra styles.

So I will come back to this.

wongjiahau:
@ButchersBoy  Noted. Thanks for the reply. Please do call me up if you need me to add extra things to this control.

Keboo:
I am going to close this for now. Though I understand the desire to have a control that includes an icon, the `Button` is a control control and can easily have an icon added to its content. This would then have the pack icon properly change for dark theme. This method also leaves you in control of things like the margin around the icon as well. 

For example:
```xaml
<Button Click="MenuPopupButton_OnClick">
        <materialDesign:PackIcon Kind="Settings"/>
        <TextBlock Text="Hello World" />
    </StackPanel>
</Button>
```

PR #762, opened 2017-08-02
Snack bar queuing improvements and fixes.

### LINKED ISSUE #614: SnackbarMessageQueue Bug?
Hi There,

I may of found a bug in the SnackbarMessageQueue class. Can you please confirm if this is correct?

Cheers

### LINKED ISSUE #761: Possible wrong enqueue handling in SnackBarMessageQueue.cs
I tried to implement the very nice Snackbar feature in my application and run into problems. I wanted to show the message every time, even if it is the same. It seems that everything, to use this kind of behavior, is implemented but not used.

```
public void Enqueue(object content, bool neverConsiderToBeDuplicate)
        {
            if (content == null) throw new ArgumentNullException(nameof(content));

            _snackbarMessages.AddLast(new SnackbarMessageQueueItem(content));
            _messageWaitingEvent.Set();
        }
```

```
public void Enqueue(object content, object actionContent, Action actionHandler, bool promote)
        {
            if (content == null) throw new ArgumentNullException(nameof(content));

            _snackbarMessages.AddLast(new SnackbarMessageQueueItem(content, actionContent, actionHandler));
            _messageWaitingEvent.Set();
        }
```

In both cases the boolean variables, which to my understanding, are responsible for this kind of behavior  are not used.

Keboo:
Separated the concepts of promote and ignore duplicate. To allows these to be controlled independently.
Fixes #761
Fixes #614

### DIFF
--- a/MaterialDesignThemes.Wpf/ISnackbarMessageQueue.cs
+++ b/MaterialDesignThemes.Wpf/ISnackbarMessageQueue.cs
@@ -41,7 +41,7 @@ namespace MaterialDesignThemes.Wpf
         /// <param name="content">Message.</param>
         /// <param name="actionContent">Content for the action button.</param>
         /// <param name="actionHandler">Call back to be executed if user clicks the action button.</param>
-        /// <param name="promote">The message will promoted to the front ot the queue and never considered to be a duplicate.</param>
+        /// <param name="promote">The message will promoted to the front of the queue.</param>
         void Enqueue(object content, object actionContent, Action actionHandler, bool promote);
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


MaterialDesignThemes.Wpf/ISnackbarMessageQueue.cs
```diff
@@ -51,7 +51,21 @@ public interface ISnackbarMessageQueue
         /// <param name="actionContent">Content for the action button.</param>
         /// <param name="actionHandler">Call back to be executed if user clicks the action button.</param>
         /// <param name="actionArgument">Argument to pass to <paramref name="actionHandler"/>.</param>
-        /// <param name="promote">The message will promoted to the front ot the queue and never considered to be a duplicate.</param>
-        void Enqueue<TArgument>(object content, object actionContent, Action<TArgument> actionHandler, TArgument actionArgument, bool promote);
```
ButchersBoy: Hi @Keboo ...for the sake of backward compatibility, can we keep this method, but call your new method setting neverConsiderToBeDuplicate to equal promote.  That will also be in keeping with the existing method docs.

Keboo: Yes, and done.

ButchersBoy:
:+1:

PR #768, opened 2017-08-10
Updated PaletteHelper 

David-Loving224:
PaletteHelper uses the default SwatchProvider constructor inside non-virtual methods.  This stops custom colour swatches from being used by the SwatchProvider.  A second constructor for PaletteHelper has been created which allows the user to pass in the assembly containing their custom swatches and generate a private static SwatchProvider pointing at that assembly.

The changing of colours using the PaletteHelper in a project with a large number of resource dictionaries resulted in 10's of seconds on the resource dictionary search.  The second constructor allows the user to refine the starting point of the search by passing a resource dictionary into the constructor.

### DIFF
--- a/MaterialDesignThemes.Wpf/PaletteHelper.cs
+++ b/MaterialDesignThemes.Wpf/PaletteHelper.cs
@@ -3,6 +3,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using System.Text.RegularExpressions;
 using System.Windows;
 using System.Windows.Media;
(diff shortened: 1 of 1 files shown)


MaterialDesignThemes.Wpf/PaletteHelper.cs
```diff
@@ -12,6 +13,30 @@ namespace MaterialDesignThemes.Wpf
 {
     public class PaletteHelper
     {
+        private static ResourceDictionary refinedBrushDictionary;
```
Keboo: In general setting static members from instance constructors is a bad idea. I would suggest making both of these readonly instance fields. This will also mean changing `GetBrush` to either be an instance method or have it take in a `ResourceDictionary` as a parameter.

MaterialDesignThemes.Wpf/PaletteHelper.cs
```diff
@@ -12,6 +13,30 @@ namespace MaterialDesignThemes.Wpf
...
+        /// <summary>
+        /// Providing ability to support custom array of colour swatch in the defined assembly.
+        /// Increased performance on searching of colour brushes by limiting the resource dictionary search field.
+        /// </summary>
+        /// <param name="assembly"> The Assembly containing the custom colour swatches.</param>
+        /// <param name="brushDictionary">The Resource Dictionary containing the Material Design Brush definition.</param>
+        public PaletteHelper(Assembly assembly, ResourceDictionary brushDictionary = null)
+        {
+            provider = new SwatchesProvider(assembly);
+            refinedBrushDictionary = brushDictionary ?? Application.Current.Resources;
```
Keboo: I would put the same null propagation on `Application.Current` here as well.

MaterialDesignThemes.Wpf/PaletteHelper.cs
```diff
@@ -12,6 +13,30 @@ namespace MaterialDesignThemes.Wpf
...
+            refinedBrushDictionary = Application.Current?.Resources;
+        }
+
+        /// <summary>
+        /// Providing ability to support custom array of colour swatch in the defined assembly.
+        /// Increased performance on searching of colour brushes by limiting the resource dictionary search field.
+        /// </summary>
+        /// <param name="assembly"> The Assembly containing the custom colour swatches.</param>
+        /// <param name="brushDictionary">The Resource Dictionary containing the Material Design Brush definition.</param>
+        public PaletteHelper(Assembly assembly, ResourceDictionary brushDictionary = null)
```
Keboo: Given that the intent of this is to allow for more control over the `SwatchesProvider` that gets used. I would suggest having this constructor take in a `SwatchesProvider` rather than an `Assembly`. This will help reduce coupling between the two classes.

I would also suggest null checking the incoming value as well.

Keboo:
May be obsoleted by #923

Keboo:
I believe the updates done in #1239 handle the same problem that this was solving.

PR #773, opened 2017-08-18
Bugfix: the PopupBox resource dictionary wasn't exported.

adrianhara:


adrianhara:
Any input on this? Or could be merged?

ButchersBoy:
Well, I'm holiday... But there is a reason why it is like that:

All themes of standard WPF controls are exported. PopupBox is a custom control for this library. For library custom controls, default templates are set up in generic.xaml.

The PopupBox is special still, as it has a default style, but also the extra style for the FAB variant, so these are current pull-in on demand.

So, it's by design. But I'm thinking about it some more.

adrianhara:
I see. I'm wondering, what would be the disadvantage of exporting it? DLL size?

adrianhara:
Closed it as: nevermind 😸

Issue #775, opened 2017-08-19
24th Minute shows up as 0 on Clock Picker


CameronAavik:
I was looking through the code to see how the Clock Picker was done and noticed a small bug on review:
https://github.com/ButchersBoy/MaterialDesignInXamlToolkit/blob/master/MaterialDesignThemes.Wpf/Clock.cs#L320

On this line, if the value is 24, then it sends through as 0, however this is expected behaviour when we want to show the 24th minute. I reproduced this issue on the demo app:

![image](https://user-images.githubusercontent.com/5800980/29484628-fd14b89e-8505-11e7-9134-58e3433b67fd.png)

wongjiahau:
This is a real bug. I had tested it too.

Keboo:
Fixed by #1155

PR #776, opened 2017-08-19
Fixed issue #775

wongjiahau:
Refer issue #775

### DIFF
--- a/MaterialDesignThemes.Wpf/Clock.cs
+++ b/MaterialDesignThemes.Wpf/Clock.cs
@@ -317,8 +317,13 @@ namespace MaterialDesignThemes.Wpf
 				button.SetBinding(Canvas.LeftProperty, GetBinding("X", button));
 				button.SetBinding(Canvas.TopProperty, GetBinding("Y", button));
 
-				button.Content = (i == 60 ? 0 : i).ToString(format);
-				canvas.Children.Add(button);
+			    if (Is24Hours) {
+			        button.Content = (i == 60 ? 0 : (i == 24 ? 0 : i)).ToString(format);
+			    }
+			    else {
+			        button.Content = (i == 60 ? 0 : i).ToString(format);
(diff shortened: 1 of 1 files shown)


CameronAavik:
I think the reason that the original code handled 24 that way was because it was being used by the GenerateButtons() method (the one with no arguments) which will pass in 24 when Is24Hours is true. In this circumstance, we do want 24 to be represented as 0, not 24.

One solution to fix this would be to say that i = i % numItems. But this breaks when we use 12 hour time, which is when we want 1, 2, 3, ..., 12 to be shown and not 0.

So the actual solution instead should be passing down the correct values instead of fixing them inside this method.

wongjiahau:
@CameronAavik Thanks for the reminder, I just realized that.

MaterialDesignThemes.Wpf/Clock.cs
```diff
@@ -317,8 +317,13 @@ private void HourReadOutPartNameOnPreviewMouseLeftButtonDown(object sender, Mous
 				button.SetBinding(Canvas.LeftProperty, GetBinding("X", button));
 				button.SetBinding(Canvas.TopProperty, GetBinding("Y", button));
 
-				button.Content = (i == 60 ? 0 : (i == 24 ? 0 : i)).ToString(format);
-				canvas.Children.Add(button);
+			    if (Is24Hours) {
+			        button.Content = (i == 60 ? 0 : (i == 24 ? 0 : i)).ToString(format);
+			    }
+			    else {
```
mokeev1995: You've mixed identation. Switch to spaces.

wongjiahau: @mokeev1995 Er... Do you mind to tell me how to fix it? Because I'm not sure how to fix that

Keboo:
This was fixed by #1155

PR #785, opened 2017-09-01
Implemented StaysOpenProperty for DrawerHost

wongjiahau:
- Currently, I needed this property for my project, because I am using DrawerHost as an error reporting mechanism, so I needed a behavior where the drawer will not be closed even if user clicks outside of it.
- Although by setting the `IsEnabled = false` can achieve the same behavior, but it also disabled all the controls in the drawers, therefore I needed the StaysOpenProperty on DrawerHost
- Please do take a look at an example usage : 
image
In this scenario, I need the *Group-Clashing* hyperlink to be clickable by the user, but at the same time I also wants to disable user from closing it by clicking outside of the drawer.

### DIFF
--- a/MainDemo.Wpf/Drawers.xaml
+++ b/MainDemo.Wpf/Drawers.xaml
@@ -7,7 +7,7 @@
              xmlns:materialDesign="http://materialdesigninxaml.net/winfx/xaml/themes"
              mc:Ignorable="d" 
              d:DesignHeight="300" d:DesignWidth="300">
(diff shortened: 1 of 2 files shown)


wongjiahau:
Not only that, I also need this StaysOpenProperty because I uses DrawerHost for the following scenario : 
1. As an alternative to dialog box, because in some situation using dialog box is too aggressive. 
2. To display progress bar while loading something and also disabling UI at the same time.

PR #786, opened 2017-09-02
Add CloseOnClickAway property for drawer host

wongjiahau:
- Currently, I needed this property for my project, because I am using DrawerHost as an error reporting mechanism, so I needed a behavior where the drawer will not be closed even if user clicks outside of it.
- Although by setting the `IsEnabled = false` can achieve the same behavior, but it also disabled all the controls in the drawers, therefore I needed the CloseOnClickAway property on DrawerHost
- Please do take a look at an example usage : 
image
In this scenario, I need the *Group-Clashing* hyperlink to be clickable by the user, but at the same time I also wants to disable user from closing it by clicking outside of the drawer.

Not only that, I also need this CloseOnClickAway property because I uses DrawerHost for the following scenarios : 
1. As an alternative to dialog box, because in some situation using dialog box is too aggressive. 
2. To display progress bar while loading something and also disabling UI at the same time.

### DIFF
--- a/MainDemo.Wpf/Drawers.xaml
+++ b/MainDemo.Wpf/Drawers.xaml
@@ -7,7 +7,7 @@
              xmlns:materialDesign="http://materialdesigninxaml.net/winfx/xaml/themes"
              mc:Ignorable="d" 
              d:DesignHeight="300" d:DesignWidth="300">
-    <materialDesign:DrawerHost x:Name="DrawerHost" Margin="64" HorizontalAlignment="Center" VerticalAlignment="Center" BorderThickness="2" BorderBrush="{DynamicResource MaterialDesignDivider}">
+    <materialDesign:DrawerHost x:Name="DrawerHost" Margin="64" HorizontalAlignment="Stretch" VerticalAlignment="Center" BorderThickness="2" BorderBrush="{DynamicResource MaterialDesignDivider}">
         <materialDesign:DrawerHost.LeftDrawerContent>
             <StackPanel Margin="16">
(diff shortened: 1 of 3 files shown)


ButchersBoy:
Hi, thanks for this.

There's been a long term desire of mine (and a few other's requests) to add some sort of "pinned" option, where when pinned the template changes and the content is "pushed" out.

So the template in pinned more would be conceptually be like,

```
<Grid>
    <drawerContent />
    <mainContent />
</Gird>
```

I'm keen to follow semver, so I need to get the API right for this.  Including your change I think this would give us three distinct modes, so the new property would probably need to be an enum, something like:

```
OpenMode
    Standard,
    StaysOpen, // your new behaviour
    Pinned // pin, push main content
```

Are you interested in helping with this?

wongjiahau:
@ButchersBoy Yes, I'm interested to help with this. But I need to know what's the difference between Pinned and StaysOpen. By the way, does Standard means the drawer will close on click away?

adrianhara:
Is there a plan to also have a sort of "mini mode"? Where when only e.g. some icons are visible?

wongjiahau:
@ButchersBoy I think I get what is the meaning of pinned already (looks something like [this](http://react-toolbox.com/#/components/layout)). Let me figure out how to do that.

wongjiahau:
@adrianhara I don't think I will implement that yet.

adrianhara:
@wongjiahau maybe i'll look into implementing it, as I would need it.

PR #789, opened 2017-09-04
Added OpenMode property for DrawerHost

wongjiahau:
@ButchersBoy  I've already implemented the OpenModeProperty for DrawerHost, but I'm not sure if the API looks like what you want. 
Currently, the API looks like this : 
```XAML
<DrawerHost OpenMode="Pinned">
    <!---Content here--->
</DrawerHost>
```
demodrawerhost

_Issues : If OpenMode is set to Pinned, all the drawers works as expected, however only the BottomDrawer does not push the main content when it is opened. I'm not sure why though._

### DIFF
--- a/MainDemo.Wpf/Drawers.xaml
+++ b/MainDemo.Wpf/Drawers.xaml
@@ -104,16 +104,18 @@
                     <Button Command="{x:Static materialDesign:DrawerHost.OpenDrawerCommand}"
                         CommandParameter="{x:Static Dock.Bottom}"
                         Grid.Row="2" Grid.Column="1" Margin="4">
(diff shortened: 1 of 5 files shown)


Keboo:
TODO Items before release

- [ ] Add unit tests for drawer host
- [ ] Previous functionality allowed for individual control for light dismiss of drawers. Is this needed?
- [ ] Evaluate naming
- [ ] MoreThanOneDrawerOpened can be optimized for early return.
- [ ] UpdateMainContentMargin, evaluate simply creating new visual states. This could eliminated the Storyboard in the code, and allow for better re-templating.
- [ ] New functionality should be unit tested.

Keboo:
This was addressed by #2232 2232

PR #793, opened 2017-09-06
Bumping the VS version in the solution file to VS 2017.

### LINKED ISSUE #788: vs2017 required but version in .sln files is vs2015
Hello!

I cloned the repo, and noticed that the different .sln at the root will open vs2015 by default, even though vs2017 is required to build, because of the header:

```
Microsoft Visual Studio Solution File, Format Version 12.00
# Visual Studio 14
VisualStudioVersion = 14.0.23107.0
MinimumVisualStudioVersion = 10.0.40219.1
```

It's not a big problem, but would be nice to get a fix :)

Thanks!

Keboo:
Fixes #788

### DIFF
--- a/MaterialDesignToolkit.Wpf.sln
+++ b/MaterialDesignToolkit.Wpf.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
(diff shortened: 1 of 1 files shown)


ButchersBoy:
Closed as another PR bumped it to 15

PR #797, opened 2017-09-11
Exposed SelectedTimeChanged event from TimePicker

adrianhara:
In case one is wrapping the TimePicker in a composite (custom) control, it's pretty useful to be able to subscribe to changes in order to propagate them to the outside world (maybe as a composite value from multiple controls.).

### DIFF
--- a/MaterialDesignThemes.Wpf/TimePicker.cs
+++ b/MaterialDesignThemes.Wpf/TimePicker.cs
@@ -10,244 +10,250 @@ using MaterialDesignThemes.Wpf.Converters;
 
 namespace MaterialDesignThemes.Wpf
 {
-	[TemplatePart(Name = ElementButton, Type = typeof(Button))]
-	[TemplatePart(Name = ElementPopup, Type = typeof(Popup))]
-	[TemplatePart(Name = ElementTextBox, Type = typeof(DatePickerTextBox))]
-	public class TimePicker : Control
-	{
-		private const string ElementButton = "PART_Button";
-		private const string ElementPopup = "PART_Popup";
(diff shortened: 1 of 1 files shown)


adrianhara:
@ButchersBoy what do you think?

MaterialDesignThemes.Wpf/TimePicker.cs
```diff
@@ -74,6 +76,8 @@ private static void SelectedTimePropertyChangedCallback(DependencyObject depende
 			timePicker.SetCurrentValue(TextProperty, timePicker.DateTimeToString(timePicker.SelectedTime));
             timePicker._isManuallyMutatingText = false;
             timePicker._lastValidTime = timePicker.SelectedTime;
+
+		    timePicker.SelectedTimeChanged?.Invoke(timePicker, new EventArgs());
```
Keboo: Personally I think it would be better to do this as a RoutedEvent with its own custom event args class that contains the new value. Very similar to how [TextBoxBase.TextChanged](https://msdn.microsoft.com/en-us/library/system.windows.controls.primitives.textboxbase.textchanged.aspx) works.

adrianhara: I changed it to be a routed event. I originally made it a standard event by looking at the framework DatePicker, but sure, routed event works as well 👍 Sorry it took 7 months, wasn't easy you know 😁

Keboo: Great work, however, there was another PR that has already resolved this issue. Thank you for your willingness to contribute to this project!

Keboo:
Already resolved by #1059

PR #815, opened 2017-10-09
Keyboard tab now ignores the Expander header

llifon:
In the current `Expander` style, the header consists of two parts: the header text button and the header arrow button. Whenever a click event happens on any of these, the `Expander` opens. When navigating with the tab (via the keyboard), the following happens (expander of the left):

Current tabbing

... that is, to navigate past an `Expander`, tab needs to be pressed twice. I believe it would make a lot more sense for only the arrow button to grab focus and show a focus visual style. The change I made results in the following look:

New tab focus


ButchersBoy:
Fixes 818

PR #821, opened 2017-10-12
Fixes #820 - Opacity of the old transition slide should now always be set to 0

### LINKED ISSUE #820: Inactive transitioner slide still visible behind the active slide
> .NET Framework 4.5.1
> MaterialDesignColors 1.1.2
> MaterialDesignThemes 2.3.1.953
> Windows 10

I've noticed when using `Transitioner` with content which has a transparent background, that the previous slide is not being hidden. The transparency has nothing to do with the cause of this issue, it was just the means that helped identify it.

This is the behaviour I'm seeing:

slides-not-being-hidden

I've created a repository which reproduces this issue: [transition issue repository](https://github.com/llifon/mdx-transition-issue)

llifon:
Hopefully this fixes #820 under all circumstances. Previously, the old slide's opacity was only being set to 0 under very specific circumstances, rather than every time. With this change, the behaviour changes to:

slides-being-hidden

rather than what was demonstrated in #820 

slides-not-being-hidden

### DIFF
--- a/MaterialDesignThemes.Wpf/Transitions/Transitioner.cs
+++ b/MaterialDesignThemes.Wpf/Transitions/Transitioner.cs
@@ -197,17 +197,23 @@ namespace MaterialDesignThemes.Wpf.Transitions
             }
 
             if (newSlide != null)
-                newSlide.Opacity = 1;                
+            {
+                newSlide.Opacity = 1;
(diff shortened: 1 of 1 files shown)


ButchersBoy:
Thank you!

PR #823, opened 2017-10-14
Fixes expander expandsite sizing to content, instead to its expander

### LINKED ISSUE #803: Unlimited Content Height/Width in Expander Control
I noticed this issue when placing a ScrollViewer inside of MaterialDesign's Expander. The ScrollViewer does not work inside of the Expander due to it having infinite height/width. This behavior is equivalent to putting a ScrollViewer in a grid with the height set to "auto" instead of "*". 

This works perfectly when using the default Expander. I tried numerous workarounds before trying to reproduce it with the bare minimum and realizing it was the MaterialDesign Expander.

Demo: https://github.com/JonathanSchafer/DemoMaterialDesignExpanderContentSizingIssue

Insire:
fixes #803 - where the expander is sizing to content, instead of the container


Insire:
From what I know and what i tested, we have the choice either to keep the current approach and live with the failing scrolling since the actual content in the expander is not resized when its being collapsed, or we can support the scrolling but live without the animation.

The following seems to be happening in the master branch:
When the expander is being closed, its content is being made invisible, by updating visibility and opacity properties. The container around the content is being resized and thats visible in the UI. But the content presenter keeps its size, since it only cares about the size of its content, not about its parents size.

What i did in the fix:
Remove all custom sizing and scale the content presenter according to its parent, which resizes everything and lets scrollviewers work properly, when they are introduced by a user.


... several minutes later
</br>

I just did some more fiddling, but in the end, i need to know what size the content will be, so i can bind to it and set that either explicit or implicit via the multiplyConverter. But there is no actual property to bind to, when the size is determined by the parent container. That means until the expander is expanded, i dont know what to use as target property for the animation, once i know, its too late to animate the expansion.

I think to support both, animation and scrollviewer, we'd have to abandon the scope of this library and write custom controls. Atleast for this case.

Insire:
Maybe, if we could set a predetermined size for the content ...

.... _indistinguishable muttering_

ButchersBoy:
Hi, just thinking out loud here...but maybe we could use some kind of attached property, to generate the storyboard in code, according to content size (which may or may not be infinite), allowing us to support infinite/fixed size with/without scrolling...

Insire:
I gonna fiddle with that idea a bit by the end of the week

Keboo:
Closing this as I believe this is superseded by #1434

PR #824, opened 2017-10-17
Updated default remote path to use the original repo as source

wongjiahau:
@ButchersBoy I have updated the default remote path for the new demo app to use the original repository as the source of loading the source file. 
Because currently, the new demo app will load the file from my repo, so if I had changed anything to my repo, the user won't be able to load the source file properly.
Therefore I changed the default remote path to use this original repository as the source of loading files.


ButchersBoy:
Made a change to this to try and load locally.

wongjiahau:
Noted.

PR #829, opened 2017-10-19
Creating a new Circular Progress Bar Style

axd2201:
I only changed the file "MaterialDesignTheme.ProgressBar.xml"
I copied the whole normal style, and just removed the RotateTransform code, but i am not familiar with the project enough to properly remove it, it seems to be used in another class that i dont know of, so i just commented it out and added a TODO to it

Example of the progress bar without the rotate animation
https://streamable.com/vwnwt


Keboo:
Hi @axd2201 I have created an alternate PR #1095 that should supersede this one. Since the rotating of deterministic circular progress indicator does not match the [spec](https://material.io/design/components/progress-indicators.html#circular-progress-indicators) I have implemented a similar change as you proposed in the `MaterialDesignCircularProgressBar` style.

PR #831, opened 2017-10-26
Removed hit test from switch ripple

jizc:
When stacking multiple switches, the ripples can overlap other switches. Since the ripple was hit test visible (even when the ripple appears to be gone), one could accidentally click the wrong switch.


ButchersBoy:
Thank you!

PR #834, opened 2017-10-31
Fix NullReferenceException in ListBoxAssist

dkjonas:
Clicking near the edge of a control that uses ListBoxAssist, creates an NullReferenceException, because VisualTreeHelper.HitTest may return null in that case.


ButchersBoy:
Thank you! :+1:

PR #837, opened 2017-11-05
Update SmartHint styles to better match the material design specifica…

JeffBarnard:
Hello again,

I have updated the SmartHint styles to better match the design specification. For example, when using floating hints and the control is focused (and editable), the hint is now immediately raised and colored.

image

image

### DIFF
--- a/MaterialDesignThemes.Wpf/SmartHint.cs
+++ b/MaterialDesignThemes.Wpf/SmartHint.cs
@@ -6,7 +6,6 @@ using System.Text;
 using System.Threading.Tasks;
 using System.Windows;
 using System.Windows.Controls;
-using System.Windows.Media.Animation;
 using MaterialDesignThemes.Wpf.Converters;
 
 namespace MaterialDesignThemes.Wpf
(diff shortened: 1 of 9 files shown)


MaterialDesignThemes.Wpf/SmartHint.cs
```diff
@@ -17,15 +16,13 @@ namespace MaterialDesignThemes.Wpf
...
+    [TemplateVisualState(GroupName = ContentStatesGroupName, Name = HintRestingPositionName)]
+    [TemplateVisualState(GroupName = ContentStatesGroupName, Name = HintFloatingPositionName)]
     public class SmartHint : Control
     {        
         public const string ContentStatesGroupName = "ContentStates";
-        public const string ContentEmptyName = "ContentEmpty";
-        public const string ContentNotEmptyName = "ContentNotEmpty";
-
-        private ContentControl _floatingHintPart = null;
+        public const string HintRestingPositionName = "HintRestingPosition";
```
ButchersBoy: There is a small issue with renaming this.  They are part of the public API, and we are following semver.  Admittedly, I'd be a but surprised if people were accessing this in their code, but i'd like to stick to semver wherever possible and don't want to tick a major version.

Can we please rename them back.  Which might also mean that ``` IsHintInFloatingPosition ``` might need to be named in according with the original state names.

JeffBarnard:
Sorry I'm not that good with GitHub, did I make the requested changes available?

ButchersBoy:
Thank you for this! :+1:

Issue #798, opened 2017-09-11
MaterialDesignTextFieldBoxTextBox not complete to spec


bebenins:
[Spec here in short gif](https://material.io/guidelines/components/text-fields.html#text-fields-text-field-boxes)

- There should be bottom border (clipped on each side) even when it's not in focus.
- There should be a shadow indicting mouse hover over.
- There should be a ripple effect when focused.
- Hint should automatically move up when focused.

wongjiahau:
@bebenins Point 4 is actually implemented, try the code below : 
```xml
<TextBox 
	materialDesign:HintAssist.Hint="Floating Hint"
	Style="{StaticResource MaterialDesignFloatingHintTextBox}" />
```

For point 1 to point 3, I think these are the latest specification, should be implemented soon.

wongjiahau:
@ButchersBoy Do you plan to implement this soon? Because I thought their latest spec is more user friendly, as I recently just found out that some user are actually not familiar with the single-lined textbox, they don't feel like it's a textbox.

Keboo:
Fixed by #840

PR #840, opened 2017-11-09
Better hover and focus effects for text areas and text field boxes

spiegelp:
Better hover and focus effects for text areas and text field boxes
solves #798

The new styles fit the specs a little bit better. However text, screenshots, videos and [Material Components Web](https://material-components-web.appspot.com/textfield.html) contradict each other in details. So it is a tradeoff.


Keboo:
@spiegelp this is looking pretty good. 
Would you addressing the current conflicts so it can get merged?

spiegelp:
@Keboo I resolved the conflicts.

PR #857, opened 2017-12-05
add ValidationStep UpdatedValue to All NotEmptyValidationRules

beyazc:
i think adding ValidationStep="UpdatedValue" to NotEmptyValidationRule might be helpful. I spent sometime for searching why models property was not changing if validationresult returns false. Adding validationstep solved issue


cahitbeyaz:
Commenting is also fine. 
thanks for review

PR #868, opened 2017-12-30
Implementing ShowMeTheXAML

Keboo:
Implemented a new method for displaying XAML in the demo application. 
I moved the demo application back to its original layout.

Here are a few screenshots as a demo.
image

image

The XamlDisplay control is just a simple `ContentControl`. When specifying it in the XAML you much provide a unique value for the Key property. Failing to do so, or specifying a duplicate key, will result in a compile error. XamlDisplay also contains a Xaml property that is the string representation of the XamlDisplay element along with all of the contents. The control uses its Formatter property to format and style the string for display. There is a [default implementation](https://github.com/Keboo/ShowMeTheXAML/blob/master/ShowMeTheXAML/XamlFormatter.cs) with some basic settings.


ButchersBoy:
Really like this direction now.

Just checkout out your branch, when running in debug I'm not seeing the XAML:

image

Keboo:
@ButchersBoy Interesting. It would appear the issue is with my MSBuild target. On a fresh clone, just launching with F5 I can reproduce the issue. However after I do a rebuild, the problem disappears :/. I suspect it is my MSBuild target. I will track it down and get it fixed. In the meantime if you want to test further you should be able to just do a rebuild and get stuff to show up.

wongjiahau:
@Keboo I love this layout!

wongjiahau:
@Keboo Actually I found out this layout can be implemented easily by modifying some codes in the CodeDisplayer (you just have to change the code of [this file](https://github.com/wongjiahau/Displaying-XAML/blob/master/XamlDisplayer/CodeDisplayer/XamlDisplayer.xaml)), therefore I'm planning to open an API that allow user to use their customize layout by inheriting LayoutBuilder, perhaps it will looks something like this : 
```cs
/// in App.cs
public class MyLayoutBuilder: ILayoutBuilder { 
    public void Build(FrameworkElement element, FrameworkElement code) {
        // just for example 
        var sp = new StackPanel() { Orientation = Orientation.Horizontal};
        sp.Children.Add(element);
        sp.Children.Add(code);
    }
}
```
I'm not sure if this is doable using XAML (of course it would be better if it can be done), so what do you think about this @Keboo ?

Keboo:
@wongjiahau 
A couple thoughts. Typically in WPF layout is controlled by specifying a template for a given control. Though your idea would certainly work, it is not what is expected. A better approach would be to [create a custom control](http://wpftutorial.net/HowToCreateACustomControl.html), with a predefined template that people could then override with whatever template they like. This would let you not only define your layout in XAML, but allow other users to change the template to suit their needs.
This is how my library was done. [A custom control](https://github.com/Keboo/ShowMeTheXAML/blob/master/ShowMeTheXAML/XamlDisplay.cs#L18) with a default style key set. Then a very [simplistic base template](https://github.com/Keboo/ShowMeTheXAML/blob/master/ShowMeTheXAML/Themes/Generic.xaml). Finally a [cleaner, better looking template](https://github.com/Keboo/MaterialDesignInXamlToolkit/blob/smtx/MainDemo.Wpf/App.xaml#L22-L95) to get the desired look.

Keboo:
@ButchersBoy so after a little digging it appears that the problem is coming because of the packet restore (I am still a little new to packet so perhaps I am missing something). 

So here is what I have so far:
1. (Close VS if it is open). Recreate a fresh clone by removing all untracked files `git clean -fdX`
2. Open VS and press F5
3. The targets file from ShowMeTheXAML.MSBuild is [referenced by the csproj](https://github.com/Keboo/MaterialDesignInXamlToolkit/blob/smtx/MainDemo.Wpf/MaterialDesignDemo.csproj#L493) but the exists condition is evaluated as false because it is not on disk.
4. The [packet targets file] contains an MSBuild target (RestorePackages) which kicks off and restores the packages, but because the import from the previous step was already evaluated it never kicks off my MSBuild target to generate the XAML.
5. Doing a rebuild (or making a change and doing a normal build) will kick off the ShowMeTheXAML.MSbuild task because it now exists on disk.

As of this point I am not sure how to resolve this issue. Though I think it is worth clarifying. The issue only occurs inside of visual studio on initial build when dependencies have not been restored.
Because the appveyor.yml script [restores packages before MSBuild runs](https://github.com/Keboo/MaterialDesignInXamlToolkit/blob/smtx/appveyor.yml#L18) it should not be a problem for CI builds. 
So at this point I am not sure how to solve it (assuming that there is a valid solution). It might be worth checking in at least the targets file in packages\ShowMeTheXAML.MSBuild\build\net45 simply so that it is present for initial build. I have not tested this yet (it is late and I need sleep), but given the above order of operations, it should "fix" the issue.

Thoughts?

wongjiahau:
@Keboo I think @punker76 knows how to solve this problem, he helped me with this same issue last time.

Keboo:
@wongjiahau made some nice additions to this PR. Adding his screenshot here for reference as well. 
image

Keboo:
@punker76 I am still seeing the initial issue. As described above, it only occurs on a fresh clone, and only on the first. This should not affect the appveyor build since it does the restores the packed dependencies prior to running MSBuild.
Specifically the steps in question:
1. Do a fresh clone (or run `git clean -fdX`)
2. Open Visual Studio and select MaterialDesignDemo as the startup project
3. Press F5

After my latest changes. This will now show an error message indicating SMTX was setup and a rebuild is required (it simply checks to see if the SMIX output file had been set. Just a simple check to indicate if SMTX's targets file was loaded). 

From my research (and questioning some MSBuild experts), I don't think there is an easy way to solve this. Here are the steps that occur when pressing F5 for the first time.
1. MSBuild starts and loads the csproj file. 
2. All of the Imports are treated as though the imported file was copied in at that location. Conditions on the Imports are evaluated. For a clean MDIX repository this means that it loads the paket targets file, but skips the import on the SMTX targets file because it has not been downloaded by paket yet. Because the SMTX targets file is not imported, it never gets to run.
3. MSbuild constructs the Targets and the order they will be executed.
4. The targets are executed. At this point paket will download SMTX. However, at this point it is too late since MSBuild has already built up the targets to execute (this is untimely the part that I am not sure how to work around). 

Given that this is the only case that is affected. I added a simple target to the csproj that merely displays an error indicating that the project needs to be recompiled. Again, this should only ever show up on the first build on a clean repository. After that everything should just work.

ButchersBoy:
Great work on this guys.  Thanks @Keboo, looks great.  And thanks @wongjiahau for starting us down this path.

:100: :heart:

wongjiahau:
@ButchersBoy You're welcome :)

Issue #570, opened 2016-12-08
HintAssist.FloatingScale and ComboBox


pinki:
Hi,
is there a way to get FloatingScale work with ComboBoxes?

jr2424:
I believe this is related to issue #547. I do not believe the MaterialDesignTheme.ComboBox.xaml wpf:SmartHint is binding to the appropriate properties on the TemplatedParent. The FloatingOffset is also not working due to this.

Keboo:
Also #988

Keboo:
Resolved by #899

PR #899, opened 2018-02-17
Added Scale and Offset to the combobox template

glen-nicol:
This allows the attached properties to be put on comboboxes and affect the hint. Currently it has the side effect of the hint being cut off when the combobox is expanded and the hint is too big. Any hints on how I can fix that?


soi013:
I guess this solve #570 .

Issue #845, opened 2017-11-16
Transitioner NullReference


Symbai:
Using 2.3.1 version:
~~Randomly when dragging a new control (grid, listbox, textbox, button, doesn't matter) onto my form a nullreference exception in the designer occurs.
I don't drag the controls on the transitioner, it's just enough for it to be on my application at all.~~

Actually the issue happens randomly but quite often. Happens when dragging controls on designer, happens when exiting from debugging. Happens when controls in designer get changed.

```
NullReferenceException: Object reference not set to an instance of an object.

 at MaterialDesignThemes.Wpf.Transitions.Transitioner.ActivateFrame(Int32 selectedIndex, Int32 unselectedIndex)
   at MaterialDesignThemes.Wpf.Transitions.Transitioner.<.ctor>b__14_0(Object sender, RoutedEventArgs args)
   at System.Windows.RoutedEventHandlerInfo.InvokeHandler(Object target, RoutedEventArgs routedEventArgs)
   at System.Windows.EventRoute.InvokeHandlersImpl(Object source, RoutedEventArgs args, Boolean reRaised)
   at System.Windows.UIElement.RaiseEventImpl(DependencyObject sender, RoutedEventArgs args)
   at System.Windows.UIElement.RaiseEvent(RoutedEventArgs e)
   at System.Windows.BroadcastEventHelper.BroadcastEvent(DependencyObject root, RoutedEvent routedEvent)
   at System.Windows.BroadcastEventHelper.BroadcastLoadedEvent(Object root)
   at MS.Internal.LoadedOrUnloadedOperation.DoWork()
   at System.Windows.Media.MediaContext.FireLoadedPendingCallbacks()
   at System.Windows.Media.MediaContext.FireInvokeOnRenderCallbacks()
   at System.Windows.Media.MediaContext.RenderMessageHandlerCore(Object resizedCompositionTarget)
   at System.Windows.Media.MediaContext.RenderMessageHandler(Object resizedCompositionTarget)
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
```

olegsavelos:
This is still happening for me, was this resolved ?

Keboo:
[...]

PR #902, opened 2018-02-25
Added async handling around the Transitioner container generator

glen-nicol:
I think this fixes issue #845.

I ran into this when trying to bind the transitioner to an observablecollection and was always getting null returned from the generator call in GetSlide. I would also get a null when I used bare controls in the transitioner without a slide wrapped around it which the [Wiki page](https://github.com/ButchersBoy/MaterialDesignInXamlToolkit/wiki/Transitions#transitioner) says it will do for you. This change fixes that too.


Keboo:
I believe this issue was addressed a slightly different way. Going to close this for now.

PR #903, opened 2018-02-26
Add MaterialDesignThemeResources

mycroes:
Allows easy setup of MaterialDesignTheme defaults using BaseTheme,
PrimaryColor and AccentColor properties.

### DIFF
--- a/MainDemo.Wpf/App.xaml
+++ b/MainDemo.Wpf/App.xaml
@@ -9,13 +9,8 @@
              xmlns:materialDesignDemo="clr-namespace:MaterialDesignDemo"
              StartupUri="MainWindow.xaml">
     <Application.Resources>
-        <ResourceDictionary>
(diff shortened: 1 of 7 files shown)


mycroes:
Not sure if you like this. Either Intellisense or R# actually doesn't like the App.xaml change entirely, but I couldn't figure a better way that's compatible with the current `PalleteHelper` implementation. When creating a `MaterialDesignApplication : Application` class I had issues with undefined resources, having the `BaseTheme` defined one `MergedDictionary` deeper renders the `PalletHelper` unable to find it (fixable though).

I'm also intending to add a ResourceDictionary for control styles so that you can do
```
<ResourceDictionary>
    <materialDesign:ControlResources Control="Button" />
</ResourceDictionary>
```

Keboo:
Moving this into 2.6.0 as there is going to be a bunch a work around colors and theming.

Keboo:
This is now handled with the new theme API.

PR #905, opened 2018-03-05
Adding the possibility of DialogHost being dark

equimal:
Removed force Light theme

### DIFF
--- a/MaterialDesignThemes.Wpf/Themes/Generic.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/Generic.xaml
@@ -613,13 +613,6 @@
                         <controlzEx:PopupEx PlacementTarget="{Binding ElementName=DialogHostRoot, Mode=OneWay}"
                                             x:Name="PART_Popup"
                                             Style="{TemplateBinding PopupStyle}">
-                            <controlzEx:PopupEx.Resources>
-                                <ResourceDictionary>
-                                    <ResourceDictionary.MergedDictionaries>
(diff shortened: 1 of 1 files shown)


Tide:
I also changed
`<Setter TargetName="PART_ContentCoverGrid" Property="Background" Value="Black" />`
to
`<Setter TargetName="PART_ContentCoverGrid" Property="Background" Value="{DynamicResource MaterialDesignBody}" />`

and added
`<ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Light.xaml" />`
to the merged dictionaries.

Now the overlay will change as well.

SaviorXTanren:
Can't wait for this fix :)

Keboo:
Putting this for consideration with the 2.5.0 release as I know this has been requested several times. My only issue with it, is this type of change is technically a breaking change for any existing app that assumes the previous behavior. Will need to consider options.

Keboo:
@CroveanOS I assume you are referring to the [PaletteHelper.SetLightDark method](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MaterialDesignThemes.Wpf/PaletteHelper.cs#L15). It actually does work, but the default value for the `DialogTheme` property is `Light` (this is for backwards compatibility). If you change the value to `Inherit` your dialogs will follow the theme set by the PaletteHelper.

PR #918, opened 2018-03-20
Added Clear button to TextBox and ComboBox https://goo.gl/images/ZCj8uh

beppemarazzi:


jespersh:
Is it on or off by default? I do see this line ```<Setter Property="wpf:TextFieldAssist.HasClearButton" Value="True" />```

Adding to the Wpf demo would be a nice addition showing it off

beppemarazzi:
You are right. Now it is on by default because of the line you pointed.
Probably you are right: the best is to put the default off and add one use case in the Demo to show the usage... i'll proceed soon.
Meanwhile i fixed a little issue with tab-focus navigation.
And last but not least: many many thanks for your work!!!

Issue #901, opened 2018-02-20
Incompatibility with MahApps 1.6.0 version


macxfadz:
Please examine following **App.xaml**

```
<Application x:Class="DatGen_FrontEnd.App"
             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
             xmlns:local="clr-namespace:DatGen_FrontEnd">
    <Application.Resources>
        <ResourceDictionary>
            <ResourceDictionary.MergedDictionaries>
                <ResourceDictionary Source="pack://application:,,,/MahApps.Metro;component/Styles/Controls.xaml" />
                <ResourceDictionary Source="pack://application:,,,/MahApps.Metro;component/Styles/Fonts.xaml" />
                <ResourceDictionary Source="pack://application:,,,/MahApps.Metro;component/Styles/Colors.xaml" />
                <ResourceDictionary Source="pack://application:,,,/MahApps.Metro;component/Styles/Accents/Blue.xaml" />
                <ResourceDictionary Source="pack://application:,,,/MahApps.Metro;component/Styles/Accents/BaseDark.xaml" />
                <ResourceDictionary Source="pack://application:,,,/MahApps.Metro.IconPacks;component/Themes/IconPacks.xaml" />

                <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Dark.xaml"/>
                <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Defaults.xaml" />
                <ResourceDictionary Source="pack://application:,,,/MaterialDesignColors;component/Themes/Recommended/Primary/MaterialDesignColor.BlueGrey.xaml" />
                <ResourceDictionary Source="pack://application:,,,/MaterialDesignColors;component/Themes/Recommended/Accent/MaterialDesignColor.LightBlue.xaml" />

                <!-- Material Design: MahApps Compatibility -->
                **<ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.MahApps;component/Themes/MaterialDesignTheme.MahApps.Defaults.xaml "/>**
                <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.MahApps;component/Themes/MaterialDesignTheme.MahApps.Fonts.xaml" />
[...]

PR #898, opened 2018-02-16
WIP Update MaterialDesignTheme.MahApps for MahApps v1.6

punker76:
- [x] Update paket and dependencies for MahApps.Metro v1.6
- [x] Update nuspec dependencies
- [x] Update Flyout style
- [ ] Update RangeSlider style
- [ ] Adjust Dialog stuff

Issue #864, opened 2017-12-15
Feature request. Control colors specided by Dark or Light thems.


Vick-edit:
I have a lot of charts in my project, Most charts have color with transparency that must be sensitive ot light/dark theme. I have ot use some workaround to convert stored theme's brushes to colors for creating new brushes with transparency level.

If there were base colors in addition to brushes - it would be very useful.
[MaterialDesignTheme.Dark.xaml](https://github.com/ButchersBoy/MaterialDesignInXamlToolkit/blob/master/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Dark.xaml#L5)

Or here we can add ability to swap users own dark/ligh them brushes
[PaletteHelper.cs](https://github.com/ButchersBoy/MaterialDesignInXamlToolkit/blob/master/MaterialDesignThemes.Wpf/PaletteHelper.cs#L15-L41)

If it corresponds to your ideas about how your library should work, I could add this functionality.

Keboo:
Resolved by #923

Issue #853, opened 2017-11-23
Automatically apply themes when package is installed


ortund:
I love this toolkit but I always end up scratching my head for 20 minutes wondering why it isn't working right. That's normally because I've forgotten to add a resource somewhere.

Was wondering if it'd be possible to have the resources all added appropriately to app.xaml and such files when the package is installed.

PR #923, opened 2018-03-22
*WIP* Better theming and brush usage

Keboo:
@ButchersBoy @punker76 This is a work in progress, but I wanted to get feedback before I put in more work to make sure it is something that is viable. I would like to credit @mycroes with #903 for giving me the idea.

## Objectives

- Make the colors/themes more extensible. This would be helpful for integrations with MahApps/Dragablz. It would also allow for others to release NuGets that provide integrations with other libraries. 
- Make it easier for people to get started without needing to look up examples of which resource dictionaries need to get merged in. Even I cannot quite remember, and end up always going back to another app to copy them in.
- Easier swapping of theme/color resources within the app.

## Points of interest
- (Option 1) App.xaml in the demo app. Rather than needing to include the MDIX dictionaries simply using a new derived resource dictionary class, it adds in all of the dictionaries for you. I defaulted the values to the same as the demo app. My thought is that it is better for it to default to the wrong colors, and have the programmer figure out how to change them, rather than not loading the colors and everything look horrible (or like it is not working). This dictionary is just a simple XAML way to access the extension method below. It exposes some enums as references to the various recommended color themes.
- (Option 2) App.xaml.cs in the demo app. There is an extension method on `Application` called `WithMaterialDesign` (MaterialDesignAssist.cs). This does all of the same stuff as the previous option, but it allows for a fluent syntax that can also be used to extend MDIX. If you look at the App.xaml.cs in the MahApps demo application, you will see that there is another extension method (MaterialDesignThemes.MahApps.MaterialDesignAssist.cs) that adds in all of the stuff for MahApps (this is just feature parity with the old stuff, nothing for MahApps 1.6).
- MaterialDesignColors.Wpf.RecommendedThemeProvider.cs previously it has always felt a little clunky trying to get access to complete "themes" (like the dictionaries you would previously have to set in App.xaml) rather than just swatches. This class exposes those in a similar manner making it a bit easier to work with. 
- PaletteHelper.cs. This class has had all references to MahApps stripped out of it in favor of the fluent syntax for extensibility. I have added events that would allow for other libraries/apps to hook into changes and respond appropriately. 
- ThemeAssist.cs. This class adds attached properties that allow changing the theme/color at a lower level in the tree. Currently the enums don't have a default or "none" options and perhaps they should to allow for these properties to remove a previously added dictionary.

## Known Issues/TODOs
- There are breaking API changes in PaletteHelper. Specifically I changed the class from an instance to a static. I am happy to go back and add additional overloads and whatnot to make this a non-breaking change if that is desired.
- All of the MahApps references have been moved from the main MDIX project into the MaterialDesignThemes.MahApps (arguably where they should be). This is a breaking change.
- PaletteHelper would get another event for Accent swatch changed.
- R# is not happy about the fact that it cannot find resources. This is because they are injected at run-time. Not sure I am completely happy with this. It really only matters for the control styles since the colors/brushes are typically used with `DynamicResource`.
- There are a lot of places where the MDIX brush keys are just hard coded in the code. I would like to at least move these to constants in the colors library.
- There are enums for the various color themes. These need to be moved to the colors library.
- I unloaded the MaterialDesignThemes.Wpf.Tests project rather than deal with the errors there.

Related Issues/PRs: #901 #898 #864 #853 #903

### DIFF
--- a/MahMaterialDragablzMashUp/App.xaml
+++ b/MahMaterialDragablzMashUp/App.xaml
@@ -2,7 +2,6 @@
              xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
              xmlns:dragablz="clr-namespace:Dragablz;assembly=Dragablz"
-             xmlns:options="http://schemas.microsoft.com/winfx/2006/xaml/presentation/options"
              StartupUri="MainWindow.xaml">
     <Application.Resources>
         <ResourceDictionary>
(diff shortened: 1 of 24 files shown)


mycroes:
Thanks for the attribution! I like how you added the fluent interface on App, which would make the following so damn near perfect:

```csharp
public class Program
{
    [STAThread]
    public static void Main(string[] args)
    {
        var app = new App {ShutdownMode = ShutdownMode.OnLastWindowClose};
        app.InitializeComponent();
        app
            .WithMaterialDesign(BaseTheme.Light, PrimaryColor.DeepPurple, AccentColor.Lime)
            .Run(new MainWindow {DataContext = new AppBootstrapper()});
    }
}
```

Also (not to take more credit than I deserve), I stumbled upon my idea for #903 when I again forgot the resource uri for my own library and then figured I could apply it to MaterialDesignInXaml as well (for which I also forget the resource uri's all the time). I've never seen anyone do this sort of thing (didn't search for it either), but maybe we can get JetBrains to support it in R# (in a generic way, not tailored at MaterialDesignInXaml explicitly).

Anyway, I really like the way this simplifies the usage of MaterialDesignInXaml, can't wait to see this in production.

Keboo:
Superseded by #1239

PR #938, opened 2018-04-23
Better handling of custom themes

wexman:
- Allow creating a SwatchProvider specifying more than one assembly with themes
- Allow passing those SwatchProviders to PaletteHelper instances

This allows me to create a custom theme inside my application and still choose whether to use the custom one or one of the default themes...


Keboo:
I believe this is now handled by #1239

Issue #679, opened 2017-04-12
Add support for ComboBox in color zones


bebenins:
Perhaps same could be merged into TextBox?
You can download repo with example here (look in ColorZones screen):  https://github.com/bebenins/MaterialDesignInXamlToolkit

Basically here's the issue...  MD guidelines show that you should be able to add "accent" controls in color zones:
https://material.io/guidelines/style/color.html#color-color-system
![image](https://cloud.githubusercontent.com/assets/16761301/24979091/fc2451e4-1fa0-11e7-9164-58c515ab811c.png)

Here's a ComboBox example from Google+ app:
![image](https://cloud.githubusercontent.com/assets/16761301/24979120/1386f27e-1fa1-11e7-96a4-642a9af88237.png)
![image](https://cloud.githubusercontent.com/assets/16761301/24979141/1f0f3bec-1fa1-11e7-9f51-8a3a98c5b974.png)

But we have issues:
- no accent color
- when a control is in focus in some ColorZones, underline is invisible
- ComboBox list is same color as ColorZone
- Text does not reflect ColorZone in ComboBox list (should take care of itself when list is Card color)
![image](https://cloud.githubusercontent.com/assets/16761301/24979161/320fc9a0-1fa1-11e7-94d5-c30ac70b8920.png)
![image](https://cloud.githubusercontent.com/assets/16761301/24979263/8675aa46-1fa1-11e7-882b-c6b3fde68034.png)
![image](https://cloud.githubusercontent.com/assets/16761301/24979272/8bcae16e-1fa1-11e7-9146-af893943b015.png)

PR #942, opened 2018-05-02
Support use of ColorZoneAssist in ComboBox

NZSmartie:
First part is addressing #679. Only when the ComboBox displays it's pop-over will the ColorZone take effect.
This does not change the border away from the primary mid brush.

Things to do:

- [X] Set the background and foreground when the popover is open.
- [x] Add a property for setting the border colour
- [x] ~~Copy changes over to UWP and adjust accordingly.~~

Please note that I'm still learning WPF and may not have taken the best approach for solving the issue.

### DIFF
--- a/MainDemo.Wpf/ColorZones.xaml
+++ b/MainDemo.Wpf/ColorZones.xaml
@@ -52,6 +52,15 @@
                         <StackPanel Orientation="Horizontal"
                             materialDesign:RippleAssist.IsCentered="True">
                             <ToggleButton Style="{DynamicResource MaterialDesignHamburgerToggleButton}" />
+                            <ComboBox SelectedIndex="0" Margin="8 0 0 0" BorderThickness="0" 
+                                      materialDesign:ColorZoneAssist.Mode="Standard" 
+                                      materialDesign:TextFieldAssist.UnderlineBrush="{DynamicResource MaterialDesignPaper}"
+                                      BorderBrush="{DynamicResource MaterialDesignPaper}">
+                                <ComboBoxItem>Android</ComboBoxItem>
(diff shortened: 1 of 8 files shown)


NZSmartie:
Preview of what I've got so far

Keboo:
@NZSmartie I would not worry about doing anything for UWP. Right now it is not supported by this library.

NZSmartie:
Added a new attached property `TextFieldAssistant.UnderlineBrush` which ideally is assigned to the Underline control's background property.

NZSmartie:
there is a vertical bar visible when the drop down is open. Gonna address that next.

NZSmartie:
Any feedback @ButchersBoy? should I continue working out the visual glitches? or not worth doing this PR?

jespersh:
I noticed the vertical bar when I tested this on one of my monitors, but it didn't appear when I used the other.
Will have to check up if it was the primary or the secondary it was visible on, but it was enough to just drag it over for it to appear and disappear

Keboo:
Rebased and updated the code with the latest library changes. In the demo app with the inverted combo box you will notice the background is transparent because of the theme brush. Not currently sure how to handle this case.

Keboo:
Fixed the transparent popup background.

ahmedflix25:
@Keboo what about using ColorZoneAssist with buttons, textboxes and labels it will make it easier to change their colors?

Keboo:
@ahmedflix25 That is an interesting idea. I am not opposed to it, as it might make finding alternate color styles. I would probably want to rename it so it wasn't ColorZone specific. I will have to think about the relationship between such an attached property and the various styles that we have.

PR #948, opened 2018-05-05
Remove duplicate margin from cards

### LINKED ISSUE #947: Card has double Margin (Bug?)
Hello!

If I apply margin on a card, it also gets a padding with the same value.
Is this the case or a bug?

card - margin

Thank you for your help!

Keboo:
Fixes #947


ahmedflix25:
@mgnslndh it uses the Padding value for Margin which i think isn't a very good solution

Keboo:
@mgnslndh and @ahmedflix25 a bit of an explanation here: https://stackoverflow.com/questions/2006872/wpf-why-shouldnt-i-use-templatebinding-margin-in-controltemplate-is-margin

PR #949, opened 2018-05-05
Fix snackbar not being found

### LINKED ISSUE #946: Snackbar not working when hosting WPF within a Winform.
I am hosting the WPF user control containing the Snackbar using an ElementHost within a Winform. I have created the following example to provide a way to easily see the issue and the following fix that I used to solve the issue but this is very much a quick fix and not a good way of doing things.

[WinFormsHostingTestApplication](https://github.com/ChrisLawford/WinFormsHostingTestApplication.git)

To replicate the problem:
- Download the repo
- Add Reference Material Design (We will be changing code)
- Run the program
- Click "Add WPF Control"
- Click "This is a WPF control"
Desired outcome: The snackbar shows at this point with a message.

The Culprit:

Running the project with material design referenced you shall see that the Snackbar doesnt receive any messages when you click the button. 

```
private DispatcherOperation<Snackbar> FindSnackbar(Dispatcher dispatcher)
{
    return dispatcher.InvokeAsync(() =>
    {
        return _pairedSnackbars.FirstOrDefault(sb =>
        {
            if (!sb.IsLoaded || sb.Visibility != Visibility.Visible) return false;
            var window = Window.GetWindow(sb);
            return window != null && window.WindowState != WindowState.Minimized;
        });
    });
}
```
My quick fix is to change this to:
```
private DispatcherOperation<Snackbar> FindSnackbar(Dispatcher dispatcher)
{
    return dispatcher.InvokeAsync(() =>
    {
        return _pairedSnackbars.FirstOrDefault(sb =>
        {
            if (!sb.IsLoaded || sb.Visibility != Visibility.Visible) return false;
            //var window = Window.GetWindow(sb);
            //return window != null && window.WindowState != WindowState.Minimized;
            return true;
        });
    });
}
```

Keboo:
When parent window cannot be found the paired snackbar is still available. This can occur if the Snackbar is hosted outside of a WPF Window such as in a WinForms application.

Fixes #946

### DIFF
--- a/MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs
+++ b/MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs
@@ -328,7 +328,7 @@ namespace MaterialDesignThemes.Wpf
                 {
                     if (!sb.IsLoaded || sb.Visibility != Visibility.Visible) return false;
                     var window = Window.GetWindow(sb);
-                    return window != null && window.WindowState != WindowState.Minimized;
+                    return window?.WindowState != WindowState.Minimized;
                 });
             });
         }


jespersh:
Isn't `window?.WindowState != WindowState.Minimized;` the same as `window != null && window.WindowState != WindowState.Minimized`?

mgnslndh:
@jespersh this tripped me up as well, but...

```
window?.WindowState != WindowState.Minimized
```
if `window` is `null` then `window?.WindowState` will evaluate to `null` and the whole expression will evaluate to `true` because `null` is not equal to `WindowState.Minimized`

Not obvious at first but pretty nice!

Issue #885, opened 2018-01-21
Copy button missing from some places where ShowMeTheXAML is impemented


rc201612:
Hi,

Following the merge of #868, I have noticed that on the Expander, Menu & Toolbars and Dialogs views the "Copy" button is not present.

![2018-01-21_22-41-57](https://user-images.githubusercontent.com/18450631/35199885-b328e87e-fefc-11e7-8a26-d26ebc1722f7.png)

![2018-01-21_22-42-22](https://user-images.githubusercontent.com/18450631/35199886-b8b26022-fefc-11e7-8000-602d5b0d3d6b.png)

![2018-01-21_22-46-03](https://user-images.githubusercontent.com/18450631/35199896-e418e9e8-fefc-11e7-89b7-4429e0d77483.png)

PR #952, opened 2018-05-15
Updating ShowMeTheXaml to 1.0.10

### LINKED ISSUE #929: The "BuildXamlDictionaryTask" task failed unexpectedly.
Trying to run the demo app and i keep getting this error

The "BuildXamlDictionaryTask" task failed unexpectedly. 
System.TypeLoadException: Could not load type 'System.ValueTuple`2' from assembly 'mscorlib, Version=4.0.0.0		

materialdesignerror

Please i need help!

Keboo:
Fixes #929


ahmedflix25:
there is no scrollbar in the controller and i can't see the long codes like this one https://imgur.com/02qpbu0

Keboo:
@ahmedflix25 I believe that issue is the same as #885 and should be fixed by PR #887

ahmedflix25:
@Keboo that it, thanks

ahmedflix25:
it works as expected but there is that small thing when the length of the code is bigger than the popup width in a line, the popup width will increase -if there is enough space- and/or the horizontal scrollbar will show up -if there isn't enough space- 
continue scrolling to a normal length line and the popup will restore its width again and the horizontal scrollbar will collapse it's not a problem but it's annoying https://imgur.com/a/RsDEPPC

Keboo:
@ahmedflix25 agreed. That behavior is actually coming from the Avalon control that is being used to get the nice syntax highlighting. I suspect it is due to the virtualization that is used (making work well for large blocks of text). I have not looked into options for addressing that yet.

Issue #372, opened 2016-05-10
Add 24-hour format to TimePicker


arc95:
The TimePicker should have a way to override the Windows regional setting and show 24-hour time format in both the popup and also in the field. As it is, it only shows it in the popup, even with Is24Hours="True"

Keboo:
Fixed by #953

PR #953, opened 2018-05-16
Fix #941

### LINKED ISSUE #941: TimePicker doesn't actualize Gui after change of Is24Hours-Property
It's not possible to change the property Is24Hours of the TimePicker during runtime.
The AM/PM Buttons are correctly hiding when changing from 12 hour mode to 24, but the control isn't showing the second hour row (12 to 23 o'clock).

[https://github.com/speedyalibaba/MaterialDesignTest](url)

pavl0v:
This pull request to fix open issues #941, #372

### DIFF
--- a/MaterialDesignThemes.Wpf/TimePicker.cs
+++ b/MaterialDesignThemes.Wpf/TimePicker.cs
@@ -251,8 +251,9 @@ namespace MaterialDesignThemes.Wpf
 
             if (_lastValidTime != null)
             {
-                SetCurrentValue(SelectedTimeProperty, _lastValidTime.Value);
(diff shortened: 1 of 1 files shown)


Keboo:
Upon further testing it looks like there are still issues with this. If I toggle the Is24Hour properly back to false I properly get the AM/PM buttons but the clock still renders the extra hours (12-23).

pavl0v:
@Keboo I commited changes to regenerate clock buttons on Is24Hours changed. Seems like need to make some changes to modify entered text value on Is24Hours changed as well (in a case user does not make any changes in popup).

pavl0v:
@Keboo I made one commit more. Prevent create/destroy buttons. Create buttons once only and keep them in separate collections. Call HideButtons/ShowButtons on change hours/minutes/[seconds] mode.

Issue #961, opened 2018-06-07
can't use progressbar Indeterminate with MaterialDesignRaisedAccentButton or MaterialDesignRaisedButton


ahmedflix25:
using `materialDesign:ButtonProgressAssist.IsIndeterminate="True"` with any button with `MaterialDesignRaisedAccentButton ` or `MaterialDesignRaisedButton ` as its style has no effect at all and won't work

ahmedflix25:
https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/pull/964

PR #964, opened 2018-06-10
Indeterminate Progressbar on button #961

Streissi:
- changed the border (old progress indicator) to a real progress bar
- so a progress bar on a button can now be indeterminate

also updated the demo:
newbuttonprogressbar

### DIFF
--- a/MainDemo.Wpf/Buttons.xaml
+++ b/MainDemo.Wpf/Buttons.xaml
@@ -277,8 +277,21 @@
                                />
                 </Grid>
             </smtx:XamlDisplay>
-
-            <smtx:XamlDisplay Key="buttons_27">
+            
+            <smtx:XamlDisplay Key="buttons_27" Margin="5 0 0 0">
+                <Grid Width="124">
+                    <Button Style="{StaticResource MaterialDesignRaisedButton}"
+                            materialDesign:ButtonProgressAssist.Value="-1"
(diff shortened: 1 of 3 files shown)


jespersh:
Great stuff.
Think you could make the Opacity changeable?

ahmedflix25:
https://github.com/ButchersBoy/MaterialDesignInXamlToolkit/issues/961 thanks

ahmedflix25:
setting the materialDesign:ButtonProgressAssist.IsIndicatorVisible to false has no effect

Streissi:
I take a look in it later today

ahmedflix25:
also the Progressbar for the rounded buttons doesn't work anymore

Streissi:
the problem with the `FloatingActionButton` was, there was no opacity set by default. I set the default value to 1 for the `FloatingActionButton`

Streissi:
also the `IsIndicatorVisible` should work as expected now

ahmedflix25:
everything seems to work now, thanks

ahmedflix25:
another problem any button with accent color is washed out like 'MaterialDesignRaisedAccentButton' i think it's because of Opacity also

ahmedflix25:
another thing i noticed just now all buttons with accent color as their style like 'MaterialDesignRaisedAccentButton' is washed out now i think because of Opacity

PR #991, opened 2018-07-23
Added Documentation

ChinKX:
I'm currently using Visual Studio 2015 and I noticed that to run the project I have to import the necessary packages as well as to correct the programming syntax in the current code which is not supported by the C# version in VS 2015 (C# 6.0).

It would be helpful if we give a reminder about the programming syntax correction to new contributors using VS 2015.


ChinKX:
I have made another new commit. I corrected some grammatical errors inside CONTRIBUTING.md

ChinKX:
Welcome. I’m glad to be able to involve in such a great project ;)

PR #992, opened 2018-07-23
Fix #826 timepicker 24 hour bug

### LINKED ISSUE #826: Timepicker bug with 24h format
Hi, 

we found a problem with the timepicker in 24h format. 

We use two timepickers in our wpf control. If we change the value for one of them the wrong time format for the other is set. You can see this in your expample application, too as it's shown in this video:
https://giant.gfycat.com/UnsungOldArgusfish.webm

Is there a workaround for it?

yosefmah:
Not sure why this converter only returns Date property but it seems to resolve, or at least provide a workaround to, issue #826 

I had a look at [this binding](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/2740f14a814896d42032ae0013b765a8a0ec04c3/MaterialDesignThemes.Wpf/TimePicker.cs#L416) and [this line](https://github.com/ButchersBoy/MaterialDesignInXamlToolkit/blob/2740f14a814896d42032ae0013b765a8a0ec04c3/MaterialDesignThemes.Wpf/Clock.cs#L120) which gets hit.  The issue is it gets hit resulting in a change from PM -> AM time when removing the value in any unrelated Timepicker in the control which I couldn't figure out.  .

### DIFF
--- a/MaterialDesignThemes.Wpf/Clock.cs
+++ b/MaterialDesignThemes.Wpf/Clock.cs
@@ -114,11 +114,11 @@ namespace MaterialDesignThemes.Wpf
 		private static void IsPostMeridiemPropertyChangedCallback(DependencyObject dependencyObject, DependencyPropertyChangedEventArgs dependencyPropertyChangedEventArgs)
 		{
 			var clock = (Clock)dependencyObject;
-			if (clock.IsPostMeridiem && clock.Time.Hour < 12)
-				clock.Time = new DateTime(clock.Time.Year, clock.Time.Month, clock.Time.Day, clock.Time.Hour + 12, clock.Time.Minute, clock.Time.Second);
-			else if (!clock.IsPostMeridiem && clock.Time.Hour >= 12)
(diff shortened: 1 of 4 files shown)


yosefmah:
Made small change to improve the workaround for 24 hour clocks as the [IsPostMeridiemPropertyChangedCallback](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/2740f14a814896d42032ae0013b765a8a0ec04c3/MaterialDesignThemes.Wpf/Clock.cs#L114) keeps getting hit when changing any other TimePicker also in the control.

(Unfortunately the clock pop-ups don't appear in the video)

https://giant.gfycat.com/CommonSmoggyGuillemot.webm

Unfortunately TimePickers in 12-hour mode still get changed.  

https://giant.gfycat.com/InsidiousEagerCicada.webm

Essentially, when changing/deleting the text in a TimePicker, it causes the other TimePickers to have their SelectedTime properties changed - and the IsPostMeridian flag getting reversed is what causes the PM -> AM conversion.  

It only happens when the Picker being changed is AM.  I've looked and can't see any obvious reason for this.

yosefmah:
Closing as bug has been fixed.

PR #1045, opened 2018-09-21
DialogHost and VisualTree

Streissi:
This PR fixes two issues:

1. The fadeout background animation from the dialoghost #1022 
2. the visual tree in studio #1005 

dialoganimation
visualtree

PR #1046, opened 2018-09-23
Fixed DialogHost background fade out

### LINKED ISSUE #1022: 2.5.0: When dialog closes, darkened background is removed immediately
In the current demo project on the master branch, when closing a dialog in the Dialogs part of the demo, the darkened background is immediately made bright instead of a gradual fade. (It fades in correctly when opening a dialog, it's just when closing it that the missing fade occurs.)

fade

jizc:
Fixes #1022

**Cause**
The background brush was bound to the `DialogHost.IsOpen` property. This property is set to `false` before the close transition starts. Therefore when the closing transition starts, the background brush would already be set to `null`.

**Solution**
Moved setting of the background brush from the triggers to the visual states as part of the storyboard.


jizc:
Oops. Somehow missed that there already exists a PR that fixes this issue (#1045).

PR #1058, opened 2018-10-04
Typography update

spiegelp:
New styles for `TextBlock` and `Hyperlink` according to the new [specification](https://material.io/design/typography/the-type-system.html).

### DIFF
--- a/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Hyperlink.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Hyperlink.xaml
@@ -1,69 +1,8 @@
 ﻿<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
                     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
-    
-    <!-- initial Material Design specification -->
-    <Style TargetType="{x:Type Hyperlink}" x:Key="MaterialDesignBody1Hyperlink">
-        <Setter Property="FontSize" Value="13"/>
-        <Setter Property="FontWeight" Value="Regular"/>
-        <Setter Property="TextDecorations" Value="None" />
-        <Setter Property="Foreground" Value="{DynamicResource MaterialDesignBody}" />
-        <Style.Triggers>
-            <Trigger Property="IsEnabled" Value="False">
-                <Setter Property="Foreground" Value="{DynamicResource MaterialDesignCheckBoxDisabled}" />
-            </Trigger>
-            <Trigger Property="IsEnabled" Value="True">
-                <Setter Property="Cursor" Value="Hand" />
-            </Trigger>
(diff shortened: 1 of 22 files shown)


jespersh:
Is that the 2.0 spec or a revision to the 1.0? 
And since I'm already asking, how do you find changes to material design spec?

spiegelp:
@jespersh Google always made changes to the specification. However, Google introduced huge changes to Material Design and launched a new website earlier this year. As far as I know, they never called it officially Material Design 2 or used any revision numbers. Changes are only mentioned by month and year in a "What's new" section. 

Google relaunched the old website with the "old" specification last summer. You will find the old typography specification [here](https://material.io/archive/guidelines/style/typography.html).

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Hyperlink.xaml
```diff
@@ -59,4 +61,76 @@
...
+                <MultiTrigger.Conditions>
+                    <Condition Property="IsEnabled" Value="True" />
+                    <Condition Property="IsMouseOver" Value="True" />
+                </MultiTrigger.Conditions>
+                <Setter Property="TextDecorations" Value="Underline" />
+            </MultiTrigger>
+        </Style.Triggers>
+    </Style>
+    <Style TargetType="{x:Type Hyperlink}" x:Key="MaterialDesign2Body2Hyperlink"
+           BasedOn="{StaticResource MaterialDesignBody1Hyperlink}">
```
Keboo: It looks like several of these styles are still being based off of the old hyperlink style. Is that intentional?

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.TextBlock.xaml
```diff
@@ -77,4 +79,95 @@
...
+        <Setter Property="FontWeight" Value="Regular" />
+    </Style>
+    <Style TargetType="{x:Type TextBlock}" x:Key="MaterialDesign2Body2TextBlock">
+        <Style.Resources>
+            <Style TargetType="Hyperlink" BasedOn="{StaticResource MaterialDesign2Body2Hyperlink}" />
+        </Style.Resources>
+        <Setter Property="FontSize" Value="14" />
+        <Setter Property="FontWeight" Value="Regular" />
+    </Style>
+    <Style TargetType="{x:Type TextBlock}" x:Key="MaterialDesign2CaptionTextBlock">
```
Keboo: My personal opinion would be to drop the "2" from the key name on these. I do realize that there will be naming conflicts once that is done. However, some of these styles don't appear to be changing anything from the original style. This one for example appears to match the original `MaterialDesignCaptionTextBlock` unless I have missed something.

Keboo:
For this I think we are just going to replace the existing styles, so moving to the 3.0.0 milestone since that will be a breaking change.

PR #1060, opened 2018-10-05
Fixed TimePickerFormat issue

stefan-alexa:
The fix provided enforces the 24 hour format


Keboo:
@stefan-alexa are you able to address the merge conflicts? 
Also can you please add some description as to exactly what problem this is expected to fix? It looks like there may be multiple changes in this PR.

stefan-alexa:
@Keboo sorry for being so late with my changes. Indeed there were some more changes as I was using an older version of the code, but I've updated my fix with the latest version.
When you have the time please have a look again.

Keboo:
@stefan-alexa sorry for the long delay on this. When I got to looking at it closer it became apparent there was more than one issue with how the TimePicker was formatting and parsing time strings. I decided to fix as many as I could find in a separate PR (#1207). I believe I have also addressed the same problem you were solving here. Let me know if there is still an outstanding issue.

PR #1065, opened 2018-10-13
Add a Key to NumericUpDown Style

ahmedflix25:
add the key to be used with "MahApps:DataGridNumericUpDownColumn" "EditingElementStyle" to avoid the textbox Width problem like
`<controls:DataGridNumericUpDownColumn EditingElementStyle="{DynamicResource MaterialDesignNumericUpDown}"/>`

Before https://imgur.com/Xp0M6XG

After https://imgur.com/1QZGc8D

*note the MaterialDesignThemes.MahApps nuget 0.0.12 package is not compatible with MaterialDesignThemes 2.5.0 prerelease packages*


Keboo:
@ahmedflix25 is it possible to resolve this using the typed key for the style?
`<controls:DataGridNumericUpDownColumn EditingElementStyle="{StaticResource {x:Type mah:NumericUpDown}}"/>`

By applying an key to it, will no longer an [automatically apply](https://docs.microsoft.com/en-us/dotnet/framework/wpf/controls/styling-and-templating#relationship-of-the-targettype-property-and-the-xkey-attribute) to all `NumericUpDown` controls,

ahmedflix25:
@Keboo yes, it will not automatically apply to all NumericUpDown controls i was so focus on solving the DataGridNumericUpDownColumn problem with the same naming scheme for the project i forgot about NumericUpDown and it's Style.
sorry about that

Issue #1077, opened 2018-10-20
Change Datepicker months names 


AmRo045:
How can i change the months names in DatePicker control to local name?
I changed the CultureInfo and it worked in default WPF styles, but in Material Design it not work.

[Default Mode Screenshot](https://www.dropbox.com/s/6egjivxgemuzg3p/Normal.PNG?dl=0)

[Material Mode Screenshot](https://www.dropbox.com/s/s4ii7p1kwvwf1j7/MD.png?dl=0)

jespersh:
Wondering about these questions:
* Does it correct itself if you change month?
* When are you setting the application CultureInfo?
* How are you setting the CultureInfo?

spiegelp:
@AmRo045 Have a look at [App.xaml.cs](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MainDemo.Wpf/App.xaml.cs). The code in the comment shows how to set a CultureInfo. WPF apps usually run with the operation system's language. However you need to call `FrameworkElement.LanguageProperty.OverrideMetadata(typeof(FrameworkElement), new FrameworkPropertyMetadata(XmlLanguage.GetLanguage(CultureInfo.CurrentCulture.IetfLanguageTag)));` in order that WPF applies the language to string formats in XAML.

AmRo045:
> @AmRo045 Have a look at [App.xaml.cs](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MainDemo.Wpf/App.xaml.cs). The code in the comment shows how to set a CultureInfo. WPF apps usually run with the operation system's language. However you need to call `FrameworkElement.LanguageProperty.OverrideMetadata(typeof(FrameworkElement), new FrameworkPropertyMetadata(XmlLanguage.GetLanguage(CultureInfo.CurrentCulture.IetfLanguageTag)));` in order that WPF applies the language to string formats in XAML.

I used this code, but that thrown exception:

> Specified time is not supported in this calendar. It should be between 03/22/0622 00:00:00 (Gregorian date) and 12/31/9999 23:59:59 (Gregorian date), inclusive.

spiegelp:
[...]

PR #1080, opened 2018-10-21
Support for PersianCalendar

spiegelp:
basic date picker support for `PersianCalendar`

solves issue #1077


MaterialDesignThemes.Wpf/MaterialDateDisplay.cs
```diff
@@ -14,11 +14,19 @@ static MaterialDateDisplay()
...
+            UpdateComponents();
         }
 
         public static readonly DependencyProperty DisplayDateProperty = DependencyProperty.Register(
-            nameof(DisplayDate), typeof (DateTime), typeof (MaterialDateDisplay), new PropertyMetadata(default(DateTime), DisplayDatePropertyChangedCallback));
+            nameof(DisplayDate),
+            typeof (DateTime),
+            typeof (MaterialDateDisplay),
+            new PropertyMetadata(
+                System.Threading.Thread.CurrentThread.CurrentUICulture.Calendar != null ? System.Threading.Thread.CurrentThread.CurrentUICulture.Calendar.MinSupportedDateTime : default(DateTime),
```
Keboo: I don't think we want to have different default value based on culture. Rather I think we should take this culture into account at the places where the `DisplayDate` property is used.

Keboo: Yea I think that would be better. Just controlling it with a default value, also doesn't guard against someone setting a `DateTime` that is outside of the range. 
I think in these cases I am happy simply adjusting the invalid date to be in the valid range. I would also favor using the Language setting from the control over the culture specified on the thread. Something like:
```C#
MaterialDateDisplay materialDate = ...;
CultureInfo culture = materialDate.Language.GetSpecificCulture();
var minDateTime = culture.Calendar.MinSupportedDateTime;
var maxDateTime = culture.Calendar.MaxSupportedDateTime;

DateTime value = (DateTime) basevalue;
if (value < minDateTime)
{
    return minDateTime;
}

if (value > maxDateTime)
{
    return maxDateTime;
}

return value;
```

spiegelp: I implemented your suggestions.

Cyrusbm:
I think this problem has not been resolved completely. Months names is not shown in Persian calendar type.
Untitled
Untitled2

jamalhosseini:
This #1077 do not solved yet. the months name changed on header but not in calendar body. also days formation should be based on shamsi-date , but this is not.

@Cyrusbm have noticed this issue well.

Keboo:
@jamalhosseini can you log a new issue for this, so that it doesn't get lost,

jamalhosseini:
> @jamalhosseini can you log a new issue for this, so that it doesn't get lost,

you can see #1646

jamalhosseini:
> @jamalhosseini can you log a new issue for this, so that it doesn't get lost,

@Keboo please see #1646

Issue #1035, opened 2018-09-06
TimePicker control disabled hours/minutes


artoo80:
It would be really nice if the TimePicker control (and its Clock control) made the developer choose which hours/minutes are enabled and which are not. This would make the control more flexible to the various 
software needs.
For example in an appointment application it could be needed to choose just the timings every half an hour, so the only enabled minutes would be 00 and 30.
Thanks

dhilmathy:
Would like to pick this item. I got two approaches,

1. Two properties, `EnabledHours` & `EnabledMinutes` of type `string`; which can hold comma separated values. *(Easy for XAML use)*
2. Two properties, `EnabledHours` & `EnabledMinutes` of type `string[]`; holds enabled *Hours* and *Minutes* respectively. *(Easy for Binding)*

@Keboo Comments & Suggestions?

Keboo:
@dhilmathy that would be great. As for the design I am thinking that we probably want something that resembles the [DatePicker.BlackoutDates](https://docs.microsoft.com/en-us/dotnet/api/system.windows.controls.datepicker.blackoutdates?view=netframework-4.7.2). Also to stick with the pattern I would invert the naming of the properties. So rather than a list of enabled times (where the user specifies the hours/minutes that are enabled), do something more like blackout times (where the user specifies the hours/minutes that are disabled).

As for the types of the properties. I think it would probably be best to mimic the DatePicker and do a class (`TimePickerTimeRange`?). Again copying the DatePicker class since we want the timePicker to feel very similar. Not 100% sure on the exact structure of this (ideas welcome), but I am anticipating something that takes in a `DateTime` (perhaps some overloaded constructors to make it easier) to specify a range of hours/minutes and simply ignore the date portion of the `DateTime` (just like how the `DatePicker` ignores the time portion of the DateTime).

dhilmathy:
@Keboo Thanks for the suggestions. I can start as it's implemented in [DatePicker.BlackoutDates](https://docs.microsoft.com/en-us/dotnet/api/system.windows.controls.datepicker.blackoutdates?view=netframework-4.7.2).

One more thought,
For `TimePickerTimeRange` class, can we have [`TimeSpan`](https://docs.microsoft.com/en-us/dotnet/api/system.timespan?view=netframework-4.7.2) to represent `Start` and `End`, instead of using `DateTime` and ignoring the date portion of it?

Keboo:
[...]

PR #1103, opened 2018-11-02
TimePicker: added intervals

speedyalibaba:
Now its possible to define Intervals for the TimePicker.
Also added an example to the demo.

Hours: 
image

Minutes: 
image


Keboo:
Related to #1035

Keboo:
I think rather than allowing intervals to be done, we will simply allow blacking out hours/minutes as discussed [here](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/1035#issuecomment-423660787)

PR #1163, opened 2019-01-12
changed DialogHost to use a TaskCompletionSource

henon:
Changed DialogHost to use a TaskCompletionSource to await the session end instead of a ManualResetEvent.

DialogHost.ShowInternal used a ManualResetEvent to literally wait for the dialog to close. This wasted a threadpool thread which did nothing but wait blockingly for the user to close the dialog. Using a TaskCompletionSource this can easily be done without blocking any threads.

### DIFF
--- a/MaterialDesignThemes.Wpf/DialogHost.cs
+++ b/MaterialDesignThemes.Wpf/DialogHost.cs
@@ -58,7 +58,9 @@ namespace MaterialDesignThemes.Wpf
 
         private static readonly HashSet<DialogHost> LoadedInstances = new HashSet<DialogHost>();
 
-        private readonly ManualResetEvent _asyncShowWaitHandle = new ManualResetEvent(false);
+        // waiting for the Dialog to end is now handled by a TaskCompletionSource in DialogSession
+        //private readonly ManualResetEvent _asyncShowWaitHandle = new ManualResetEvent(false);
+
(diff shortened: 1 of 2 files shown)


henon:
OK, I see you like clean code ;).  Also I couldn't say with 100% certainty that DialogSession.Close was called in all cases. Thanks for reviewing. I am good with the changes.

henon:
Looking over the changes in detail, I actually think they are incorrect. I put the TCS into the DialogSession precisely because it needs to be created anew for every session in order to provide a virtual task for that dialog. Reusing one TCS for more than one dialog doesn't work. You would always await the first session end which would of course immediately return due to having already ended for any subsequent sessions.

Keboo:
@henon good catch. I have addressed it.

Keboo:
@henon yea there was still an issue if the DialogHost was opened with the static Show method and then closed using the IsOpen property. However this would have also caused a hang in the current release so I suspect very few users are trying to do this.

I have addressed it, as well as added in unit tests for some of the DialogHost functionality. This is an area of the repo that has been significantly lacking. Largely because most of the theme stuff is not really unit testable. However, custom controls like the DialogHost really should have some tests around them.

henon:
@Keboo: I agree about the tests. It is great to have such tests, not only for the obvious reason but also for newcomers like myself to be able to look at them and see how a class is supposed to work. Furthermore it helps a lot in the future, if further changes are to be made to that class.

PR #1179, opened 2019-01-23
WIP: Feature/enhanced palette better theming

### LINKED ISSUE #1057: Secondary colors.
The material design guidelines suggest secondary colors are like primary in that they have a light, mid, and dark tone. It would be awesome if the color palette was adapted to include these variations out of the box. I'm interested in hearing what others have to say on the matter.

Keboo:
Fixes #1057


rrs:
I have spent a bit of time on this over the past few days. The core of it is working but I am not happy with how it sits together. A big plus is at the moment there would be no breaking changes, which would allow a transition phase. The teething issues are around resources and what should have responsibility of them.

### Control Resource Order

The current goal was to use an extensible startup method to hook up all the core resources. The big issue here is order of static resource dependencies. For example the MahApps Styles need to be loaded before the MD Defaults. The colors and themes don't rely on order.

One way to solve this would be to remove the controls added by MD in the .MahApps extension and then re add them in the desired order.

### Managing resources by source

The limitation with this method is it locks down the resources to a set of known resource dictionaries and is therefore not very extensible. 

Instead of replacing whole dictionaries all dynamic resources would just use the ReplaceEntry method

### Palette Helper

Keeping this as an instance I think is important, so it is mockable. Having a static default instance serves the same purpose as the class being static. However it isn't going to be very friendly to use it on theming sub sections

### Theming sub sections

Using the same method as ThemeAssist we could apply different palettes to different sections. However linking up an instance of the palette helper would be cumbersome.

We could add a DP for the palette helper


## A Theme manager control

The above issues together make it difficult to work with. Another option to consider is a ThemeManager (or similarly named) control that could be used like

```
<md:ThemeManager Base="Light" Primary="DeepPurple" Secondary="Lime">
  <Grid />
</md:ThemeManager>
```

This would be a natural and easy to remember way to setup Theming. It also allow sub sections to be easily themed.

Internally ThemeManager would add the resources from the newly added classes and you would have the option of binding Base, Primary, and Secondary to your own versions if you want to tweak the specific colors. It would listen to routed events for theme changes.

Let me know what you think about this option

rrs:
Taking the theme manager approach further. An obvious flaw is that no ui element can exist at the top level above a window and therefore basic theming does not work on high level components (windows). This has clearly been thought about before and why @Keboo created a material design resource dictionary. I have explored this path because i believe it is the path of least resistance.

### MaterialDesignResourceDictionary

@Keboo's idea of a material design resource dictionary - which is really a neat idea - allows us to consolidate resource dependencies into one single point of entry which can also support visual studio auto complete suggestions (leveraging enums), for BaseTheme, PrimaryColor, and SecondaryColor. It also allows control over the resource dictionaries added or not added.

A major flaw in this approach is it is not very extensible. The previously detailed issue of resource dictionary merge order is not easily solved and using a custom resource dictionary, i have not found a better way than to completely override the resource dictionary class as you can see in the mah apps compatibility library.

### ThemeManager

This class is essentially an analog to the the resource dictionary which will set resources on a sub section of a visual tree. In an effort to stop the need for individual theme managers per extension library i have tried to make the theme manager delgate to the top most parent, which should always be the app level command handler on window. This means that the actual change action, be it theme, or palette etc can be tweaked in the custom resource dictionary


### Resource Dictionary V2

Really to be complete the resource dictionary needs properties which can be binded to for the theme and primary/secondary colors. They are currently an enum to cover the majority of cases but to allow for full customization in xaml it should be possible to bind in xaml to classes that describe the full range of static resources available. I plan to add a customtheme, customprimary, etc etc to the resource dictionary

Keboo:
@rrs I spent some time going through your comments and the code that is there. Here are some thoughts:

I am not sure I like the idea of the Theme manager control. Though it does make it nice for controlling color themes at an individual element level, I don't believe this is a very common use case. 

I think for this I am liking the resource dictionary approach. I do realize the issue with resource ordering but that is less of a concern for me since that will always be the case. This approach also handles the desire to change the color theme at the element level since a resource dictionary doesn't have to be attached at the App.xaml level if that is not desirable. 

So a few thoughts on that:
1. I would avoid explicitly setting up the resource dictionary in its own constructor (`Application.Current.Resources = this;`). This would prevent it from being used at the element level. Instead simply setting this as the resource dictionary in the App.xaml I think would be easiest. 
2. I would expect the `MahAppsResourceDictionary` to derive from the `MaterialDesignResourceDictionary`. To deal with the resource ordering I would suggest something like in `MaterialDesignResourceDictionary`
```C#
protected virtual IEnumerable<ResourceDictionary> GetResourceDictionaries()
{
    yield return CreateDefaultThemeDictionary();
    yield return CreateEmptyPaletteDictionary();
    yield return CreateEmptyPaletteDictionary();
}
```
And then something like this in the constructor:
```C#
public MaterialDesignResourceDictionary()
{
    foreach (var resourceDictionary in GetResourceDictionaries())
    {
        MergedDictionaries.Add(resourceDictionary);
    }
...
 }
```
This would allow the MahApps (or any other theme manager) to be able to derive and merge in its own dictionaries either before or after the MDIX ones.
3. I do also like the primary code way of interacting with these resource dictionaries to be either through the PaletteHelper/ThemeManager classes. Changing this to be an instance class is a much needed change. I think taking these a bit further and codifying the brushes such that it is easier for someone to be able to build up their own theme in code would be desirable as well.

Overall I like where this is headed. If you would like me to also jump in and PR into your branch I am happy to help out as well.

rrs:
@Keboo I hear you on 2. 1 also if resources are to be uses for subsections but more on that first..

Regarding the ThemeManager, if it was a resource dictionary it should really be a separate one, as you wouldn't want to add the control theme resources at sub levels, just the brushes. So in our examples you would end up with MdResourceDictionary, MdBrushResourceDictionary, MahAppsResourceDictionary, and MahAppsBrushDictionary. Putting the ThemeManager in the visual tree is really easy for users to use, it flows with the xaml. Making it just a proxy to the app level resource dictionary also means you only need the one theme manager rather than using a specific custom resource dictionary for any sub sections you might use. Using routed commands also allows for firing them in just plain xaml as well as code behind. How would you allow control of sub sections themes using something similar to the PaletteHelper? It doesn't know where the sub section resource dictionary is. Essentially the palettehelper, static or instance is serving a static function, because it uses the app level resourcedictionary.

For customizing the theme you can implement IBaseTheme and I can expose another bindable prop on the resource dictionary. This would override the enum based one. I was thinking of doing the same for the palettes although I wanted to think about alternative palettes or what if you have a primary and secondary theme but also have additional colors for each so you have primarylight, primarymid, primaryDark, alternative1, alternative2, and alternative3.

Keboo:
So after spending some more time with this. I don't think we need to include the control themes dictionary as part of these custom resource dictionaries. I am not sure including it as part of this adds much value. So if we keep it to just the theme and color brushes I think it gets a little easier. Ultimately I want to make setting of the theme colors easier, but at the same time I don't want to deviate too far from the built-in resource resolution in WPF. 

As for putting it in the visual tree, I think we can avoid it by simply allowing attached properties on the ThemeAssist. This would allow for changing the color theme at nested levels in the visual tree (though I suspect this case is not very common). 
In addition I would expect that we could add another attached property for attaching routed command handlers. This would then make changing the colors at a lower level easier too.

Thoughts?

rrs:
Sounds reasonable to me.  I don't mind you working on the branch by the way, i'll make sure I update here when i make any progress.

rrs:
Have a look over the latest commit, needs some review, may needs some consolidation but want some affirmations on the general direction

Keboo:
@rrs yes I like the direction this is heading. I had a little time today so I also did a little work in the branch and pushed some changes. Let me know if anything I did does not line up with what you were thinking.

rrs:
ok great, i don't see any commits other than mine on this branch though

rrs:
p.s. done the start of a hsl picker in the style of the google "color picker"

image

Keboo:
@rrs Sorry apparently there is a permissions issue with me pushing into your branch (perhaps because I was the one who opened the PR?). I did push them to [this branch](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/tree/feature/EnhancedPaletteBetterTheming)

rrs:
and now its in.. not complete yet though

image

@Keboo seems odd, can i do anything my end to sort the perms? I'll take a look anyway

rrs:
@Keboo there is code in md resgen for the swatch classes. Changes look fine except i don't agree with the restriction on MaterialDesignColors.MaterialDesignColor. The new color tool opens up the core palette by making each of the original palette variants 50, 100, 200 etc its own core palette (equivalent of the old 500 values.) When you click on any of the palette colors in the tool it takes one of those let's say Blue200 and generates a suitable light and dark  variant, therefore it necessitates the ability to specify a Blue200 value. The swatches facilitate that in the most part but they are not user friendly in xaml (no suggestions). That enum allows users to specify whatever colors they play with on the md palette in xaml, for example i might pick Blue600 for primary and Teal400. This should cover the 95% and the custom picker and the ability to take raw Colors covers the other 5%.

rrs:
In the new custom picker I'm really no happy with the way the values Hue, Saturation, Brightness, Color and Hsb all hang together. Depending on each other and needing to flag which prop is updating, nightmare. Anyway I know there is a better way but gone a bit blind so have a look and see if you can make it reasonable.

rrs:
@Keboo Last commit doesn't run when selecting the custom picker for me, but I see what you are doing and looks good thanks. Regarding the preview on the custom picker. I was purely emulating the google web picker but its not strictly necessary in the context of the color tool, as you have a preview in the buttons at the bottom. It makes sense if the control is used in its own right of course. So the slider and canvas picker should probably be their own control, and a preview can be added in a higher level control.
On the pickers in general, I've made the hex picker obsolete now so it could be removed completely, or it could be polished off so its usable in for example a popup, but I think the HSB picker can probably work in a tight space just as well so i'll let you decide on that. I would at least like one to be a full fledged control though so that we can add them to apps. 
Other ideas i have for the color tool are.

- Allow user to select the light and dark variants and select them manually.
- Allow users to select any of the color resources like MaterialDesignPaper and select a color. 

You would of course open the door to horrendous schemes but they could do that anyway. It would allow you to tweak backgrounds etc and see them in use in the demo app. Not a priority yet but just putting thoughts down

Keboo:
@rss Sorry about that, I was coding late at night. The error should be fixed now.

I think making the current HSB color picker should be made into a full fledged control. I like the idea of pulling the preview portion out of it. Likely just moving it into the demo app for now, I am wanting to limit how many controls get added into the main library. 
For this color picker I am thinking about the following: 
Rename to simply be ColorPicker. The fact that it is using the HSV color space internally seems like an implementation detail. Since the user just needs to get a `Color` out of it, either remove or hide the Hsb class/property from the public interface and just leave the `Color` property. Thoughts?


As for the demo app and picking the color theme; I like your ideas. I think ultimately what I would like the demo app to provide is a way for the user to play with different colors and then be able to view the code they need to apply to their app to get those same colors (maybe leverage something similar to the ShowMeTheXAML stuff that is used to provide the XAML for the styles). Obviously the exact form of this takes, and the UI depends on how the API for theming ends up looking.

Keboo:
@rrs be aware I did rebase this branch on top of master to pick up the work that was done for adding in unit tests. I am going to start adding tests for the ColorPicker since it will end up being a full control in the lib.

Tide:
I was literally going to address this today.  So glad to see that exactly what I wanted is already done. Can't wait to see it merged.  Super excited to have the color picker for other purposes as well.

Nice work and thanks.

Keboo:
Closing this one for now. The basic theme stuff is now merged in and I think the additional features forit should come in other PRs.

Issue #1191, opened 2019-02-15
Add RoutedCommand to PopupBox


Keboo:
Currently if the PopupBox is set with `StaysOpen` true it is non-trivial to simply have a button in the popup that closes the popup window. This should be improved by adding a new [routed command](https://docs.microsoft.com/en-us/dotnet/framework/wpf/advanced/how-to-create-a-routedcommand) to the PopupBox.

[This project](https://github.com/Keboo/MaterialDesignInXaml.Examples/tree/master/PopupBox/PopupBox.Simple) is a good example of the setup.

The following should be done.
1. Create a new routed command on PopupBox named `ClosePopupCommand`. You can use the existing `CloseDialogCommand` on the [DialogHost](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MaterialDesignThemes.Wpf/DialogHost.cs#L56) as an example.
2. In `PopupBox.OnApplyTemplate()` add a CommandBinding to `_popup` to handle the routed command. Again you can use the code in the [DialogHost as an example](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MaterialDesignThemes.Wpf/DialogHost.cs#L214). 
3. When the command is executing simply set `IsPopupOpen` to be `false`
4. Add the new command to the existing ["Save" and "Cancel" buttons on the demo app](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MainDemo.Wpf/Buttons.xaml#L157-L166).

Keboo:
Fixed by #1195

PR #1195, opened 2019-02-24
Add RoutedCommand to PopupBox - Issue:#1191

AlbertAlonso:
I think I have followed the requests, but I am happy to learn from my mistakes. I am not sure about point 2, as in the Dialog it does add to the CommandBindings directly and here it says to add it to the member _popup. 

1.  Create a new routed command on PopupBox named `ClosePopupCommand`. 
2.  In `PopupBox.OnApplyTemplate()` add a CommandBinding to `_popup` to handle the routed command.
3. When the command is executing simply set `IsPopupOpen` to be `false`
4. Add the new command to the existing ["Save" and "Cancel" buttons on the demo app]

### DIFF
--- a/MaterialDesignThemes.Wpf/PopupBox.cs
+++ b/MaterialDesignThemes.Wpf/PopupBox.cs
@@ -105,7 +105,7 @@ namespace MaterialDesignThemes.Wpf
         public const string PopupIsClosedStateName = "IsClosed";
 
         /// <summary>
-        /// Routed command to be used inside dialog content to close a dialog. Use a <see cref="Button.CommandParameter"/> to indicate the result of the parameter.
+        /// Routed command to be used inside of a popup content to close it.
         /// </summary>
         public static RoutedCommand ClosePopupCommand = new RoutedCommand();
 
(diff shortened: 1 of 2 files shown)


MaterialDesignThemes.Wpf/PopupBox.cs
```diff
@@ -394,7 +400,9 @@ public override void OnApplyTemplate()
             _popup = GetTemplateChild(PopupPartName) as PopupEx;
             _popupContentControl = GetTemplateChild(PopupContentControlPartName) as ContentControl;
             _toggleButton = GetTemplateChild(TogglePartName) as ToggleButton;
-            
+
+            _popup.CommandBindings.Add(new CommandBinding(ClosePopupCommand, ClosePopupHandler));
```
Keboo: I would add a null check here because it is possible that `_popup` will be `null`.
```suggestion
            _popup?.CommandBindings.Add(new CommandBinding(ClosePopupCommand, ClosePopupHandler));
```

MaterialDesignThemes.Wpf/PopupBox.cs
```diff
@@ -103,6 +103,12 @@ public class PopupBox : ContentControl
         public const string PopupContentControlPartName = "PART_PopupContentControl";
         public const string PopupIsOpenStateName = "IsOpen";
         public const string PopupIsClosedStateName = "IsClosed";
+
+        /// <summary>
+        /// Routed command to be used inside dialog content to close a dialog. Use a <see cref="Button.CommandParameter"/> to indicate the result of the parameter.
```
Keboo: This looks like the old comment from the DialogHost. I would update this to be "Routed command to be used inside of a popup box to close the popup box"

Keboo:
Fixes #1191

Issue #1097, opened 2018-10-29
DialogHost refresh issue on some computers (Surface laptop in this example)


bebenins:
It does not happen on my old ThinkPad but it does happen on 2 Surface laptops I've tried.

This is the example in video: https://photos.app.goo.gl/kJjbu4wraiRJFVWH7
And this is repo of a slimmed down app that you can download and try: https://github.com/bebenins/DialogRefreshIssue

I'm pretty sure this started happening with 2.5.0 release because I noticed this one some dialogs once I updated to 2.5.0:
![image](https://user-images.githubusercontent.com/16761301/47676924-3bd11100-db94-11e8-9ab0-00fc4a2de9d7.png)

jespersh:
Could you try software rendering mode as a test?
https://docs.microsoft.com/en-us/dotnet/api/system.windows.media.renderoptions.processrendermode?view=netframework-4.7.2

Adding in app.xaml.cs
```
protected override void OnStartup(StartupEventArgs e)
{
    RenderOptions.ProcessRenderMode = System.Windows.Interop.RenderMode.SoftwareOnly;
}
```

bebenins:
> Could you try software rendering mode as a test?
> https://docs.microsoft.com/en-us/dotnet/api/system.windows.media.renderoptions.processrendermode?view=netframework-4.7.2
> 
> Adding in app.xaml.cs
> 
> ```
> protected override void OnStartup(StartupEventArgs e)
> {
>     RenderOptions.ProcessRenderMode = System.Windows.Interop.RenderMode.SoftwareOnly;
> }
> ```

Yep, that fixed it.  So is this something to keep in mind or is there a way to "build" it into this toolkit and no one has to think about it again?
Thanks!

bebenins:
LOL, it works nicely for my example project but for some reason my actual application doesn't handle this code.  It builds fine but when I execute it, I don't see GUI on the screen...  I see the EXE in the tack manager but no GUI.  Any ideas?

jespersh:
Sadly the issue is your hardware and/or drivers.
Mozilla maintains a public list of blocked hardware accelerated gpus. More details here: https://wiki.mozilla.org/Blocklisting/Blocked_Graphics_Drivers
Disabling it for everyone just because it doesn't work on a few gpus isn't very reasonable

bebenins:
[...]

Issue #1198, opened 2019-02-25
DialogHost on MetroWindow causes rendering artifacts


Tokter:
As you move your mouse around, or tab between different controls. Horizontal and vertical lines appear:

![image](https://user-images.githubusercontent.com/3439101/53374099-38645100-390c-11e9-9507-297cc80d519c.png)

You can recreate the issue by adding a DialogHost to the MahAppsDragablzDemo MainWindow.xaml

```xaml
<controls:MetroWindow x:Class="MahMaterialDragablzMashUp.MainWindow"
                      xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
                      xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
                      xmlns:controls="http://metro.mahapps.com/winfx/xaml/controls"
                      xmlns:dragablz="clr-namespace:Dragablz;assembly=Dragablz"
                      xmlns:mahMaterialDragablzMashUp="clr-namespace:MahMaterialDragablzMashUp"
                      xmlns:dockablz="clr-namespace:Dragablz.Dockablz;assembly=Dragablz"
                      xmlns:wpf="http://materialdesigninxaml.net/winfx/xaml/themes"                      
                      WindowTransitionsEnabled="False"
                      TextElement.Foreground="{DynamicResource MaterialDesignBody}"
                      Background="{DynamicResource MaterialDesignPaper}"					  
                      GlowBrush="{DynamicResource AccentColorBrush}"					  
                      FontFamily="{StaticResource DefaultFont}"
                      Title="Material-MahApps-Dragablz Mash Up" Height="640" Width="800">
    
    <controls:MetroWindow.Flyouts>
        <controls:FlyoutsControl>
            <controls:Flyout x:Name="LeftFlyout" Position="Left" Header="Settings" wpf:FlyoutAssist.HeaderColorMode="Accent">
                <mahMaterialDragablzMashUp:FlyoutContent />
            </controls:Flyout>
        </controls:FlyoutsControl>
    </controls:MetroWindow.Flyouts>
    <controls:MetroWindow.LeftWindowCommands>
        <controls:WindowCommands>
            <wpf:PopupBox Style="{StaticResource WindowCommandsPopupBoxStyle}">
                <StackPanel>
                    <Button Content="Hello World"/>
                    <Button Content="Nice Popup"/>
                    <Button Content="Goodbye"/>
                </StackPanel>
[...]

PR #1199, opened 2019-02-26
Fix: #1097 & #1198 Adding a UseLayoutRounding=True to the DialogHost 

Tokter:
seems to fix the issue reported in #1097 & #1198

### DIFF
--- a/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.DialogHost.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.DialogHost.xaml
@@ -20,6 +20,7 @@
         <Setter Property="wpf:ShadowAssist.ShadowDepth" Value="Depth5" />
         <Setter Property="DialogTheme" Value="Light" />
         <Setter Property="PopupStyle" Value="{StaticResource MaterialDesignDialogHostPopup}" />
+        <Setter Property="UseLayoutRounding" Value="True" />
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="wpf:DialogHost">
--- a/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.DialogHost.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.DialogHost.xaml
@@ -11,6 +11,7 @@
         <Setter Property="AllowsTransparency" Value="True" />
         <Setter Property="PopupAnimation" Value="None" />
         <Setter Property="Placement" Value="Center" />
+        <Setter Property="UseLayoutRounding" Value="True" />
     </Style>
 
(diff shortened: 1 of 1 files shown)


Keboo:
Should this also get applied to the `MaterialDesignEmbeddedDialogHost` below?

jespersh:
https://docs.microsoft.com/en-us/dotnet/api/system.windows.frameworkelement.uselayoutrounding?view=netframework-4.7.2#remarks
> You should set UseLayoutRounding to true on the root element. The layout system adds child coordinates to the parent coordinates; therefore, if the parent coordinates are not on a pixel boundary, the child coordinates are also not on a pixel boundary. If UseLayoutRounding cannot be set at the root, set SnapsToDevicePixels on the child to obtain the effect that you want. 

Is DialogHost the root element? I forgot if it is.

Tokter:
@Keboo Yes, I added it as well.
@jespersh Not sure, but it creates a popup with its own Visual Tree, so I think so...

Keboo:
@jespersh that is an excellent point. _Usually_ the dialog host is a root level element inside of the window but it doesn't have to be (the demo app [shows this case](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MainDemo.Wpf/Dialogs.xaml)). 
By setting this on the DialogHost element directly it will propagate down to both the dialog itself as well as all of the control that are being wrapped by the DialogHost. Given that, and the suggestion from the docs about wanting to set this on the root element to handle the case where a nested element may not be on the pixel boundary, I think I would suggest the following (assuming I am understanding everything correctly):
1. Set UseLayoutRounding on [MaterialDesignDialogHostPopup](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.DialogHost.xaml#L9-L14). That way it will get applied to the Popup control (and I suspect resolve your issue as well). 
2. Avoid setting it on the DialogHost since there are no guarantee that it will exist on the pixel boundaries. If people know that it is what they want, based on how they are using the DialogHost in their app, they are free to derive from the MDIX style and add it.
3. We should be able to ignore the embedded dialog host (sorry for my previous comment) since it lives in the same visual tree.

Tokter:
@Keboo Your suggestion in point 1 works, UseLayoutRounding appear on the root element:
image

But the embedded dialog is then still glitchy:
image

image

image

Tokter:
I guess you could say that if you use the embedded dialog, then its up to the user to add a UseLayoutRounding on the MainWindow.

Keboo:
@Tokter yea I think simply documenting that behavior for the embedded dialog host is the way to go. I have been working on trying to update the wiki with lots of this knowledge.

Keboo:
Wiki documentation added [here](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/wiki/Dialogs#dialog-content-shows-border-lines)

PR #1202, opened 2019-03-04
Add Converters namespace to exported namespace

tadgerb:
There are a lot of useful converters in the Converters namespace. This change allows access to that location through the exported xmlns.


Keboo:
Thank you for the contribution.

PR #1211, opened 2019-03-20
Implement round corner textbox

mikefoxnobel:
Create new outline / filled style inherit from MaterialDesignTextBox, because existing outline / filled styles are inherit from MaterialDesignFloatingHintTextBox, which looks not good when applying large radius corner.


Keboo:
I don't believe this is still an issue in the 3.0.0 release. The corner radius appears fine. Thank you for the PR.

Issue #1205, opened 2019-03-16
Outlined passwordbox and textbox disabled text colors are different


mostwantedcheater:
![image](https://user-images.githubusercontent.com/20153173/54471260-397bf600-47c7-11e9-8ebf-28ed87b94c32.png)

PR #1222, opened 2019-04-04
Changed Opacity on Disabled TextBox #1205

AlbertAlonso:
Just a fix, not sure why there are 5 commits of my previous fork in this pull though.
The fix itself just matches the passwordbox and the TextBox opacity.


Keboo:
Fixes #1205

Keboo:

Issue #1123, opened 2018-11-21
ChecBox style incorrect for indeterminate (Tri-State) CheckBox


pfeigl:
Currently, an indeterminate Checkbox is displayed with the "Check" Icon but an applied Opacity of 0.56.

Based on https://material.io/design/components/selection-controls.html#checkboxes (scroll a little down to "States"), the indeterminate state should be displayed with full opacity but a "Minus" icon instead of the "Check" Icon.

pfeigl:
This is the corresponding part in the code:

https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.CheckBox.xaml#L232

pfeigl:
Here is the SVG already, gonna prepare a PR if needed
```
M6,13
L6,11
L18,11
L18,13
M19,3
H5
C3.89,3 3,3.89 3,5
V19
A2,2 0 0,0 5,21
H19
A2,2 0 0,0 21,19
V5
C21,3.89 20.1,3 19,3
Z
```

Keboo:
Hey @pfeigl I have to apologize for dropping the ball on this one. Yes I would be happy to accept a PR for this.

It looks like you are headed in the right direction. Feel free to directly tag me with any issues/questions you run into.

PR #1229, opened 2019-04-10
Changed visuals for checkboxes based on material design guideline (#1123)

pfeigl:
See https://material.io/design/components/selection-controls.html#checkboxes for details

In detail, the following changes have been applied:
* IsEnabled trigger has been moved after IsChecked trigger to make sure the correct MaterialDesignCheckBoxDisabled (gray) background is aplied to disabled checkboxes
* The SVG for the indeterminate checkbox has been changed to the correct display
* The background color for the indeterminate checkbox has been set to the checkboxes base color
* Opacity of disabled elements is now less opaq


Keboo:

PR #1239, opened 2019-04-18
Requesting Feedback: Improved theming API

Keboo:
This is a proposed change for a cleaner API. If you are able I would love to get feedback on this. 
This is heavily based on the work @rrs did on PR #1179. 

### Questions I would like answers too.
1. Does this meet the needs of your app? Are there any use-cases I have missed?
2. Would a RoutedCommand to change the theme be helpful? We could register class level handlers on the Window to make changing the theme a bit more automatic. This would open up the possibility of pure XAML being able to make theme changes. 
3. I am not really happy with the `BundledTheme` name. Any suggestions on what to call the markup extensions that build up the brush resources? 
4. The old naming scheme used `MaterialDesignPaper` and `MaterialDesignBody`. The new API mirrors these names, but it doesn't have to. Should these be renamed in the new API to something more like `Foreground` and `Background`?
5. Any other feedback?

### Highlights 
Because this is based on PR #1179 this includes the new `ColorPicker` control as well as a new Color Tool area in the demo app. The old Palette section is still in the demo app, so the comparison can show you both the old API and the new API working together. 
ColorTool

However the goal of this was the simplify the theming API and make it simpler to work with when you want to do custom themes or colors. To this end there are some new classes. Everything revolves around the `ITheme` interface. This defines all of the colors used by this library. So if you want a completely custom theme, you can simply instantiate a new `Theme` class and populate all of the values. There are extension methods on `ResourceDictionary` to `GetTheme`/`SetTheme`. The rest of the API is built on top of this.

#### Setting up your initial theme
You have several options here.
The old way that still works, in your App.xaml looks like this:
```XAML
<Application.Resources>
    <ResourceDictionary>
            <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Light.xaml" />
            <ResourceDictionary Source="pack://application:,,,/MaterialDesignColors;component/Themes/Recommended/Primary/MaterialDesignColor.DeepPurple.xaml" />
            <ResourceDictionary Source="pack://application:,,,/MaterialDesignColors;component/Themes/Recommended/Accent/MaterialDesignColor.Lime.xaml" />

            <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Defaults.xaml" />

```
With the new API this can be reduced using the new markup extension:
```XAML
<Application.Resources>
    <ResourceDictionary>
            <materialDesign:BundledTheme BaseTheme="Light" PrimaryColor="DeepPurple" AccentColor="Lime" />

            <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Defaults.xaml" />

````
Or if you would prefer to use your own colors rather than one of the built-in themes there is an alternate markup extension for that.
```XAML
<Application.Resources>
    <ResourceDictionary>
            <materialDesign:CustomColorTheme BaseTheme="Light" Primary="Aqua" Accent="DarkGreen" />

            <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Defaults.xaml" />
```
Alternatively if you would prefer to do things in code. You can do it like this:
```C#
protected override void OnStartup(StartupEventArgs e)
{
    //This is an alternate way to initialize MaterialDesignInXAML if you don't use the MaterialDesignResourceDictionary in App.xaml
    Color primaryColor = SwatchHelper.Lookup[MaterialDesignColor.DeepPurple];
    Color accentColor = SwatchHelper.Lookup[MaterialDesignColor.Lime];
    ITheme theme = Theme.Create(new MaterialDesignLightTheme(), primaryColor, accentColor);
    Resources.SetTheme(theme);

    base.OnStartup(e);
}
```


Tide:
I have been watching this and am very happy to see it getting closer.  I created a similar custom color / theme creator for a previous application made with MahApps.  Now using this library I wanted to create the same options.  Thankfully, you guys have done exactly what I wanted.  To be honest you did it in a better way than I did.  I really like the look of it.  Here are my thoughts:

1. Does this meet the needs of your app? Are there any use-cases I have missed?
      - Yes and very well.

2. Would a RoutedCommand to change the theme be helpful? We could register class level handlers on the Window to make changing the theme a bit more automatic. This would open up the possibility of pure XAML being able to make theme changes.
     - I have no opinion so far.

3. I am not really happy with the BundledTheme name. Any suggestions on what to call the markup extensions that build up the brush resources?
     - StandardTheme, CombinedTheme, BuiltInTheme

4. The old naming scheme used MaterialDesignPaper and MaterialDesignBody. The new API mirrors these names, but it doesn't have to. Should these be renamed in the new API to something more like Foreground and Background?
     - I always did find this odd.  However, I am sure there was an original reason for it. Foreground and Background would work for me.

5. Any other feedback?
     - Just looking forward to this and thank you to those who contributed to it.

rrs:
1.
Yes
2.
I think they would be useful (biased of course) but as you have previously
raised, they are a bit of an edge case usage. At least i think. I can't
imagine most users of the toolkit provide live theme changes to their end
users. However I don't think its a large burden to maintain it.
3.
I struggle with naming as it is but I would say the original cut of
MaterialDesignResourceDictionary is easier to remember. Its just a splice
of the 2 things that it is combining. MaterialDesignTheme... but that would
stomp on the current class that uses it.
4.
The original names are based on the original material design spec. The
names come from the elements, controls etc from that spec and how they are
emulated in this toolkit (at least thats what i've always thought
https://material.io/archive/guidelines/). The spec has moved on in some
ways, like the theming (The color tool is a clone of the google one after
all) but i haven't read it all in depth. Ideally I would say as the spec
evolves so should the toolkit, it would require cross referencing the
changes, or if there is a change history in the spec then applying changes
where appropriate.
5.
Good work on cleaning this up and I look forward to seeing it generally
available :)

Rob

Tide:
On #2 if that part makes live theme changes possible then I am all for it.  I always try to provide the greatest flexibility I can.  I have not instituted live theme changes in my previous applications but why not?  If it were an option I am sure some would find it useful (Windows does).

rrs:
5.
How come you left out the secondary light / dark variants?

Keboo:
@rrs the rational was that those are currently not used by the library and would only be there if someone wanted to consume them in their own app. If extra brushes like this are desirable, they can be easily added when the theme changes. This is effectively the same thing that the [MahApps integration has to do](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/theme_api/MaterialDesignThemes.MahApps/MahAppsBundledTheme.cs).

rrs:
@Keboo I added them to match the current MD spec. The way I was intending it, was it would be there to use for anyone from day 1 and we could implement some of the controls using the new variants, it would be quick and easy to add button variants for example. Most of the styles would just be copy and paste and change the resource name. It was on my TODO list but right at the end ;)

Keboo:
@rrs that makes sense. I am ok putting them in as part of this. If you want, we can merge this branch back into yours and keep working on this. There is still more cleanup and some renaming that I want to do with this branch before it gets merged.

Keboo:
Outstanding items to review before merge:
Ripple.RippleOnTop
ShadowAdornerAssist
ThemeManagerAssist
ButtonAssist.CorenrRadius - There is another in 2.6.0 that adds this. Consider merging it in along with this.

rrs:
@Keboo 
Ripple.RippleOnTop - Probably clear from the name. I had issues with the large clickable primary / secondary palette previews, this was to allow the ripple to go over the button content rather than under.
ShadowAdornerAssist - I had a further play with what I was trying with the shadow adorner a while ago. I'll move it into my fork and you can have a look.

Keboo:
@rrs sounds good. I was just reviewing this PR to try and get it merged for 2.6.0. I wanted to at least make sure I understood all of the changes that were going into the core lib before completing the merge. 

Not sure if you had a chance to look, but most of my review work on this PR I did on a [live stream](https://www.twitch.tv/videos/422803302) (PR stuff starts around the 15:00 mark).

rrs:
@Keboo ive ported the shadow adorner implementation I had in a test project over to my form on the theme_api branch. I'm not happy with it though. 

For clarity I'll define my intentions.

The material design spec details at length the use of Shadows to highlight pertinent sections. If you have a flat and tight card layout you should raise it above its siblings to draw attention to it. It should ideally animate too.

Applying the shadow assist to a controls in a grid without margins on the items leaves no space for the shadow. Z order dictates which part of a shadow will overlap another sibling. The last item in a list will show the shadow over items before it but will still get clipped by the parent control.

Ideally we would allow shadows to bleed over anything with a lower shadow depth. The whole UI would need to have the concept of depth. I don't know where to start or even if it's possible in wpf.

What I have tried is using the adorner layer. First approach was to use a transparent border in the adorner layer but the drop Shadow effect only renders when there is something tangible there. Either a border thickness or background colour. So I found a hack that draws a white rectangle with border and cuts the middle out to show underneath. This apparently is an expensive manoeuvre although it functions fine for what I'm doing. It would probably take it's toll if there where lots of these elevated borders. Even so in my test project it works full for my purpose but in the demo project the shadows render over their siblings correctly but still don't render on the outer border, which is the dialog host. It might be possible to have a shadow host that has 5 layers and you pass the shadow to that.

Another approach I tried was to lift the entire control up to the adorner layer. Swap in a filler control and place it in the adorner layer. The issue I had here was with the IsMouseOver property. It would set back to false when trying to move the control to the adorner layer. I expect it would do the same if there was a shadow host or layer host.

It may not be worth it for this PR but the principle I'm very keen on. Flat spaceless lists using shadows to highlight are popular and shadows are big part of the MD spec in general.

Maybe you have ideas on how it could be done or can improve on what I've done .

rrs:
@Keboo Here is the little test app i was playing with. It's much clearer for seeing what I really want to achieve. https://github.com/rrs/ShadowAssist

rrs:
@Keboo  Sorted it in the demo project now. Didn't realize the scollviewer had its own adornerdecorator. Which was causing it to appear clipped. Still not really happy with it for the reasons outlined above but it does what i intended it to do at least. Not sure I actually like the look of it, doesn't quite look like a naturally raised effect. The border on the buttons is messing with the effect i think. You'll have to have a look to see what i mean

rrs:
Made a start on expanding the styles to use the secondary variants. Just to get the ball rolling

image

rrs:
Almost forgot about this, so while its in my head...
## Freeze or not to Freeze
The original Theme dictionaries all use solid color brushes that are frozen. If you look at the code in the function SetSolidColorBrush of [ResourceDictionaryExtensions.cs](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/theme_api/MaterialDesignThemes.Wpf/ResourceDictionaryExtensions.cs) it essentially does this.

- Swap the incoming color for the given key
- If the existing value is **not** frozen then animate from the old value to the new value.

Given that it was never used for anything else other than swapping out the color brushes you can probably infer the history of the function. I suspect the brushes were not frozen originally, and that this was put in as you would expect, so that when changing the palette you get the cute animation, however it was later realised that there is a performance hit leaving them unfrozen. The frozen attribute was added for this reason, but the function was left in its original state, with the obsolete check on if the brush was frozen or not.

Anyway if the above is correct or not it is probably unwise to leave it out. It was on my mental list of TODOs towards the end. I was going to call the freeze method on the brushes before they are added to the resource dictionary.

I do like the animation though and last year when i started work on this i was having a play with that bit of code and i splintered it off into its own branch here [PaletteHelper.cs](https://github.com/rrs/MaterialDesignInXamlToolkit/blob/feature/ColorChangeAnimation/MaterialDesignThemes.Wpf/PaletteHelper.cs) it is in the method ReplaceEntry.

Issue #1075, opened 2018-10-19
Request: Outline button style


mgnslndh:
I would like a new style for a flat button with an outline. This style is described in the official guidelines.

Suggested style name `MaterialDesignOutlinedButton`

## Reference

https://material.io/design/components/buttons.html#outlined-button

mgnslndh:
Turns out this is all that needs to be added if we're OK with everything else being same as a flat button:

```
<Style x:Key="MaterialDesignOutlinedFlatButton" TargetType="{x:Type Button}" BasedOn="{StaticResource MaterialDesignFlatButton}">
     <Setter Property="BorderThickness" Value="1"/>
     <Setter Property="BorderBrush" Value="{DynamicResource MaterialDesignFlatButtonClick}"/>
</Style>
```

Keboo:
@mgnslndh That looks like a pretty good start. It does look like the spec calls for the ripple to follow the primary color.
![image](https://user-images.githubusercontent.com/952248/47229395-82489380-d37c-11e8-9592-02ea732e04da.png)

This may be as simple as adding (would need to test to see which):
```XAML
<Setter Property="wpf:RippleAssist.Feedback" Value="{DynamicResource PrimaryHueMidBrush}" />
<!-- or -->
<Setter Property="wpf:RippleAssist.Feedback" Value="{DynamicResource PrimaryHueLightBrush}" />
```

jamesport079:
Those outline buttons look slick. The corners of the ones in the spec are a tiny bit more rounded however. CornerRadius 3 or 4. 
![image](https://user-images.githubusercontent.com/36583787/47524540-278eca80-d89b-11e8-8fcb-a07fa033f941.png)

PR #1248, opened 2019-05-08
Added MaterialDesignOutlinedButton Style

AlbertAlonso:
Basically get the Flat Design and added a `BorderThickness` = 1, and changed the `RippleEffect` color. Also changed the background when it is pressed to not see the gray color, as in
[MaterialDesign Outlined Button](https://material.io/design/components/buttons.html#outlined-button)

Also added two buttons in the Demo Enabled and disabled, next to the flat ones. 

MaterialDesignOutlinedButton

Issue: [#1075](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/1075)


Keboo:
Thanks @AlbertAlonso. Great work

PR #1253, opened 2019-05-16
MainDemo-ButtonCleanup

Tide:
Working on cleaning up the button section of the demo.


Tide:
@Keboo I think we are good to go on this one.  The only pages that have left/right scrolling at the default window size are Dialogs and Snackbars.

I tried to get the margins the same on the buttons page.  I tried to not do too much since I think this page will be getting worked over again after the new theme system.  The main thing I worked on was organization and margins.

Tide:
I will be happy to combine what @rrs is dong [here](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/pull/1239#issuecomment-492854335) after #1239 is merged.

PR #1256, opened 2019-05-22
Adding Hand Cursor on SwitchToggleButton

mariedegen:
Adding Hand Cursor on SwitchToggleButton to make it visible when it's clickable


Keboo:
Thank for the PR, but I think leaving this out of the core library is probably best. If this is desirable, I suggest people add this to their own application styles. 

At least in Windows, the hand cursor is [typically used to denote links](https://docs.microsoft.com/en-us/windows/win32/uxguide/inter-mouse). 
> Text and graphics links use a hand or "link select" pointer (a hand with the index finger pointing screen shot of hand with index finger pointing ) because of their weak affordance. While links may have other visual clues to indicate that they are links (such as underlines and special placement), displaying the hand pointer on hover is the definitive indication of a link.

> To avoid confusion, it is imperative not to use the hand pointer for other purposes. For example, command buttons already have a strong affordance, so they don't need a hand pointer. The hand pointer must mean "this target is a link" and nothing else.

Issue #1265, opened 2019-06-02
Multi-lines textbox only clickable on the first line 


ahmedflix25:
add a new textbox with `AcceptsReturn="True"` and enough height to show more than one line

you can only click to set focus and start typing on the bottom line of the textbox by default the other lines aren't clickable or if you set the `VerticalContentAlignment` to `Top` the top line will be the only clickable one until you fill the rest with actual characters or white spaces.

PR #1297, opened 2019-06-25
Fix multi-lines textbox height

ahmedflix25:
Fix for https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/1265


jespersh:
I don't think paket-files/paket.processlock was supposed to be added?

PR #1302, opened 2019-07-01
1291/transitioner command enhancement

### LINKED ISSUE #1291: Transitioner MoveNext/Previous commands should use parameters
Right now if you use the routed commands on the transitioner control, it ignores any passed parameters and simply moves one slide in either direction.

We should do the following:

- [ ] In the `MoveNextHandler` and `MovePreviousHandler` methods you will need to handle the parameter on the `ExecutedRoutedEventArgs` if **it is a non-negative integer** then go ahead and use it in the current algorithm to set the `SelectedIndexProperty` otherwise simply user `1`
- [ ] Add unit tests to ensure the control behaves as expected. Something like this should help get you started writing the unit tests:
```C#
[StaFact]
public void WhenMoveNext_ItCanAdvanceMultipleSlides()
{
    //Arrange
    var child1 = new UserControl();
    var child2 = new UserControl();
    var child3 = new UserControl();

    var transitioner = new Transitioner();
    transitioner.Items.Add(child1);
    transitioner.Items.Add(child2);
    transitioner.Items.Add(child3);
    transitioner.SelectedIndex = 0;

    object parameter = 2;

    //Act
    Assert.True(Transitioner.MoveNextCommand.CanExecute(parameter, transitioner));
    Transitioner.MoveNextCommand.Execute(parameter, transitioner);

    //Assert
    Assert.Equal(3, transitioner.SelectedIndex);
}
```

dhilmathy:
Fixes #1291 
 
- [x] In the `MoveNextHandler` and `MovePreviousHandler` methods you will need to handle the parameter on the `ExecutedRoutedEventArgs` if it is a **non-negative integer** then go ahead and use it in the current algorithm to set the `SelectedIndexProperty` otherwise simply use `1`
 
- [x] Add unit tests to ensure the control behaves as expected.

### DIFF
--- a/MaterialDesignThemes.Wpf.Tests/MaterialDesignThemes.Wpf.Tests.csproj
+++ b/MaterialDesignThemes.Wpf.Tests/MaterialDesignThemes.Wpf.Tests.csproj
@@ -55,6 +55,7 @@
     <Compile Include="PopupBoxTests.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="TimePickerUnitTests.cs" />
+    <Compile Include="TransitionerTests.cs" />
     <Compile Include="VisualTreeHelper.cs" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


Keboo:

Issue #1262, opened 2019-05-27
Disabled listbox item accepts selection


danaki:
Thank you for such a great project!

Im trying to make toggled buttons like so:
```
<ListBox Style="{StaticResource MaterialDesignToolToggleListBox}" SelectedIndex="0" Margin="20,0,0,0">
    <ListBoxItem>Pause</ListBoxItem>
    <ListBoxItem IsEnabled="False">Unpause</ListBoxItem>
    <ListBoxItem>Finish</ListBoxItem>
</ListBox>
```

Notice that 2nd list item is actually disabled. It's grayed out but still accepts clicks and can be selected. I believe it's a bug.

Tide:
I tested your code and it looks like that is the expected behavior.  The selected item is disabled so in your code you can say if the item is disabled then do nothing.  I do not think you can stop the listbox from having a selected item without writing a custom ListBox or at least using a custom style to control the look in those situations.

Keboo:
Fixed by #1312

PR #1312, opened 2019-07-05
Fix ListBoxAssist.IsToggle toggling disabled items

ins0mniaque:
ListBoxAssist.IsToggle is used to emulate toggle buttons from list box items, and buttons should not respond to mouse clicks if disabled.

Used in MaterialDesignToolToggleListBox style.

This bug is also referenced in issue #1262.


bebenins:
Thanks!  I actually asked about this just few days ago in the chat.  This is a must have for a ToolBar

Keboo:
Thank you for fixing this!

PR #1327, opened 2019-07-19
Fix PackIcon

Keboo:
Changing the PackIcon to no longer derive from a generic class. This was causing a XamlParseException when used in element syntax (attribute syntax worked) on a menu item in the TextBox style, when the TextBox was created on a Window in Powershell.

I suspect this will also address several issues with the Xaml designer not being able to display a preview.

Added additional null check to TextFieldAssist


Awsmolak:
Not sure if this is the right place to comment on this. However, it seems like there is still a problem with PackIcon in the XAML designer.

After implementing pull request #1357 I merged in this branch to see if all of the designer issues were fixed. I am still having a problem loading MaterialDesignTheme.Defaults.xaml.

It appears that there is a problem in MaterialDesignTheme.ListBox.xaml lines 170 and 171 where Width and Height properties on PackIcon are inacessable or not recognized. This is kindof beyond my expertise so I'm not sure where to go from here.

I pushed a branch in my fork with both of the relevant commits merged here:
https://github.com/Awsmolak/MaterialDesignInXamlToolkit/tree/mergeFix1327andFix1357

PR #1346, opened 2019-08-02
Copied the ContextMenu of the TextBlock theme to the editable ComboBox theme

### LINKED ISSUE #1343: Context Menu in ComboBox is not themed
Reported by @ondrasvoboda on gitter.

Reproduction steps:
1. Create a combo box and set `IsEditable=True`
2. Right click on the editable text box in the combo box. 

Observed: The context menu is the default WPF context menu

Expected: It should be themed in the same way that the text box is.

You can also see this in the demo app:
image

MichelMichels:
Fixes #1343


Keboo:
Thank you for the contribution!

PR #1361, opened 2019-08-10
Adds a TextBox in DemoApp to edit the color

Mrxx99:
This pull request adds a TextBox to the "Custom" mode of the ColorTool which allows to edit the hex value of the selected color.
This also allows copy & pasting of hex values.


MainDemo.Wpf/ColorTool.xaml
```diff
@@ -407,7 +409,23 @@
                             <ColumnDefinition Width="160" />
                             <ColumnDefinition />
                         </Grid.ColumnDefinitions>
-                        <Rectangle Fill="{Binding Color, ElementName=ColorPicker, Converter={StaticResource ColorToBrushConverter}}" Margin="2,0,10,2" />
+                        <DockPanel>
+                            <TextBox Style="{StaticResource MaterialDesignFilledTextFieldTextBox}" materialDesign:HintAssist.Hint="Color HEX value"
+                                     DockPanel.Dock="Top" Margin="2,0,10,2">
+                                <TextBox.Text>
+                                    <Binding Path="SelectedColor"
```
Keboo: This is fine, however it causes the alpha channel to get dropped. This is a feature that will likely be added to the color picker before 3.0.0 releases. 
As a shortcut, if you just bind directly to the Color property on the Color picker it will use the type converter to map between the TextBox's Text property and the ColorPicker's Color property. This has the added benefit of supporting 3 digit colors, or even just typing in the name of the color (all of the same support you get in XAML). 
```
<TextBox.Text>
    <Binding Path="Color"
             ValidatesOnDataErrors="True"
             NotifyOnValidationError="True"
             ElementName="ColorPicker">
    </Binding>
</TextBox.Text>
```
Any reason to not do it that way?

Mrxx99: I had it similar to your proposal in my first commit (only binding to the ViewModels SelectedColor directly). I introduced the converter+validation because it displayed the alpha channel. I didn't know you want that to be added to the color picker as well, so I thought it does not make sense to only have it on the textbox and not the color picker. That's why I added the converter to get rid of the alpha channel.

Mrxx99: I changed to binding directly to color pickers color and removed converter and validation rule again.

Keboo:
Great work. Thank you.

PR #1362, opened 2019-08-10
added a ColorChanged event to ColorPicker (fixes #1356)

### LINKED ISSUE #1356: Add ColorChanged event to ColorPicker
In 2.6.0 we added a new [ColorPicker control](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MaterialDesignThemes.Wpf/ColorPicker.cs). There is a Color dependency property that can be bound, but not event. We should create a new routed event that is raised when the color property changes. This event should pass through the new color as it changes.

You can reference the routed event that is used inside of the [RatingBar as an example](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/febc81efebebd54f322518d60c8934d5161a6d1d/MaterialDesignThemes.Wpf/RatingBar.cs#L71-L82).

Mrxx99:
I added a ColorChanged routed event to the ColorPicker as described in #1356.

### DIFF
--- a/MaterialDesignThemes.Wpf.Tests/ColorPickerTests.cs
+++ b/MaterialDesignThemes.Wpf.Tests/ColorPickerTests.cs
@@ -63,6 +63,39 @@ namespace MaterialDesignThemes.Wpf.Tests
         }
 
         [StaFact]
+        public void SettingTheColorRaisesColorChangedEvent()
+        {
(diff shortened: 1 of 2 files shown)


Keboo:
Great job. Thank for the unit test as well.

Issue #1344, opened 2019-07-31
Set dialoge Host background to transparent 


RaoHammas:
hey ! can i set somehow back color to transparent ? i mean back color of content host always show white color even if i set transparent. Or can i change opacity of that black mask container of content ? I am loading user control as content of dialogHost . and trying to achieve something like Windows 10 Menu . Please help me. Thanks

![image](https://user-images.githubusercontent.com/16499626/61995223-d5cfd680-b09e-11e9-81c8-a1c126f19bc9.png)

_Originally posted by @RaoHammas in https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/1193#issuecomment-515685436_

Keboo:
What I think needs to be done for this:

- [ ] Create a dependency property on the `DialogHost` control  of type `Brush` with name `OverlayBackground`. This property should have a default value of Black.
- [ ] Remove the `BlackBackground` resource from MaterialDesignTheme.DialogHost.xaml and replace its usage with bindings to the new property on the `DialogHost`.

RaoHammas:
@Keboo  Hey thanks for response . So ill be seeing this feature in next update or what ?

Keboo:
@RaoHammas yes this was added to the latest 3.0.0 release. It should be there if you update to the latest 3.0.0 preview NuGet package.

RaoHammas:
@Keboo  Thanks a lot ! stay blessed :)

ckyRoomee:
how about the back color of content host? it always show white color even if i set transparent.Thanks in advance

ckyRoomee:
@RaoHammas please help me.Thanks

Issue #1193, opened 2019-02-17
Dialog Host Background Color


AmRo045:
How can i change Dialog host background color?

AlexejLiebenthal:
There is unfortunately no direct/easy way to do this.

You only can set the background via DialogTheme DP to `Light` or `Dark`.

If you like to apply a custom background color you'll need to overwrite the template.
A good quick start is to look how @Keboo did this on the default style/template [here](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/80aee475368e31c58bf8579a5397d9a9c316487d/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.DialogHost.xaml#L18).
You can than ignore most of the code there. Only focus on the `x:Name="PART_PopupContentElement"` Control and apply your changes their.

AlexejLiebenthal:
:man_facepalming:
There is actually a much easier way for background.

Put a `Panel` like a `Grid` to your `DialogContent[-Template]` and set the `Panel.Background` to your custom color.

PS:
As you originally asked for `DialogHost`. I assuming that you meant the dialog which is showed.
The host can't have any color, as it is a host for your normal app content and the dialog content. It also gives you the possibility to control a dialog (open/close etc.).

Keboo:
@AlexejLiebenthal is correct, and without going to the effort of re-tempting this is the best solution. It is worth noting that due to the rounded corners setting the background of the content of the dialog may result in a few pixels around the edge showing the theme (light/dark) color.
There is a new `ThemeAssist.Theme` attached properties that you can set on the DialogHost to change it from Dark/Light theme.

RaoHammas:
@Keboo hey ! can i set somehow back color to transparent ? i mean back color of content host always show white color even if i set transparent. Or can i change opacity of that black mask container of content ? Please help me. Thanks

RaoHammas:
![image](https://user-images.githubusercontent.com/16499626/61995223-d5cfd680-b09e-11e9-81c8-a1c126f19bc9.png)

Issue #736, opened 2017-07-02
How to change the black color of  DrawerHost 


ScriptSun:
Hi 

I am trying for few days change the DrawerHost black color I want it to be transparent no any color 

So i added this code by overriding the main DrawerHost code

Here is my code 

[Here is my code ](http://prntscr.com/fqt3q0)

I tried to make it whiote but it dont looks good for performance its like lighting and swiching its nott good 

so maybe you can help me fixing it 

Regards

ButchersBoy:
Hi, I can't access that file from where I am.  GitHub repos are better.

Are you looking for ```.LeftDrawerBackground```?

Doughouzlight:
Hi 
I am talking about the .RightDrawerBackground
Here is my code 
![alt tag](https://image.prntscr.com/image/Y2iibwQCSxa8bi0CpY8Vuw.png "Description goes here")

I want to remove the black background color from the slider 

Regards

ButchersBoy:
When you say "slider", you mean the semi-transparent bit which overlays the main content?

Doughouzlight:
Hi 
Thanks for fast reply 
i mean this << https://i.stack.imgur.com/8QkRn.png >>

This black on the page i want it to be normaly without any color

Regards

Doughouzlight:
Hi ButchersBoy
Do you know any solution?

Regards

Keboo:
@Doughouzlight That color is [hard coded in the control's template](https://github.com/ButchersBoy/MaterialDesignInXamlToolkit/blob/master/MaterialDesignThemes.Wpf/Themes/Generic.xaml#L690).

The simplest way to change it will be to [re-template the control](http://drwpf.com/blog/category/control-templates/) in your project. I would start by copying in the existing template and modifying it to suit your needs.

Doughouzlight:
Hi @Keboo 

Thank you for respond i really tried that all  i could not find how to make it **black** and change the speed of the DrawerHost during the opening and the closing of it.

i wrote this lines of code >>> http://prntscr.com/ft5ee8

I really will be thankful if you can write me here 3 lines of code to fix this problem

All i need make the black color to transparent and make it opening and closing fast like telegram app  

Regards 

Elin Doughouz

Keboo:
@Doughouzlight I think you are making it more difficult by trying to modify the existing template in C#. It is much easier to simply replace the template with one of your own in XAML.

[...]

Issue #568, opened 2016-12-07
Drawerhost/DialogHost overlay custom property for opacity and color


StealthRT:
I would like you to revisit the overlay for the flyout issue that you **do** add this as a property @ButchersBoy . I would like to be able to change the overlay color and also its opacity. I feel this would be a nice added feature and much needed depending on the users design habits.

If its also possible, it would be nice to also include the fade in/out time it takes if someone wants an instant background to appear or slowly fade in/out/etc.

varunvasisht84:
@ButchersBoy : The partial fix of this issue is done under https://github.com/ButchersBoy/MaterialDesignInXamlToolkit/pull/779/commits/d8aeb6373d97bfe3429b068af702ab257bf3bc44 . However I could not create the new pull request. It got merged under my previous open PR (#779). Just making sure it is not lost.

MichelMichels:
Is there any work left on this feature request?

Keboo:
Looks like it should have been resolved with #1364

PR #1364, opened 2019-08-10
Custom DialogHost background

Mrxx99:
This adds  a dependency property to the DialogHost called "OverlayBackground" which allows to set a custom brush for the background behind the dialog (as described by @Keebo in #1344).
I had to remove setting the background in the VisualStateGroup and move it to the normal Triggers, because Binding is not possible inside VisualState Storyboards for Colors ([like explained here](https://social.msdn.microsoft.com/Forums/vstudio/en-US/027c364f-5d75-424f-aafd-7fb76b10b676/templatebinding-on-storyboard?forum=wpf)).
This should not have any visual changes because the duration of the animation was set to 0.

I don't now if this counts as fix for #1193 or #1344 because it is not clear for me if this OverlayBackground is what is asked for in there or if they mean the background of the dialog itself.

grafik

------

> 
> [Keebo on #1344] What I think needs to be done for this:
> 
> - [x] Create a dependency property on the DialogHost control of type Brush with name OverlayBackground. This property should have a default value of Black.
> - [x] Remove the BlackBackground resource from MaterialDesignTheme.DialogHost.xaml and replace its usage with bindings to the new property on the DialogHost.

### DIFF
--- a/MainDemo.Wpf/Dialogs.xaml
+++ b/MainDemo.Wpf/Dialogs.xaml
@@ -147,12 +147,13 @@
                 <!--#region SAMPLE 4-->
                 <TextBlock TextWrapping="Wrap" HorizontalAlignment="Center" VerticalAlignment="Top"
                            Grid.Column="3" Grid.Row="0" Margin="8 0 8 0"
-                           >SAMPLE 4: Dialog managed from view model using IsOpen and custom commands (ignoring the provided routed commands).</TextBlock>
+                           >SAMPLE 4: Dialog managed from view model using IsOpen and custom commands (ignoring the provided routed commands). This also uses a custom brush to dim the background.</TextBlock>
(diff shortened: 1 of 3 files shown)


Mrxx99:
Also related to #568 and #736

Mrxx99:
I also looked into creating a dependency property for the opacity (#568) which uses anymation so I coud not simply move it to the normal Triggers. Because binding in VisualStates is not possible I had to set up the binding in code. You can take a look at it [here](https://github.com/Mrxx99/MaterialDesignInXamlToolkit/commit/9de7a34ac19a3ea0575a80763c58b0a0f0436691).

Setting up the binding in code behind is not nice, but it works. If you like this solution I can add this to this pull request or make a new one.

Keboo:
Great work! Thank you.

PR #1377, opened 2019-08-19
Fixed System.ArgumentNullException when double click the grayed grid to dismiss the dialog.

sandrohanea:
On the Close Dialog Command, a CanExecute method was on place to check if the `CurrentSession` isn't changed in the meantime, but on the  PreviewMouseUpwasn't checked this state.

Because of this, if the user double-clicked the grayed grid and the CurrentSession was assigned to null before the second click was intercepted by PreviewMouseUp, an ArgumentNullException occurs on the constructor of the DialogClosingEventArgs.


Keboo:
Thanks for the contribution!

PR #1398, opened 2019-09-04
Expose Popup HorizontalOffset and VerticalOffset properties

### LINKED ISSUE #1387: Additional PopupBox PlacementMode options
Currently, the PopupBox control has 12 placement modes:

        BottomAndAlignLeftEdges
        BottomAndAlignRightEdges,
        BottomAndAlignCentres,
        TopAndAlignLeftEdges,
        TopAndAlignRightEdges,
        TopAndAlignCentres,
        LeftAndAlignTopEdges,
        LeftAndAlignBottomEdges,
        LeftAndAlignMiddles,
        RightAndAlignTopEdges,
        RightAndAlignBottomEdges,
        RightAndAlignMiddles,

I was hoping to extend this to offer more options. Specifically, I'm looking for "Center" and a version of each existing placement mode where the PopupBox starts at a position that overlaps the popup button (rather than offset away from the button as they are now). This brings me to two questions.

1. First of all, is this functionality even desired? Personally it would be very useful as I think the current options are quite limited. However I am not sure how much customization is in the scope of this control so I ask to make sure.

2. If it is desired, I was wondering about the best way of implementing it. Doing it seems easy enough by extending the `switch (PlacementMode)` that PopupBox.cs currently uses. However, this would quickly make the enum count extremely large and the code as well. I was hoping someone had insight on a better direction to take (eg. maybe an enum for Center but a field for an overlapping option or something like that).

Thanks!

Anopob:
Closes #1387.

The bindings to the properties work as intended. However, some offset combinations seem to not work perfectly (for example, in my tests with all the bottom placement styles, I could not get the popup to open on the right of the button). This seems to be an issue with either the PopupEx itself or my tests in the demo, as can be tested by plugging in numbers directly in HorizontalOffset and VerticalOffset. Since the problem doesn't seem related to this code of binding the properties, I am leaving it as it is. Perhaps a new issue should be created regarding the offset problem (if it is a problem).

### DIFF
--- a/MaterialDesignThemes.Wpf/PopupBox.cs
+++ b/MaterialDesignThemes.Wpf/PopupBox.cs
@@ -316,6 +316,36 @@ namespace MaterialDesignThemes.Wpf
         }
 
         /// <summary>
+        /// Get or sets the popup horizontal offset in relation to the button.
+        /// </summary>
+        public static readonly DependencyProperty PopupHorizontalOffsetProperty = DependencyProperty.Register(
+            nameof(PopupHorizontalOffset), typeof(double), typeof(PopupBox), new PropertyMetadata(default(double)));
+
+        /// <summary>
+        /// Get or sets the popup horizontal offset in relation to the button.
+        /// </summary>
+        public double PopupHorizontalOffset
+        {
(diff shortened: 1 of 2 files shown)


Keboo:
Great work. Thank you for your contribution.

PR #1426, opened 2019-09-24
Update nuspecs to inclusive lower bound

### LINKED ISSUE #1415: Latest NuGet package throws warning NU1602
This issue can be observed by creating an empty project and adding the latest NuGet package (2.6.0 at the time of this writing). The warning can then be observed in the Error List window:
```
MaterialDesignThemes 2.6.0 does not provide an inclusive lower bound for dependency MaterialDesignColors (> 1.1.3 && < 2.0.0). An approximate best match of MaterialDesignColors 1.2.0 was resolved.
```

Based on this StackOverflow post (https://stackoverflow.com/questions/46254362/how-to-correct-dotnet-restore-warning-nu1604-does-not-contain-an-inclusive-lowe) it would seem that the way to fix this would be to modify the current relevant portion of the NuSpec:
```
    <dependencies>
      <dependency id="MaterialDesignColors" version="(1.1.3, 2.0)" />
    </dependencies>
``` 
to include a inclusive lower bound (using the interval notation specified in the docs https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges-and-wildcards) like so:
```
    <dependencies>
      <dependency id="MaterialDesignColors" version="[1.1.3, 2.0)" />
    </dependencies>
```

This would also be a good time to change it to include `2.0` if that was the original intention as well.

Overall, not an issue that causes any real problems, other than driving you crazy by constantly popping up the Error List window in VS whenever you build 😉

rgomez90:
Fixes #1415

### DIFF
--- a/MaterialDesignThemes.MahApps.nuspec
+++ b/MaterialDesignThemes.MahApps.nuspec
@@ -15,8 +15,8 @@
     <copyright>Copyright 2015 James Willock/Mulholland Software Ltd</copyright>
     <tags>WPF XAML WinRT Material Design Theme Colour Color UI UX</tags>
     <dependencies>
-      <dependency id="MaterialDesignColors" version="(1.1.3, 2.0)" />
-      <dependency id="MaterialDesignThemes" version="(2.5.1,3.0)" />
+      <dependency id="MaterialDesignColors" version="[1.2.0, 2.0)" />
+      <dependency id="MaterialDesignThemes" version="[2.5.1, 3.0)" />
       <dependency id="MahApps.Metro" version="1.6.*" />
     </dependencies>
   </metadata>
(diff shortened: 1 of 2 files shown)


MaterialDesignThemes.MahApps.nuspec
```diff
@@ -15,8 +15,8 @@
     <copyright>Copyright 2015 James Willock/Mulholland Software Ltd</copyright>
     <tags>WPF XAML WinRT Material Design Theme Colour Color UI UX</tags>
     <dependencies>
-      <dependency id="MaterialDesignColors" version="(1.1.3, 2.0)" />
-      <dependency id="MaterialDesignThemes" version="(2.5.1,3.0)" />
+      <dependency id="MaterialDesignColors" version="[1.2.0, 2.0)" />
+      <dependency id="MaterialDesignThemes" version="[2.5.1, 3.0)" />
```
Keboo: Can we change the lower bound of this to be `[2.6.0, 3.0)` please?

rgomez90: Done

Keboo:
Great work! Thank you for the contribution.

PR #1431, opened 2019-09-28
Removed and adjusted spaces from paket.lock

SanyuktaSaha:
Properly adjusted all the lines of code to make it more readable and manageable.


jespersh:
I don't think this should be merged.
In `paket.dependencies` have these changes been made and they're invalid:
* Multiple `~>` have been changed to invalid `->`. I am unfamiliar with paket, but I can't find they're using these anywhere.
In `paket.lock` have these changes been made to the paket **script generated**: 
* `netcoreapp3.0` has been changed to the invalid `netcoreapp 3.0`
* `netstandard1.3` changed to `netstandard 1.3`
* All the spacing has been changed which seems to confuse paket, with its indentation-dependent lock file.

I'd love to hear the reasoning behind these changes. 
Never modify `paket.lock` manually. Always modify `paket.dependencies` and use one of the commands described here: https://fsprojects.github.io/Paket/lock-file.html#Performing-updates
Hope this doesn't discourage you and please look into issues if you wish to contribute :-)

Keboo:
This appears to cause several breaks to the build. Going to close this for now as it appears there are several issues that need to be addressed.

PR #1446, opened 2019-10-05
Change Calendar TodayBackground to Transparent with a circle around it

ahmedflix25:
Calendar Today Background color and Selected Background color have very similar color and it's confusing for some now the Today Background Color has accent brush color


jizc:
I agree that the current style is not ideal. But I think a better solution would be to changed it to the current spec of an ellipse with a foreground brush stroke (1 thickness). You can see it in the first image here: https://material.io/components/pickers/

ahmedflix25:
image

jizc:
Using `Black` as the `Stroke` is going to be an issue when using the dark theme. I would suggest using `{TemplateBinding Foreground}` instead.

ahmedflix25:
the number inside the circle for current date will always be black and i can't figure out what makes it behave like that

jizc:
The black text on Today is caused by the `ObjectAnimationUsingKeyFrames` on line [218](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/aa293c74a30b134a0d48151879bf2fc3a39d7d40/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Calendar.xaml#L218). It is no longer needed, as the background of the Today indicator is no longer filled. Removing it solves the problem.

ahmedflix25:
i think we did it thanks for the help @jizc

jizc:
Happy to help! You might want to change the title of the PR to reflect your latest changes.

Keboo:
Very nice work. And thank you to @jizc for the code reviews.

PR #1455, opened 2019-10-07
Fix: setting HintAssist.HintOpacity on DatePicker wasn't possible...

adrianhara:
…now it is.


Keboo:

Issue #1430, opened 2019-09-25
Filter chips (chips that act like toggle buttons)


Taschenbuch:
Request for new Feature:

Filter Chips are Chips that can be clicked and react like a toggle button. They have a true and false state. Example: https://material.io/components/chips/#filter-chips

Requirement:
- they should be able to have different content shown when true or false. For example:
  - false: text
  - true: text and icon
- or
  - false: text + icon
  - true: other text + other icon
- the color for true and false state should be grey tones, but it would be great to have the option to change the colors for inactive and active state.
- there should be an non-outlined and and outlined version.

rgomez90:
I would like to help with this one, but I will need some advice for sure

Keboo:
@rgomez90 I am thinking something like this (there may be more but this is probably a good starting place that we can iterate on). From going through the spec, filter chips basically just maintain a "checked" state and display an alternate thumbnail icon when check. There are also several other chip behaviors that would be nice to implement after this, so I think we do something like this:

- [ ] Create a new boolean dependency property on the Chip class called IsSelected. Its default should be false. I suspect we will use this state for the other chip features as well, so I am a little hesitant to call it IsChecked.
- [ ] Create a routed event on Chip called "IsSelectedChanged". You can see an example of this with the BagedEx.BadgedChangedEvent. A unit test verifying that changing the IsSelected property raises this event would also be nice.
- [ ] Create two visual states for the Chip. You can take a look at the Flipper control for an example of a control with two visual states. When IsActive changes the visual state of the chip should change (`VisualStateManager.GoToState`). For the last parameter passed to `GotToState` we check the value of TransitionAssist.GetDisableTransistions on the chip itself. This will let people disable the animation if they don't want it.
- [ ] Create a new style for the Chip called MaterialDesignFilterChip. You can copy the existing Chip style as a starting point. 
- [ ] In the new style, it should default the Icon property (using a Setter) of the chip to an icon that looks like the check (from a quick glance the first one I see is `CheckBold`). This will let someone change out the icon if there is something else they would prefer, while providing a reasonable default.
- [ ] In the new style add the VisualStateManager with the two visual states created above (again take a look at Flipper for an example). To animate the the IconControl hiding I would put a ScaleTransform on the LayoutTransform property, and animate its ScaleX property (it has the range 0.0 to 1.0). This should achieve a nice slide in/out like in the specification.
- [ ] Add an example of this in the demo app.

[...]

PR #1456, opened 2019-10-07
Filter Chips

mgnslndh:
A stab at implementing a ListBox style for [filter chips](https://material.io/components/chips/#filter-chips).

FilterChips

This is work in progress. Only supports light mode at the moment. No fancy animations. Feedback or help appreciated.

There is an open feature request  #1430 for adapting the Chip control to support the filter chip spec. This PR tries to add support for filter chips in a ListBox.


Keboo:
Looking good so far.

Issue #1408, opened 2019-09-07
MaterialDesignSwitchToggleButton background colour is hardcoded


wwarby:
As best I can tell, the background colour on the "track" for the switch-style toggle button is hardcoded to `Black`. Since my app's background colour is close to black, this makes the default style of the switch toggle button unusable. I've tried to override the default style, so far without success - this is probably because I don't understand XAML style inheritance well enough to target the property correctly, but I wonder if perhaps it shouldn't be hardcoded to `Black` anyway, especially if the selected theme is `Dark`? What would be great is if it could just inherit the `Background` brush for the ToggleButton itself. This was my failed attempt at overriding:
```
  <Style x:Key="MaterialDesignSwitchToggleButton" TargetType="{x:Type ToggleButton}" BasedOn="{StaticResource MaterialDesignSwitchToggleButton}">
    <Style.Resources>
      <Style TargetType="{x:Type Rectangle}">
        <Setter Property="Fill" Value="Red" />
      </Style>
    </Style.Resources>
  </Style>
```

Keboo:
For this we should do the following:

- [ ] Add a new attached property in `ToggleButtonAssist` called SwitchTrackBackground of type `Brush`. It should default to `Black`
- [ ] Use this new attached property as the [Fill for the rectangle](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.ToggleButton.xaml#L335) on the `MaterialDesignSwitchToggleButton`
- [ ] Update the demo app so one of the switches shows off changing this color.

Keboo:
@laxmy yes all yours.

laxmy:
Hi @Keboo ,
I've raised a PR. Please review

PR #1460, opened 2019-10-09
#1408: Adding Track Background to SwitchToggleButton

laxmy:
This is to add Custom Track Background color to Switch Toggle Button. Added AttachedProperty in ToggleAssist class for the same and updated the demo app for its usage


Keboo:
Thanks for the contribution. I realized my initial direction were wrong and made some updates.

PR #1469, opened 2019-10-14
Update readme to reflect usage of VS2019

WillEllis:


Keboo:
Good catch. Thank you!

PR #1501, opened 2019-11-05
Filled and Outlined DatePicker styles

### LINKED ISSUE #1478: Filled Datepicker is missing
Filled Datepicker is missing. It could look like this

image

jizc:
Fixes #1478

FilledAndOutlinedDatePickerStyles

The new `DatePickerTextBox` style is a modified version of `MaterialDesignTextBoxBase`. The spacing in the styles are perhaps not exactly like the spec, but it's identical to the new Filled and Outlined TextBox styles in this library. I felt it was better to be consistent with those.

This PR replaces the template in `MaterialDesignDatePicker`, but I've attempted to make certain that the default style is not changed.


Keboo:
Great work! Thanks for the contribution

PR #1508, opened 2019-11-12
Force ComboBox popup direction

ondrasvoboda:
PopupPlacementMode attached property added to be able to force ComboBox popup opening direction.


MaterialDesignThemes.Wpf/ComboBoxPopup.cs
```diff
@@ -115,6 +121,22 @@ public ComboBoxPopupPlacement PopupPlacement
...
 
+        #region PopupPlacementMode property
+
+        public static readonly DependencyProperty PopupPlacementModeProperty
+            = DependencyProperty.Register(nameof(PopupPlacementMode),
+                typeof(PopupPlacementMode),
+                typeof(ComboBoxPopup),
+                new PropertyMetadata(PopupPlacementMode.Auto));
+
+        public PopupPlacementMode PopupPlacementMode
```
Keboo: Can we leverage the existing PopupPlacement property for this behavior rather than creating a new property/enum?

ondrasvoboda: I think it is possible, but it would require to replace internal changes to PopupPlacementProperty with some private variable to store current popup placement value. So the direction is not fixed once it is first time calculated.

Maybe one more improvement could be added - when e.g. UP direction is set but there is not enough space it should probably go down anyway. Then it wouldn't be forced direction but lets say preferred direction. What do you think?

I can prepare new PR for it, or can I somehow edit this one with new commit?

Keboo: Yea lets give that a shot. You can simply push changes to your `ComboBoxPopupDirection` branch and this PR will automatically update.

ondrasvoboda:
It turned out to be too complicated to use PopupPlacement property for it. I came up with simple solution using new PreferPopupUpDirection property. See my commit.

jespersh:
This seems a bit.. awkward? Maybe give it some time to come up with the right idea.
I got one, but I am unsure how to put the idea into code or text

Keboo:
Going to close this for now as it appears stale. Hopefully going to do a change of how the combo box works to address some of these common issues that keep cropping up.

Issue #1531, opened 2019-11-25
In GridViewColumnHeader PART_HeaderGripper is not displaying the correct cursor


Jesus805:
Hi. 

This problem can be replicated in the MaterialDesign Demo.
In line 91 of _MaterialDesignTheme.ListView.xaml_, there is a thumb `PART_HeaderGripper` that uses the style `MaterialDesignGridColumnHeaderGripper`. In _MaterialDesignTheme.Thumb.xaml_, we can see that `MaterialDesignGridColumnHeaderGripper` is supposed to have cursor SizeWE.
However, if you use Snoop to see the Thumb's cursor you can see that it's set to "None" [image](https://i.imgur.com/aNM87sf.png). This is also evident that the cursor looks different than SizeWE as seen here ![image2](https://i.imgur.com/7Stk6h1.png)
Accidentally named it SideWE in the picture, whoops!

Jesus805:
Been looking into this issue some more and here is what I can conclude.
- The Cursor needs to be set at the ControlTemplate level, setting PART_border to SizeWE will yield the correct cursor when hovering over the thumb. (I had to create a separate ControlTemplate for MaterialDesignGridRowHeaderGripper and MaterialDesignGridColumnHeaderGripper)
- Cursor TemplateBinding has not been working, PART_border will inherit from PART_HeaderGripper which has the cursor set to "None", even if I explicitly set a cursor..
- When dragging the thumb, the cursor will once again go to "None", I'm not sure how to resolve this, I've tried setting this trigger but that didn't help.
```
<ControlTemplate.Triggers>
    <Trigger Property="IsDragging" Value="True">
        <Setter Property="Cursor" Value="SizeWE"/>
    </Trigger>
</ControlTemplate.Triggers>
```
Is it possible to set Mouse.OverrideCursor in a xaml trigger?

Jesus805:
I came up with a solution to this problem but I was only able to implement it on my custom control.
https://github.com/Jesus805/SortableGridView

Jesus805:
Issue fixed with my latest pull request

PR #1542, opened 2019-12-02
Custom Thumb Control and ControlTemplate to fix issue #1531

Jesus805:
Hi,
I created a ControlTemplate (in ```MaterialDesignGridColumnHeaderGripper```) that sets the cursor to SizeWE. This is necessary because the Cursor needs to be explicitly set in ControlTemplate or else the cursor is set to "None" or in some cases disappears in the GridViewColumnHeader.
I also created a Thumb Control to override the cursor when dragging. This is necessary because when dragging, the cursor will revert to "None".

With these changes, a GridView be consistent with the DataGrid control.

### DIFF
--- a/MaterialDesignThemes.Wpf/GridViewColumnThumb.cs
+++ b/MaterialDesignThemes.Wpf/GridViewColumnThumb.cs
@@ -1,0 +1,22 @@
+﻿using System.Windows.Controls.Primitives;
+using System.Windows.Input;
+
+namespace MaterialDesignThemes.Wpf
+{
+    public class GridViewColumnThumb : Thumb
+    {
+        public GridViewColumnThumb() : base() { }
+
+        protected override void OnMouseLeftButtonDown(MouseButtonEventArgs e)
+        {
+            Mouse.OverrideCursor = Cursors.SizeWE;
+            base.OnMouseLeftButtonDown(e);
+        }
+
+        protected override void OnMouseLeftButtonUp(MouseButtonEventArgs e)
+        {
+            base.OnMouseLeftButtonUp(e);
+            Mouse.OverrideCursor = null;
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


MaterialDesignThemes.Wpf/GridViewColumnThumb.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System.Windows.Controls.Primitives;
+using System.Windows.Input;
+
+namespace MaterialDesignThemes.Wpf
+{
+    public class GridViewColumnThumb : Thumb
```
Keboo: For now lets make this an internal class. I am trying to avoid accidently exposing a class that we may change in the future.

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.ListView.xaml
```diff
@@ -88,7 +88,7 @@
             <Setter.Value>
                 <ControlTemplate TargetType="GridViewColumnHeader">
                     <DockPanel>
-                        <Thumb x:Name="PART_HeaderGripper"
+                        <wpf:GridViewColumnThumb x:Name="PART_HeaderGripper"
```
Keboo: Rather than changing the `MaterialDesignGridColumnHeaderGripper` template, lets simply do an explicit style on this element and base it on `MaterialDesignGridColumnHeaderGripper` and set the template here. This way we won't accidently change any other thumbs that may not be expecting this behavior.

Jesus805: I created a new style named `MaterialDesignGridViewColumnHeaderGripper`. The GridView Gripper does not use ```<Setter Property="Cursor" Value="SizeWE"/>``` so I opted to set the base as `MaterialDesignThumb`

MaterialDesignThemes.Wpf/GridViewColumnThumb.cs
Jesus805: Done

PR #1610, opened 2020-01-18
Update Flat and Outlined button border and ripple feedback brush

AmRo045:
Update Flat and Outlined button border brush and ripple feedback based on [material.io documentation](https://material.io/components/buttons/)

**Changes Preview:**
Changes Preview


Keboo:
Great work. Thank you!

Issue #1391, opened 2019-08-29
can't set GroupBox borders to a custom color


ahmediddma:
as the title said i can't set the groupbox brorders to a custom color and can only choose from the options available at the colorzone

PR #1611, opened 2020-01-18
Add Custom mode to ColorZone

ahmedflix25:
Now you can use `CustomBackground` and `CustomForeground` for the ColorZone if you choose `Custom` as its mode because sometimes you need your own colors like for a GroupBox or something 

fixes https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/1391


Keboo:
Though this works I suspect there is a slightly simpler solution.
Rather than adding a "Custom" member to the enum, I would suggest we simply add a "None" member. Then simply remove the style trigger, CustomBackground, and CustomForeground.
This then lets you use the Foreground and Background properties. 

Thoughts?

ahmediddma:
None for the Mode may give the wrong idea  i think we shuld keep Custom but removing CustomBackground and CustomForeground and using the already existing Background and Foreground  is a good idea

jespersh:
> Thoughts?

Thinking about this after looking into it a bit more. Wouldn't setting it to Standard (the first enum member) and setting Foreground and Background give the same result?

ahmedflix25:
> > Thoughts?
> 
> Thinking about this after looking into it a bit more. Wouldn't setting it to Standard (the first enum member) and setting Foreground and Background give the same result?

yes, i just checked

ahmediddma:
> 
> 
> > Thoughts?
> 
> Thinking about this after looking into it a bit more. Wouldn't setting it to Standard (the first enum member) and setting Foreground and Background give the same result?

yes. i just checked it

Keboo:
Upon further consideration I think we should actually keep this implementation. The linked issued #1391, it about wanting some mechanism of control for setting the colors of the GroupBox header. It is true you can hack around it by declaring alternate theme resources but this also would affect the content inside of the group  box which is problematic.

I would like to see this PR move a little further forward in implementing that link (likely through attached properties) with a sample in the demo app of changing the group box header color. But other than that I think this is fine.

ahmedflix25:
There are 1 custom ColorZone and 1 custom GroupBox in the demo now

You can set the custom colors from `ColorZoneAssist` class instead of the `ColorZone` class with `ColorZoneAssist.Foreground` and `ColorZoneAssist.Background` instead of `CustomForeground/Background` 

you can't run the demo and see the changes for now because of a runtime error 

`System.Windows.Markup.XamlParseException: 'Could not load file or assembly 'ICSharpCode.AvalonEdit, Version=6.0.178.0, Culture=neutral, PublicKeyToken=9cc39be672370310' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)'`

Keboo:
Great work. Thanks for the contribution!

PR #1633, opened 2020-02-01
Add None value to PackIconKind enum

### LINKED ISSUE #1627: PackIconKind should have a "None" value
Right now we automatically generate the PackIconKind enum using a [command line tool](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/mdresgen/IconThing.cs).
The enum that it generates does not contain a default "None" member. 
We should do the following:

- [ ]  Adjust the script so that the first member is always called "None"
- [ ]  Ensure the PackIcon control shows nothing and properly handles this new "None"

schulz3000:
- [x] Adjust the script so that the first member is always called "None" (new PackIconKind enum + PackIconDataFactory generated)
- [x]  Ensure the PackIcon control shows nothing and properly handles this new "None" (Tested it with Demo. Looks like no further changes needed to display "nothing" for None)

Fix #1627


HClausing:
And what about to add a new DependencyProperty to pass a custom Geometry (custom icon) when Kind is set to none?

PR #1636, opened 2020-02-03
Change hint container from TextBlock to ContentPresenter

### LINKED ISSUE #1635: FloatingHintTextBox wrong sample in Demo
For the Hint in FloatingHintTextBox demo a Stackpanel is defined. This Stackpanel is not displayed but "System.Window.Control" as string.
As I understand only simple string values can be diplayed as Hint there.

image

AmRo045:
Fix #1635


jespersh:
Just to be sure, can you check if you can tab to the contentpresenter? Might be needed to set IsTabStop to false.

AmRo045:
I checked it and doesn't need to `IsTabStop` property. Thank you for the recommendation.

Keboo:
Great work. Thank yo.

PR #1644, opened 2020-02-06
Fixes #1642 Change Line to stretched Path

### LINKED ISSUE #1642: Disabled TextBox is not being resized correctly
Disabled TextBoxes are not being resized properly.

Specifically, I am seeing this when a TextBox is wrapped in a ScrollViewer. In that case, when the TextBox is disabled, the TextBox's width can be increased (eg, when the top-level window is resized), but its width can then not be decreased (when the top-level window is restored to its previous size).

I've created a minimal project that demonstrates this issue:

https://github.com/philpowers/DisabledTextBoxResize

To reproduce:

1. Verify that the TextBox resizes correctly when the "Enable TextBox" option is checked by resizing the top-level window.

2. Uncheck the "Enable TextBox" option

3. Resize the top-level window. The TextBox will expand when the window width is increased, but it will not shrink when the window width is decreased.

jespersh:
This allows having a TextBox with IsEnabled set to false inside a scrollviewer and resize the window (smaller)

### DIFF
--- a/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.DatePicker.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.DatePicker.xaml
@@ -146,10 +146,16 @@
                                         HorizontalAlignment="Stretch"
                                         VerticalAlignment="Bottom"
                                         SnapsToDevicePixels="True" />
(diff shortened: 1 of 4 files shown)


jespersh:
#1644

jespersh:
Fixes #1642

jespersh:
`TimePicker`, `PasswordBox`, `DatePicker` also had the bug as described in the issue

Issue #1526, opened 2019-11-21
Wrong alignment with PopupBox.ToggleContent


alekstrust:
Hi
I have a PopupBox with a Border control inside the ToggleContent. .I want the the Border be stretched horizontally but it does not work.

Tried:
```
<materialDesign:PopupBox PlacementMode="TopAndAlignCentres" HorizontalAlignment="Stretch" HorizontalContentAlignment="Stretch">
    <materialDesign:PopupBox.ToggleContent>
        <Border>Content</Border>
    </materialDesign:PopupBox.ToggleContent>
    
    <ListBox/>
</materialDesign:PopupBox>
```
But it's always centered.
![image](https://user-images.githubusercontent.com/4238486/69379286-33a3b200-0c7e-11ea-9ca0-77da99fc2aca.png)

PR #1647, opened 2020-02-10
Fix horizontal and vertical content aligment in PopupBox (#1526)

FooPenguin:


Keboo:
Great work. Thank you.

PR #1668, opened 2020-02-22
Improve duplicate handling

w-syss:
I was very confused when using the IgnoreDuplicate Property of the SnackBar as it did the exact opposite of what i expected it to do. I think the wording "IgnoreDuplicate" implies that duplicates are in fact ignored and not that the IgnoreDuplicate property is ignored.

These changes aim to clarify what these properties actually do, along with a bit of simplification in the comparison logic.

I am aware that breaking the public API might not be the best idea, however I think this clarification is a good QoL change.


jespersh:
I suppose I can agree with that. Nice work.

Could I get you tempted to write a few tests to see that duplicates are properly discarded depending on different content and actioncommands?
It would be very appreciated :-)

MaterialDesignThemes.Wpf/SnackbarMessageQueueItem.cs
```diff
@@ -47,8 +48,37 @@ internal class SnackbarMessageQueueItem
         public bool IsPromoted { get; }
 
         /// <summary>
-        /// Still display this message even if it is a duplicate.
+        /// Always show this message, even if it's a duplicate
         /// </summary>
-        public bool IgnoreDuplicate { get; }
+        public bool ShowAlways { get; }
```
Keboo: I am wondering about changing the order of this to be "AlwaysShow". IMO it reads a little nicer (like the comment above it)

MaterialDesignThemes.Wpf/SnackbarMessageQueueItem.cs
```diff
@@ -47,8 +48,37 @@ internal class SnackbarMessageQueueItem
...
+
+        /// <summary>
+        /// Last time this message was shown
+        /// </summary>
+        public DateTime LastShownAt { get; set; }
+
+        public bool MessageExpired()
+        {
+            if (LastShownAt == null)
```
Keboo: `LastShownAt` is a value type, and will never be `null`. Perhaps make `LastShownAt` nullable

MaterialDesignThemes.Wpf/SnackbarMessageQueueItem.cs
```diff
@@ -47,8 +48,37 @@ internal class SnackbarMessageQueueItem
...
+            else
+                return LastShownAt <= DateTime.Now.Subtract(Duration);
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (!(obj is SnackbarMessageQueueItem message))
+                return false;
+
+            return Content.Equals(message.Content) && ActionContent.Equals(message.ActionContent);
```
Keboo: `Content` or `ActionContent` could be null here. I suggest using `EqualityComparer<object>.Default` to do the equality check rather `.Equals`

MaterialDesignThemes.Wpf/SnackbarMessageQueueItem.cs
```diff
@@ -47,8 +48,37 @@ internal class SnackbarMessageQueueItem
...
+        }
+
+        public override int GetHashCode()
+        {
+            var hashCode = -724337062;
+            hashCode = hashCode * -1521134295 + EqualityComparer<object>.Default.GetHashCode(Content);
+            hashCode = hashCode * -1521134295 + EqualityComparer<object>.Default.GetHashCode(ActionContent);
+            return hashCode;
```
Keboo: Though this implementation works, with the new C# value tuple there is a simpler method.
```
return (Content, ActionContent).GetHashCode();
```

MaterialDesignThemes.Wpf/SnackbarMessageQueueItem.cs
w-syss: Agreed.

MaterialDesignThemes.Wpf/SnackbarMessageQueueItem.cs
w-syss: Correct, i missed that. I think either replacing the null check with "LastShownAt == default(DateTime)" or just removing it would work. The default value of DateTime would produce correct results anyway (excluding extreme edge cases), your thoughts?

MaterialDesignThemes.Wpf/SnackbarMessageQueueItem.cs
w-syss: Agreed. I thought about using null-conditionals, but this seems the better choice in terms of readability.

MaterialDesignThemes.Wpf/SnackbarMessageQueueItem.cs
w-syss: This requires System.ValueTuple, doesn't it? I remember that being weird with versions of the .NET Framework <= 4.7. In my case I had to download the Nuget-Package for System.ValueTuple.

w-syss:
Thank you for you advice. I will add unit tests.

MaterialDesignThemes.Wpf/SnackbarMessageQueueItem.cs
Keboo: Yea I am ok simply removing it.

MaterialDesignThemes.Wpf/SnackbarMessageQueueItem.cs
Keboo: Yes you are correct. In this case I think I am ok taking a dependency on it for the case of `net45`.

w-syss:
I added `System.ValueTuple 4.3.1` as a dependency. For some reason for versions `>= 4.4.0` there is a dependency issue with `System.Core`.

Keboo:
@w-syss I apologize it has taken so long to get to the 4.0.0 release. Do you mind revisiting this PR and seeing about updating it so it can be merged?

w-syss:
Sure, just give me a couple of days.

Keboo:
This is resolved by #2240

Issue #1667, opened 2020-02-22
WPF Access key should not be used in list view items


schorsch2489:
### MaterialDesign automatical remove character underscore at label and listview

I writting a little programm for myself and use a database for it.
If i save a entry and it's already exists, i will give them automatically a new name,
Using the currentname and adding "_1" to it.
then i show all items at a ListView.
Therefore i add ListviewItems during the runtime to my ListView.
And here is the bug: i read the correct name from the database (checked it many times), 
but it displays not the correct name. 
Read the name "water_1" correctly from the Database, but what it shows is "water1".
Tryed also to adding a label instead of a listviewitem.
Doesn't solve the issue.
Strange is, that only these underscores (is this the name of these character: _ ?) is removed
After trying many different things, i used made a very easy testproject (only nessesary code for reading from database and showing in a listview).
And now it shows my entries correctly.
Trying to manage, what is at my project other and remove all stylings.
Now it show my entries correctly.
Adding one style after the other and compiling and testing, it was easy to say,
that the MaterialDesign and it's styles are the cause of the problem.
Added some pictures to show the behavior
### Repro steps

Please provide the steps required to reproduce the problem

1. Add MaterialDesign 3.0.1 to your project and use the themes

2. use one of the themes (tryed it with many different)

3. get data from behind and fill a ListView with ListviewItems, where some of the content has a _ in the namen - first of the _ will be removed.

### Expected behavior

Show the name correctly

### Actual behavior

Remove the first of the _ 
For example:
water_1 is shown as water1 in the listview
water_1_1 is shown as water1_1 in the listview

### Known workarounds
don't knowing any workarounds 

**pictures for the issue:**

_using materialdesign themes:_
![1 show listview with materialdesign](https://user-images.githubusercontent.com/61337326/75084330-c8d57180-551f-11ea-9e84-71b074b5b303.png)

_that is my background datatable:_
![2  background  data table](https://user-images.githubusercontent.com/61337326/75084331-c96e0800-551f-11ea-8e66-3ee4387151f7.png)

_without using materialdesign themes:_
![3  correct view without materialdesign](https://user-images.githubusercontent.com/61337326/75084333-ca069e80-551f-11ea-81d3-0194cedfeca3.png)

jespersh:
Hi, and welcome!
Please see this for an explanation: https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/1427

[...]

PR #1674, opened 2020-02-29
Disable AccessKeys for ListBoxItems.

w-syss:
References #1667.


w-syss:
**Before:**

Before

**After:**

After

Issue #1021, opened 2018-08-21
Inactive window and the dialog


LiptonOlolo:
The first click on the buttons in the inactive window, where the dialog is open, does not work, the first click will make the window active.

![gif](https://media.giphy.com/media/jxcmXQ09LMyKQ6wVxD/giphy.gif)

jespersh:
Doesn't every application do this? I tested with Firefox and it behaves the same way. No accidental clicks.

LiptonOlolo:
Ah, okay

greuelpirat:
Since my issue #1281 was marked duplicate, I want to comment here, that this is an issue of this material design libary.

As I described in #1281 this error happens only on dialogs with TextBoxes. Check the WPF-Demo app. The first dialog with fruits have this issue. the second dialog with only a Button does not. It may have something to do with the getfocus-animation of the TextBox.

jespersh:
Appears any binding or command is blocked on the first click when hosted inside a `DialogHost.DialogContent`

```xaml
<Window x:Class="WpfApp13.MainWindow"
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:materialDesign="http://materialdesigninxaml.net/winfx/xaml/themes"
        Title="MainWindow" Height="450" Width="800">
    <materialDesign:DialogHost>
        <materialDesign:DialogHost.DialogContent>
            <StackPanel>
                <CheckBox>No command</CheckBox>
                <CheckBox IsChecked="{Binding CheckedBinding}">Has IsChecked binding</CheckBox>
                <CheckBox Command="materialDesign:DialogHost.CloseDialogCommand">Closes dialog</CheckBox>
                <Button Command="materialDesign:DialogHost.CloseDialogCommand" Content="Close"/>
            </StackPanel>
        </materialDesign:DialogHost.DialogContent>
        <Button Command="{x:Static materialDesign:DialogHost.OpenDialogCommand}" Content="Open"/>
    </materialDesign:DialogHost>
</Window>
```

```C#
namespace WpfApp13
{
    public partial class MainWindow : Window
    {
        public bool CheckedBinding { get; set; }
        public MainWindow()
        {
            InitializeComponent();
            DataContext = this;
        }
    }
}
```

Issue #1281, opened 2019-06-13
Focus problem on dialog with textbox


greuelpirat:
If you open a dialog with an input field and go to another application, to proceed with the app you need an additional click to reactivate the dialog. (Version MaterialDesignThemes 2.5.1)

you can reproduce this with the wpf-demo:
1. In the WPF-Demo-Applikation go to "Dialogs"
2. In the first sample click "Add Fruit"
3. Focus an other applikation
4. Click cancel in the "Add Fruit"-Dialog

Expected: Dialog closed
Actual: Textbox got Focus-Animation. Dialog still open. Second click on cancel closes the Dialog.

This focus issue does not appear on dialogs with no input fields. (like Sample 2 in the Demo)

jespersh:
I seem to recall this has been up before in an issue. Anyone?

cpyfferoen:
#1021 Is the same.

greuelpirat:
Any chance for a fix?

greuelpirat:
I continue this issue on #1021

Issue #232, opened 2016-01-26
Dialog/Drawer Host does not receive focus.


amoamare:
If a DialogHost or a DrawerHost is shown and the user moves to another application and it loses focus, when they return to the application it does not get focus anymore and you can not type how ever the buttons do work except the app still does not get focus. You must click on the top bar to regain focus on the application to be able to interface with it.

ButchersBoy:
Fixed for dialog.  Couldn't repro' for drawer.  But there are some little drawer improvements in the pipeline.

PR #1710, opened 2020-03-25
Fix lost mouse clicks on inactive window (#1021)

greuelpirat:
The click on the dialog is lost when `window.Activate()` is called. The clicks works normal when the call is removed or the window gets the focus with `window.Focus()`

### DIFF
--- a/MaterialDesignThemes.Wpf/DialogHost.cs
+++ b/MaterialDesignThemes.Wpf/DialogHost.cs
@@ -632,7 +632,10 @@ namespace MaterialDesignThemes.Wpf
         {
             var window = Window.GetWindow(this);
             if (window != null && !window.IsActive)
+            {
                 window.Activate();
+                window.Focus();
+            }
             base.OnPreviewMouseDown(e);
         }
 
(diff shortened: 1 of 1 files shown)


greuelpirat:
It seems that `window.Focus()` does not work. I think the better solution would be to give up the `window.Activate()` completely.

I removed the whole `OnPreviewMouseDown` since all it does is to activate the window.

Are there known issue for `window.Activate()` to be required?

greuelpirat:
I made some further tests without `window.Activate()`:
click on Button in inactive SAMPLE 1 Dialog works
click on Button in inactive SAMPLE 3 Dialog works
click on TextField in inactive SAMPLE 3 Dialog does not works.

Do not merge this pull request. I need to make changes again.

Keboo:
@greuelpirat no problem I added "WIP" (work in progress to the title), simply remove it when you are done.

greuelpirat:
I looked deeper into it and found that the focus logic implemented with #232 was the problem. It prevented that the mouse event could proceed its logic. It was also a problem that when there are multiple `DialogHost` in a single window (like in the demo app) they could steal the focus each other.

I made some tests and it showed that `window.Activate()` is fully sufficient to handle the focus issues  #232 and #1281.

Keboo:
Great research and work. Thank you for the contribution.

greuelpirat:
@Keboo Thank you!

This was my first pull request and I missed to link it to the issue #1021. So the issue is still open. Would you close it? Thank you.

Keboo:
@greuelpirat I missed that links as well. Closed, thank you.

PR #1729, opened 2020-04-08
added VisualStudioToolsManifest to show controls in Toolbox

Mrxx99:
This adds the control to the ToolBox Window (and to Blend) as described here:
https://docs.microsoft.com/en-us/nuget/guides/create-ui-controls

I tried to select those controls that are meant to be used by the user of the library but that can easily be adjusted.

grafik


Keboo:
This is great thank you!

PR #1750, opened 2020-04-20
Set Dialog-Task-Result after IsOpen to prevent "DialogHost is already open."

greuelpirat:
We experienced a threading problem on slow devices when we chained dialogs:

Imagine the following flow:

```
var result1 = await DialogHost.Show(new Button
{
	Content = "Close it",
	Command = DialogHost.CloseDialogCommand,
	CommandParameter = "result1"
}, "RootDialog");

await DialogHost.Show(new Button
{
	Content = $"previous result: {result1}",
	Command = DialogHost.CloseDialogCommand,
}, "RootDialog");
```

When the first dialog is closed then (sometimes) an `InvalidOperationException` with `DialogHost is already open.` is thrown.

The reason for that lies in the order of the two instructions DialogHost#598

```
_dialogTaskCompletionSource?.TrySetResult(parameter);
SetCurrentValue(IsOpenProperty, false);
```

The task gets the result before IsOpen is set to false. And depending on the threads if the await continues before IsOpen updated the show()-call for the second dialog fails with the exception above.

I did not experienced this race condition on my PC and to be honest I do not really understand this, But on other slower devices this exception is easy to reproduce. We did some workarounds using dispatcher, but they were never a good solution.

So my proposal here is to set the result for the `_dialogTaskCompletionSource` after IsOpen is set to false. To pass the parameter I use an internal property in `DialogSession`

### DIFF
--- a/MaterialDesignThemes.Wpf.Tests/DialogHostTests.cs
+++ b/MaterialDesignThemes.Wpf.Tests/DialogHostTests.cs
@@ -64,6 +64,22 @@ namespace MaterialDesignThemes.Wpf.Tests
         }
 
         [StaFact]
+        public async Task CanCloseDialogWithRoutedEvent()
+        {
+            Guid closeParameter = Guid.NewGuid();
+            Task<object> showTask = _dialogHost.ShowDialog("Content");
+            DialogSession session = _dialogHost.CurrentSession;
+            Assert.False(session.IsEnded);
+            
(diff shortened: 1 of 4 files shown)


Keboo:
Thank you. I will dig into this more later. I suspect your change is likely fine, but there is a complex callback on the IsOpen property so I want to give it a closer look.

Keboo:
Great work. I added in a few additional unit tests just to make sure everything still works.

PR #1755, opened 2020-04-21
Added Snackbar.ActionButtonPlacement to support own line for action button

### LINKED ISSUE #1423: Snackbar should allow placing the action button on its own line.
From the [specification](https://material.io/components/snackbars/#specs)

> If an action is long, it can be displayed on a third line.

[Example](https://storage.googleapis.com/spec-host/mio-staging%2Fmio-components%2F1568406561463%2Fassets%2F11e9M-CnUBfAefLLAMtckb_opQih-E3Gm%2Funderstanding-snackbars-text-2-lines-longer-action.png)

To do this we should do the following (this assumes #1422 has already been completed):
[ ] - Add a new enum dependency property to the Snackbar control. The enum should have the values `Inline`, `SeparateLine`, and `Auto`. The default of the property should be `Auto`.
[ ] - When the property is set to `Inline` the action button should display to the right of the message (as it works now).
[ ] - When the property is set to `SeparateLine` the action button should be moved below the message and right aligned.
[ ] - When the property is set to `Auto` we should put in a trigger that only moves the action button onto its own line when there are two or more lines of text. We can detect this case by simply watching the height of the Content.
[ ] - Ensure that an example of each is present in the demo application.

greuelpirat:
Resolves #1423 

I used a `DockPanel` to switch between `Dock.Right` and `Dock.Bottom` for layout and a `IMultiValueConverter` to determine the value based on snackbar height and the newly snackbar property `ActionButtonPlacement`

(I also fixed minor warnings and updated some lines to use newer syntax)


MainDemo.Wpf/Snackbars.xaml
```diff
@@ -66,34 +66,30 @@
             <!--
                     * using a message queue to manage queuing of messages, and getting onto correct thread
                     * notice the shorthand syntax {materialDesign:MessageQueue} for creating a new message 
-                    queue without expanded XAML. useful in code-behind scenarios if you dont want to bind 
+                    queue without expanded XAML. useful in code-behind scenarios if you don not want to bind
```
Keboo, line 69: Minor typo here.
```suggestion
                    queue without expanded XAML. useful in code-behind scenarios if you do not want to bind 
```

MaterialDesignThemes.Wpf/Snackbar.cs
```diff
@@ -1,12 +1,40 @@
...
 namespace MaterialDesignThemes.Wpf
 {
+    public enum SnackbarActionButtonPlacementMode
+    {
+        Auto,
+        Inline,
+        SeparateLine
+    }
+
+    public class SnackbarActionButtonPlacementModeConverter : IMultiValueConverter
```
Keboo, line 19: Let's make this converter internal and moved into the Converters namespace/folder.

MaterialDesignThemes.Wpf/Snackbar.cs
```diff
@@ -1,12 +1,40 @@
...
+    {
+        public object Convert(object[] values, Type targetType, object parameter, CultureInfo culture)
+        {
+            var mode = (SnackbarActionButtonPlacementMode) values[0];
+            var snackbarHeight = (double) values[1];
+            return mode switch
+            {
+                SnackbarActionButtonPlacementMode.Auto when snackbarHeight > 53 => Dock.Bottom,
```
Keboo, line 27: Why the value 53? Is the the point where text wraps onto a second line?

MaterialDesignThemes.Wpf/Snackbar.cs
```diff
@@ -1,12 +1,40 @@
...
+                SnackbarActionButtonPlacementMode.Auto => Dock.Right,
+                SnackbarActionButtonPlacementMode.Inline => Dock.Right,
+                SnackbarActionButtonPlacementMode.SeparateLine => Dock.Bottom,
+                _ => Dock.Right
+            };
+        }
+
+        public object[] ConvertBack(object value, Type[] targetTypes, object parameter, CultureInfo culture) => new[] { Binding.DoNothing, Binding.DoNothing };
```
Keboo, line 35: In general MultiValueConverters should be only used with mode OneWay, so I am ok with this simply throwing an InvalidOperationException.

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Snackbar.xaml
```diff
@@ -11,43 +11,50 @@
...
                         <Trigger Property="IsEnabled" Value="false">
-                            <Setter Property="Opacity" Value="0.23"/>
+                            <Setter Property="Opacity" Value="0.23" />
                         </Trigger>
                     </ControlTemplate.Triggers>
                 </ControlTemplate>
             </Setter.Value>
         </Setter>
+        <Style.Triggers>
+            <Trigger Property="DockPanel.Dock" Value="Bottom">
```
Keboo, line 53: Is it possible to move this on an element inside of the ControlTemplate like `Root`? I suspect most people are not using this style outside of the snackbar and even less likely they have also set a Margin/Padding directly on the element, but it is possible. If we do it inside of the template, then we don't have to worry about overwriting values set directly on the Button.

w-syss:
Is it possible to split up your commit a little?
I feel like 5 changed files in one commit with as many insertions and deletions is too much for a single commit.
Even splitting up the cleanup and the newly added functionality would be beneficial i would think.
What do you think @Keboo?

MaterialDesignThemes.Wpf/Snackbar.cs
greuelpirat: sure, I will do that

MaterialDesignThemes.Wpf/Snackbar.cs
greuelpirat: Yes, thats looks light some magic. It is the point, where the placement of the button is decided. For me I oriented me at the hint in the issue:

> We can detect this case by simply watching the height of the Content.

So I measured the height of a single lined snackbar message and it something below 53. And I simply took this as a threshold. But I do understand, that this looks somewhat arbitrary. Do you have an idea?

MaterialDesignThemes.Wpf/Snackbar.cs
greuelpirat: 👍

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Snackbar.xaml
greuelpirat: I moved it.

MaterialDesignThemes.Wpf/Snackbar.cs
Keboo: Perhaps something like this. Create an attached property on SnackbarMessage (something like `SingleLineHeight`). Then set the attached property on the snackbar in its style and default it to 53. Finally pass it into the converter as a third parameter. 
That way we can attach a name to the value, and it allows someone to override it if needed.

Thoughts?

Keboo:
As for the commit size I am not too worried about it. I tend to always squash merge, but it can be helpful splitting refactoring and feature work into different PRs.

MaterialDesignThemes.Wpf/Snackbar.cs
greuelpirat: I added `SnackbarMessage.InlineActionButtonMaxHeight` as you suggested.

Then I found the limiting `MaxHeight` on the `ContentPresenter`. I also added `SnackbarMessage.ContentMaxHeight` for this. But to make it consistent with the first property I moved its usage to the wrapping `DockPanel` and adapted the value. That way both values relates to the SnackbarMessage.

PR #1749, opened 2020-04-19
Page for ComboBoxes in demo app

greuelpirat:
Resolves #1704.

* Created own page for ComboBoxes
* Moved all ComboBoxes from Field-Page
* Rearranged fields on Field-Page

PR #1772, opened 2020-04-26
Set size of DashedLine explicitly

### LINKED ISSUE #1735: DrawerHost crash when there's a textbox inside
DrawerHost crash when there's a textbox inside.
Problem happens only in version 3.1.0 of MaterialDesignThemes. 
I use Visual Studio and NuGet, .net framework 4.8

greuelpirat:
Resolves #1735 

I found that this trigger caused the high CPU usage:

https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/71af77775b205fd3beed48741b240ae6876ac747/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.TextBox.xaml#L278-L285

So I guess that the disabled `DashedLine` (disabled, because of placement in `RightDrawerContent`) had no fix size. So I set it: https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/71af77775b205fd3beed48741b240ae6876ac747/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.TextBox.xaml#L165-L166

and the CPU usage was gone.

Beside `MaterialDesignTheme.TextBox.xaml` I updated other controls, since they seems to have a similar `DashedLine`
* `MaterialDesignTheme.ComboBox.xaml`
* `MaterialDesignTheme.DatePicker.xaml`
* `MaterialDesignTheme.PasswordBox.xaml`
* `MaterialDesignTheme.TimePicker.xaml`

PS: I also updated the Fields.xaml in demo app, where I forgot the disabled fields during #1749 :-(


MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.ComboBox.xaml
```diff
@@ -486,6 +486,8 @@
                   Grid.ColumnSpan="2"
                   Grid.Column="0"
                   VerticalAlignment="Bottom"
+                  Height="{Binding BorderThickness.Bottom, ElementName=toggleButton, Converter={StaticResource DivisionMathConverter}, ConverterParameter=0.75}"
```
Keboo, line 489: Though this works, binding to `BorderTickness.Bottom` exposes a [memory leak](https://blog.jetbrains.com/dotnet/2014/09/04/fighting-common-wpf-memory-leaks-with-dotmemory/) (the Binding leak section) because `BorderThickeness` is a just a [Thickness](https://docs.microsoft.com/dotnet/api/system.windows.thickness) struct. It actually appears that we have the same bug on the `StrokeThickness` as well. I would propose the following fixes:
1. Since it looks like we are trying to keep these properties in sync, why not simply bind the Height to the StrokeHeight property. `{Binding StrokeHeight, RelativeSource={RelativeSource Self}}`
2. Rather than binding through the BorderThickness property, lets create a value converter that takes in the `BorderThickness` directly, and applies the appropriate math we need for this line. I would propose that this converter be made internal for now.

greuelpirat: Good points. I have implemented those. 👍

PR #1777, opened 2020-04-28
Added light/dark theme toggle button

### LINKED ISSUE #1773: Add Light/Dark theme toggle to the demo app header
Right now you have to toggle back and forth between the theme pages, and the various control pages when you want to see how a style looks in dark vs light theme. We should add a shortcut to the demo app's header to all for each changing of the theme without the need to swap back and forth. 

I would propose that this toggle be placed in the popup box, that is in the top right corner of the main window. 
image

Jesus805:
Fixes #1773 
image
I added a button that toggles between light and dark theme.


Keboo:
Great job. Thank you.

Issue #1573, opened 2019-12-25
Issues with DataGrid columns


ahmediddma:
the down arrow for the drop down list (at the end of the DataGridComboBoxColumn) is not the same size or position as the ComboBox and  its background can't be set.

validation with INotifyDataErrorInfo for each columns doesn't fully work the column will have red borders but no text.

https://github.com/ahmediddma/WPFDataGridValidationMDIX

Keboo:
The issue here is that the validation template does not match the normal material design:
![image](https://user-images.githubusercontent.com/952248/80447172-88062a00-88cd-11ea-9052-ba44094680ee.png)
This should present as a red underline similar to how the TextBox does:
![image](https://user-images.githubusercontent.com/952248/80447238-b97ef580-88cd-11ea-9baf-5dae15370896.png)

greuelpirat:
> the down arrow for the drop down list (at the end of the DataGridComboBoxColumn) is not the same size or position as the ComboBox and its background can't be set.

is solved with #1789

> validation with INotifyDataErrorInfo for each columns doesn't fully work the column will have red borders but no text.

is solved with #1787

PR #1787, opened 2020-05-02
Revised DataGrid-Text-Cell

### LINKED ISSUE #1779: Validation red border around both cell and textbox
This issue may be related to #1573
When I have a validation error with a `System.ComponentModel.DataAnnotations` attribute. Both the padded textbox and the surrounding cell get a red border. 
image
I think the border should only be around the cell, even though DataGrids use a Textbox as their entry type, DataGrids seem more aligned with Material Data Tables, which unfortunately does not specify what a validation error should look like

greuelpirat:
Fixes #1779 (mostly). There were multiple issues I made several changes. Because they depend on each other I did not split them to different pull requests.

* 6b72362 fixes the default textbox style:
image to image
Validation:
image to image
* 9f0fa72 fixes the ValidationErrorTemplate for Text-Cells in read only (for TextBlock).
* db8548e2 sets the red underline
image
* f7dd4ef1 finally fixes the ValidationError-Display for Popup-Text-Cells:
image to image
* db655c7, 3300fe5, dfe98d3, af8bebe, 188e6b05 updates the grids page in the demo app to make these changes testable.

I could not reproduce the red border for the cell, that is still open.


greuelpirat:
Fixes #1573

greuelpirat:
I fixed the red border on the cell, an some offset issues.

There is a default-style issue for the text column remaining. See https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/1779#issuecomment-623117242

Otherwise it is done.

Keboo:
Great, thank you. Will review soon.

Jesus805:
@greuelpirat The DataGridCell Textbox margin is off center. It's set to 13,8,8,8.
I'm wondering if this has something to do with DataGridCellAssist which is also 13,8,8,8

greuelpirat:
> @greuelpirat The DataGridCell Textbox margin is off center. It's set to 13,8,8,8.
> I'm wondering if this has something to do with DataGridCellAssist which is also 13,8,8,8

@Jesus805 Yes, that is the default padding. As far as I understood, this padding should align the cells with the header element. But there are still issues. Especially with order-button.

image

I will take a look, but this will be a new pull request.

Jesus805:
@greuelpirat Ah, that makes sense. Perhaps we make the right margin 13 too? Now that there's an underline it's a little more obvious that it's not aligned

PR #1790, opened 2020-05-02
TextFieldAssist.DecorationVisibility for FilledTextField

### LINKED ISSUE #1609: TextBox Underline doesn't respect DecorationVisibility="Hidden" if HasFilledTextField="True"
`<TextBox materialDesign:TextFieldAssist.HasFilledTextField="True" materialDesign:TextFieldAssist.DecorationVisibility="Hidden" BorderThickness="0" />`

Underline is still shown. Not shown if `HasFilledTextField="False"`

If I have some free time this week I'll fix it myself.

greuelpirat:
fixes #1609

I also rounded the bottom corners, if the bottom line is gone. Think that looks better:

image


Keboo:
Though I agree that rounding the corners looks better. The [spec](https://material.io/components/text-fields#anatomy) calls for the bottom corners to be square on the filled text box.

PR #1791, opened 2020-05-02
Added support for vertical toggle list box

DiegoBM:
Added support for tool vertical toggle list box, with sensible defaults.
All tests pass and visual test performed

Please feel free to mention anything that might not be correct so that I have the chance to make it right


Keboo:
Great work. Made one minor update to support virtualization.

DiegoBM:
Thank you @Keboo! I actually read about the virtualization in StackOverflow after submitting the PR and I was going to add it as well this morning, thank you for the modification :).

PR #1826, opened 2020-05-13
DialogContent doesn't inheritingTexElement Property

yesuraj130:
In DialogHost control template, inheritable property TextElement was overridden for DialogContent. This causes we have to specify TextElement property once again in DialogContent UIElement same as specified in mainwindow.

Where:
In DialogHost control template, under "PART_PopupContentElement.

Proposed action:
Please remove following lines in favor inheriting parent property,
TextElement.FontWeight="Regular"
TextElement.FontSize="13"


Keboo:
Thanks for submitting this, however the Popup gets displayed as a different window, so its contents don't inherit the properties from the DialogHost.

PR #1827, opened 2020-05-13
[AdornerAssist] Refactoring & Tests

w-syss:
Refactored AdornerAssist to improve code quality and increase robustness.

- Added braces to single line statements
- Throws exceptions on invalid operations
- Added tests
- Renamed functions to more closely portray what a function does
- Fixed comments


MaterialDesignThemes.Wpf/AdornerAssist.cs
```diff
@@ -1,28 +1,44 @@
...
 {
     /// <summary>
-    /// this is a simple utility to add and remove a single adorner to an element since there is no built-in way to do that in xaml.
-    /// <a href="https://docs.microsoft.com/en-us/dotnet/framework/wpf/controls/adorners-overview">see here</a>
+    /// This is a simple utility to add and remove a single adorner to an element
+    /// since there is no built-in way to do that in xaml.
+    /// <a href="https://docs.microsoft.com/en-us/dotnet/framework/wpf/controls/adorners-overview">See here</a>
     /// </summary>
-    internal static class AdornerAssist
+    public static class AdornerAssist
```
Keboo, line 13: Do you have a need for this to be public? In general I try to leave things private until there is a need for them to be public.

MaterialDesignThemes.Wpf/AdornerAssist.cs
```diff
@@ -1,28 +1,44 @@
...
     {
-        public static void AddAdorner<T>(this UIElement element, T adorner) where T : Adorner
+        public static void AddAdornerOfType<T>(this UIElement element, T adorner) where T : Adorner
```
Keboo, line 15: Rather than adding "OfType" to the method name, can we rename the generic parameter from `T` to `TAdorner` instead?

MaterialDesignThemes.Wpf/AdornerAssist.cs
w-syss: Originally the idea was to align this class with other "*Assist" classes.
Thinking further about it, i think this class is misnamed. I would propose
to keep the modifier internal and name the class "UIElementExtensions".

This more closely describes what this class actually provides and also serves as an entrypoint for further extensions of the UIElement class.

MaterialDesignThemes.Wpf/AdornerAssist.cs
w-syss: I can agree with that. The OfType in this case is redundant information.
Renaming the generic parameter is sufficient.

MaterialDesignThemes.Wpf/UIElementExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+                throw new ArgumentNullException(nameof(adorner));
+            }
+            
+            var adornerLayer = AdornerLayer.GetAdornerLayer(element)
+                ?? throw new InvalidOperationException("This element has no adorner layer.");
+            
+            adornerLayer.Add(adorner);
+        }
+
+        public static void RemoveAdornersOfType<T>(this UIElement element) where T : Adorner
```
Keboo, line 28:
```suggestion
        public static void RemoveAdorners<TAdorner>(this UIElement element) where T : Adorner
```

PR #1852, opened 2020-05-21
Tests to verify color definitions

greuelpirat:
These tests verifies, that colors are registered correctly in ...
* `MaterialDesignTheme.Light.xaml` and `MaterialDesignTheme.Dark.xaml`
* `ITheme`
* `IBaseTheme` (`MaterialDesignLightTheme` and `MaterialDesignDarkTheme`)
* `ResourceDictionaryExtensions`
* `ThemeExtensions`
* and `Color` values are identical

Theses tests showed, that `ToolForeground` and `ToolBackground` were also missing. So I added them 7372bd0

### DIFF
--- a/MaterialDesignThemes.Wpf.Tests/ColorDefinitionsTests.cs
+++ b/MaterialDesignThemes.Wpf.Tests/ColorDefinitionsTests.cs
@@ -1,0 +1,102 @@
+using System;
+using System.Collections;
+using System.Linq;
+using System.Windows;
+using System.Windows.Media;
+using Xunit;
+
+namespace MaterialDesignThemes.Wpf.Tests
+{
+    public class ColorDefinitionsTests
+    {
+        public ColorDefinitionsTests()
+        {
+            if (!UriParser.IsKnownScheme("pack"))
+                Assert.NotNull(new Application());
+        }
+
+        [Fact]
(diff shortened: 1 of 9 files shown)


MaterialDesignThemes.Wpf.Tests/ColorDefinitionsTests.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            AssertThemeColorsInXaml(new MaterialDesignDarkTheme(), "MaterialDesignTheme.Dark.xaml");
+        }
+
+        private static void AssertThemeColorsInXaml(IBaseTheme baseTheme, string xaml)
+        {
+            var theme = new Theme();
+            theme.SetBaseTheme(baseTheme);
+
+            var uri = new Uri("pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/" + xaml);
```
Keboo, line 78: There is an MdixHelper class that has some of this functionality already in it to load up the theme dictionaries. Might need to make some of the methods public to leverage them here.

greuelpirat: I have updated MdixHelper and the unit test.

PR #1864, opened 2020-05-26
Unify converter access modifiers to public.

watashidesu:
When overwriting the default style with a modification of the default style template, the converter may not be referenced.
Please change the access modifiers of some converters if possible.

### DIFF
--- a/MaterialDesignThemes.Wpf/Converters/BorderClipConverter.cs
+++ b/MaterialDesignThemes.Wpf/Converters/BorderClipConverter.cs
@@ -6,7 +6,7 @@ using System.Windows.Media;
 
 namespace MaterialDesignThemes.Wpf.Converters
 {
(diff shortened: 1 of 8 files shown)


Keboo:
Though I understand the desire for all of these to be public, often classes such as converters are added as internal when we are not sure the API, naming, is not yet ready for public consumption. 

Are there particular converters you were wanting to use?

watashidesu:
@Keboo 
The converter I wanted to use is `ExpanderDirectionConverter`.

> often classes such as converters are added as internal when we are not sure the API, naming, is not yet ready for public consumption.

I understand. Sorry for my lack of understanding.

Keboo:
@watashidesu ok, let me look into it and see about getting that one ready to be public.

watashidesu:
@Keboo
Thank you for b360fed11eef534f3b2078abd5ddba569a0b2962.

PR #1869, opened 2020-05-28
Refactoring of Card

w-syss:
- Removed class comment (added no value, visual clutter)
- Streamlined dependency properties
- Beautified OnRenderSizeChanged

### DIFF
--- a/MaterialDesignThemes.Wpf/Card.cs
+++ b/MaterialDesignThemes.Wpf/Card.cs
@@ -5,15 +5,37 @@ using System.Windows.Media;
 
 namespace MaterialDesignThemes.Wpf
 {
-    /// <summary>
-    /// A card is a content control, styled according to Material Design guidelines.
-    /// </summary>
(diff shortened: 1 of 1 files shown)


MaterialDesignThemes.Wpf/Card.cs
```diff
@@ -5,15 +5,37 @@
...
 namespace MaterialDesignThemes.Wpf
 {
-    /// <summary>
-    /// A card is a content control, styled according to Material Design guidelines.
-    /// </summary>
     [TemplatePart(Name = ClipBorderPartName, Type = typeof(Border))]
     public class Card : ContentControl
     {
+        private Border _clipBorder;
+        private static readonly double DefaultUniformCornerRadius = 2.0;
```
Keboo, line 12: Any reason to not make this a `const`?

w-syss: No, i was just used to defaults being reference types. Const is certainly more appropriate.

PR #1873, opened 2020-05-30
Changed colors for outlined button

### LINKED ISSUE #1863: Buttons have different appearance from demo app
Issue:
------

Buttons have missing hover background color and outlined button has purple outline instead of gray.

Steps to reproduce:
---------------------
Using Visual Studio Community 2019:
- Follow [Super Quick Start guide](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/wiki/Super-Quick-Start)
- Add button to main window with `MaterialDesignFlatButton` style
- Add button to main window with `MaterialDesignOutlinedButton` style

Result:
-------
Capture

XAML:
-------
```
<Grid>
    <StackPanel HorizontalAlignment="Center" VerticalAlignment="Center">
        <Button Margin="0 10" Style="{StaticResource MaterialDesignFlatButton}">MaterialDesignFlatButton</Button>
        <Button Margin="0 10" Style="{StaticResource MaterialDesignOutlinedButton}">MaterialDesignOutlinedButton</Button>
    </StackPanel>
</Grid>
```

greuelpirat:
Changed outlined button to match https://material.io/components/buttons#outlined-button:

* Grey border
* Brighter hover background

I also removed a redundant property.

Resolves #1863

### DIFF
--- a/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Button.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Button.xaml
@@ -210,7 +210,6 @@
     <Style x:Key="MaterialDesignOutlinedButton" TargetType="{x:Type ButtonBase}" BasedOn="{StaticResource MaterialDesignFlatButton}">
         <Setter Property="BorderBrush" Value="{DynamicResource MaterialDesignDivider}"/>
         <Setter Property="BorderThickness" Value="1"/>
-        <Setter Property="wpf:RippleAssist.Feedback" Value="{DynamicResource PrimaryHueMidBrush}" />
         <Setter Property="wpf:ButtonAssist.CornerRadius" Value="2" />
         <Setter Property="Template">
             <Setter.Value>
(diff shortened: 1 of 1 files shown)


jizc:
A curious wrinkle to this, is that in the interactive demo at the top of the Button definition page, the outline is not grey.

greuelpirat:
> A curious wrinkle to this, is that in the interactive demo at the top of the Button definition page, the outline is not grey. But it appears to be the only instance of it, everywhere else I can find it's grey.

I can not find it. Can you post a screenshot?

jizc:
Sure thing 👍 

image

To find it just follow the link you posted to material.io, then scroll all the way to the top of the page, then click the OUTLINED tab of the interactive demo.

jizc:
My guess is that the outline previously used the primary color, and they forgot to update the interactive demo when they changed it to grey.

greuelpirat:
Oh, I have overlooked this. Hm, but I would agree to say that they changed that. MDIX also uses primary color so may be that is the reason. Personally I find the grey border better, but that can decide should decide the reviewers.

Keboo:
Yea I agree, I think the gray outline is the intended color from the spec with their demo simply being out of date.

midasw:
@greuelpirat I believe there is another (cosmetic) issue with this button which is visible when changing from mouseover state to pressed state. When the button is pressed the existing background color disappears completely before the ripple appears. This doesn't seem to happen with any of the other button styles.

PR #1874, opened 2020-05-31
Fix thread unsafe code in SnackbarMessageQueue.

bombipappoo:
HashSet and LinkedList are not thread safe.
Add lock when accessing _pairedSnackbars and _snackbarMessages.


bombipappoo:
Example 3 on MaterialDesignDemo's Snackbar page says "Messages can be queued from any thread".
In any case, since PumpAsync() is running on worker thread, it needs to be synchronized.

Unfortunately, .NET does not have a thread safe LinkedList or HashSet, maybe.

PR #1876, opened 2020-05-31
Properly stop the Task waiting for an event.

bombipappoo:
In method `SnackbarMessageQueue.WaitForCompletionAsync`:  
If the Snackbar is closed by time span, the second Task remains without completion.
On the other hand, if the Snackbar is closed by `ActionClickEvent`, the first Task remains without completion.

In the Visual Studio Threads window, you can find the thread whose location is `mscorlib.dll!System.Threading.WaitHandle.InternalWaitOne` or `MaterialDesignThemes.Wpf.dll!MaterialDesignThemes.Wpf.SnackbarMessageQueue.WaitForCompletionAsync.AnonymousMethod__0`.

### DIFF
--- a/MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs
+++ b/MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs
@@ -56,11 +56,11 @@ namespace MaterialDesignThemes.Wpf
                 };
             }
 
-            public WaitHandle WaitHandle => _waitHandle;
+            public EventWaitHandle WaitHandle => _waitHandle;
 
-            private void UiElementOnMouseLeave(object sender, MouseEventArgs mouseEventArgs)
+            private async void UiElementOnMouseLeave(object sender, MouseEventArgs mouseEventArgs)
             {
-                Task.Factory.StartNew(() =>
+                await Task.Run(() =>
                 {
                     try
                     {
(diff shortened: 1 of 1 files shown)


bombipappoo:
When MaterialDesignDemo is launched, there are two "Worker Thread" as shown below.
This is due to the "Welcome to ..." Snackbar message on startup.

ThreadsView1

Now, if you click the Send button of Example 4 on the Snackbar page, the Snackbar will appear 6 times.
Then, the "Worker Thread" is increased by 6 as below.

ThreadsView2

Also, if you close the Snackbar with the TRACE button, you will see another "Worker Thread" is increased.

ThreadsView3

In this way, each time you display the Snackbar message, the "Worker Thread" is increased by one.
This is due to the following situation in the method `SnackbarMessageQueue.WaitForCompletionAsync()`.

### When closed by timeout

The following code starts the thread that waits for `actionClickWaitHandle` to be signaled.
However, when the Snackbar is closed, no one can set `actionClickWaitHandle` to the signaled state.
Therefore, this thread never ends.

```C#
    Task.Factory.StartNew(actionClickWaitHandle.WaitOne)
```

### When closed by `ActionClick`

`mouseNotOverManagedWaitHandle.WaitHandle` is signaled when the mouse is not on the Snackbar.
If the action button was clicked, it means the mouse is on the Snackbar, so `mouseNotOverManagedWaitHandle.WaitHandle` is the non-signaled state. When Snackbar is closed, `MouseNotOverManagedWaitHandle.UiElementOnMouseLeave()` is called, but since `MouseNotOverManagedWaitHandle.Dispose()` is called before that, with the code below, the event will not be signaled.
```C#
    if (!_isWaitHandleDisposed)
        _waitHandle.Set();
```
Therefore, the thread below never ends either.
```C#
    Task.Factory.StartNew(() =>
    {
        WaitHandle.WaitAll(new[]
        {
            mouseNotOverManagedWaitHandle.WaitHandle,
            durationPassedWaitHandle
        });
    })
```

Keboo:
Great work. Thank you.

Issue #1885, opened 2020-06-03
MaterialDesignTabRadioButton does not support ButtonProgressAssist


talapoin:
ButtonProgressAssist is not supported in MaterialDesignTabRadioButton style even though you can add attach ButtonProgressAssist to radio button with that style.
I'll suggest a PR with a change.

PR #1886, opened 2020-06-04
ButtonProgressAssist support in MaterialDesignTabRadioButton

talapoin:
Support ButtonProgressAssist in MaterialDesignTabRadioButton. Related to #1885 
image


Keboo:
Thanks for this work. I am curious to understand the use case for this. Can you explain what you are trying to accomplish with this type of UI interaction?

talapoin:
I needed to indicate a running process in underlying "tab".  Imagine you don't have the indication and you start a long time process in **First tab** and then go **second tab**. Now if you want to see status of the process you must go back to first tab. Our users didn't like it. So we've decided to replicate _windows like_ functionality from task bar.

Keboo:
@talapoin thank you for the explanation. Great work.

PR #1890, opened 2020-06-04
Enabled strong name for WPF projects

shjin404:
MaterialDesignInXamlToolKit is missing strong-name. Some organizations require their assemblies to be strong-named due to various reasons including historical reasons, and it causes a blocking issue in some cases especially when build time conversion of NuGet assemblies to strong name is not acceptable.

Given Microsoft's new recommendation for open source project, I am including the private key as-is to the repository as strong name is no longer recognized as security mechanism but purely a unique identity mechanism.

See https://docs.microsoft.com/en-us/dotnet/standard/assembly/strong-named

Given people are consuming this by NuGet package, this change should not cause issues as building their project again with NuGet with strong named assemblies would generate correct new reference metadata.


greuelpirat:
I can not start MaterialDesignDemo any more:

```
System.IO.FileLoadException: Die Datei oder Assembly "ShowMeTheXAML, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null" oder eine Abhängigkeit davon wurde nicht gefunden. Eine Assembly mit starkem Namen ist erforderlich. (Ausnahme von HRESULT: 0x80131044)
Dateiname: "ShowMeTheXAML, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null"
   bei MaterialDesignColors.WpfExample.App.OnStartup(StartupEventArgs e)
...
```

I think this is because of the strong name changes. Any else have this problem?

Keboo:
@greuelpirat yea that was my fault. Should be fixed now.

PR #1892, opened 2020-06-05
Small implementation suggesting to open PopupBox on Keyboard Focus.

HClausing:
Not tested to cycle focus on children elements with TabStop yet.


HClausing:
I'll work on combing both Keyboard and Mouse behaviors. When use the KeyboardFocus enum, all the mouse operations become broken.

PR #1917, opened 2020-06-13
Disabled strong name for demo apps

bombipappoo:
Since strong name was enabled at #1890, demo apps cannot run on .NET Framework anymore.
Demo apps uses ShowMeTheXAML and Dragablz, but they have not been signed.


Keboo:
Ah shoot, yea that was my fault. Thanks for fixing it.

Issue #1878, opened 2020-06-01
If SnackBar items queue is not empty while closing the application, application remains stalled.


gitjsdr26:
If SnackBar items queue is not empty while closing the application, application remains stalled.
The application is locked into the loop `private void PumpAsync()`.

```C#
private void PumpAsync()
{
	while (!_isDisposed)
	{
		var eventId = WaitHandle.WaitAny(new WaitHandle[] {_disposedEvent, _messageWaitingEvent});
		if (eventId == 0) continue;
		var exemplar = _pairedSnackbars.FirstOrDefault();
		if (exemplar == null)
		{
			Trace.TraceWarning(
				"A snackbar message as waiting, but no Snackbar instances are assigned to the message queue.");
			_disposedEvent.WaitOne(TimeSpan.FromSeconds(1));
			continue;
		}

		//find a target
		var snackbar = FindSnackbar(exemplar.Dispatcher).Result;

		//show message
		if (snackbar != null)
		{
			var message = _snackbarMessages.First.Value;
			_snackbarMessages.RemoveFirst();
			if (_latestShownItem == null
				|| message.IsPromoted
				|| !Equals(_latestShownItem.Item1.Content, message.Content)
				|| !Equals(_latestShownItem.Item1.ActionContent, message.ActionContent)
				|| _latestShownItem.Item2 <= DateTime.Now.Subtract(_messageDuration))
			{
				ShowAsync(snackbar, message).Wait();
				_latestShownItem = new Tuple<SnackbarMessageQueueItem, DateTime>(message, DateTime.Now);
			}
		}
		else
		{
			//no snackbar could be found, take a break
			_disposedEvent.WaitOne(TimeSpan.FromSeconds(1));
		}

		if (_snackbarMessages.Count > 0)
			_messageWaitingEvent.Set();
		else
			_messageWaitingEvent.Reset();
	}
}
```

`_snackbarMessages.Count` is equal to 2, then it loops indefinitely.

PR #1922, opened 2020-06-14
Refactored SnackbarMessageQueue to use Dispatcher

greuelpirat:
I reworked the `SnackbarMessageQueue` to no longer start its own long running task. Instead it uses the `Dispatcher` to _pull_ for next item.

That makes it necessary that a new instance of `SnackbarMessageQueue` must be created from within a `Dispatcher`-thread. But this should be the standard case any way. But it is a new limitation.

This relates to #1878. Although I could not reproduce it. But I personally did not like it anyway, that the `SnackbarMessageQueue` requires its own thread.

### DIFF
--- a/MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs
+++ b/MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs
@@ -23,40 +23,27 @@ namespace MaterialDesignThemes.Wpf
         private int _pauseCounter;
         private bool _isDisposed;
 
-        #region private class MouseNotOverManagedWaitHandle : IDisposable
+        #region MouseNotOverManagedWaitHandle
 
         private class MouseNotOverManagedWaitHandle : IDisposable
         {
+            private readonly UIElement _uiElement;
             private readonly ManualResetEvent _waitHandle;
             private readonly ManualResetEvent _disposedWaitHandle = new ManualResetEvent(false);
-            private Action _cleanUp;
-            private bool _isWaitHandleDisposed;
+            private bool _isDisposed;
             private readonly object _waitHandleGate = new object();
 
             public MouseNotOverManagedWaitHandle(UIElement uiElement)
             {
-                if (uiElement == null) throw new ArgumentNullException(nameof(uiElement));
-
+                _uiElement = uiElement ?? throw new ArgumentNullException(nameof(uiElement));
                 _waitHandle = new ManualResetEvent(!uiElement.IsMouseOver);
                 uiElement.MouseEnter += UiElementOnMouseEnter;
                 uiElement.MouseLeave += UiElementOnMouseLeave;
-
-                _cleanUp = () =>
-                {
-                    uiElement.MouseEnter -= UiElementOnMouseEnter;
-                    uiElement.MouseLeave -= UiElementOnMouseLeave;
-                    lock (_waitHandleGate)
-                    {
-                        _waitHandle.Dispose();
-                        _isWaitHandleDisposed = true;
-                    }
-                    _disposedWaitHandle.Set();
-                    _disposedWaitHandle.Dispose();
-                    _cleanUp = () => { };
-                };
             }
 
             public EventWaitHandle WaitHandle => _waitHandle;
+
+            private void UiElementOnMouseEnter(object sender, MouseEventArgs mouseEventArgs) => _waitHandle.Reset();
 
             private async void UiElementOnMouseLeave(object sender, MouseEventArgs mouseEventArgs)
             {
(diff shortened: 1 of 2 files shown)


bombipappoo:
I think that `_pairedSnackbarsLock` is no longer needed since `_pairedSnackbars` is only be accessed on the dispatcher thread.

MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs
```diff
@@ -152,11 +136,12 @@ public SnackbarMessageQueue()
         public SnackbarMessageQueue(TimeSpan messageDuration)
         {
             _messageDuration = messageDuration;
-            Task.Factory.StartNew(PumpAsync, TaskCreationOptions.LongRunning);
+            _dispatcher = Dispatcher.FromThread(Thread.CurrentThread)
+                          ?? throw new InvalidOperationException("SnackbarMessageQueue must be created in a dispatcher thread");
```
w-syss, line 140: Why are we not using Application.Current.Dispatcher?

MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs
```diff
@@ -68,10 +57,10 @@ private async void UiElementOnMouseLeave(object sender, MouseEventArgs mouseEven
                     }
                     catch (ObjectDisposedException)
                     {
-                        /* we are we suppresing this? 
+                        /* we are we suppressing this? 
                          * as we have switched out wait onto another thread, so we don't block the UI thread, the
                          * _cleanUp/Dispose() action might also happen, and the _disposedWaitHandle might get disposed
-                         * just before we WaitOne. We wond add a lock in the _cleanUp because it might block for 2 seconds.
+                         * just before we WaitOne. We want add a lock in the _cleanUp because it might block for 2 seconds.
```
w-syss, line 63: I think this was supposed to say won't :D

MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs
```diff
@@ -80,66 +69,61 @@ private async void UiElementOnMouseLeave(object sender, MouseEventArgs mouseEven
...
                 //this keeps the event waiting simpler, rather that actually watching play -> pause -> play -> pause etc
                 var granularity = TimeSpan.FromMilliseconds(200);
 
                 Task.Factory.StartNew(() =>
                 {
                     //keep upping the completion time in case it's paused...
-                    while (DateTime.Now < _completionTime && !ceaseWaitHandle.WaitOne(granularity)
-                        && !signalWhenDurationPassedWaitHandle.WaitOne(TimeSpan.Zero))
+                    while (DateTime.Now < completionTime && !ceaseWaitHandle.WaitOne(granularity)
+                                                          && !signalWhenDurationPassedWaitHandle.WaitOne(TimeSpan.Zero))
```
w-syss, line 115: Maybe extract that condition with a fitting name?

MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs
```diff
@@ -155,8 +139,8 @@ public SnackbarMessageQueue(TimeSpan messageDuration)
             Task.Factory.StartNew(PumpAsync, TaskCreationOptions.LongRunning);
         }
 
-        //oh if only I had Disposable.Create in this lib :)  tempted to copy it in like dragabalz, 
-        //but this is an internal method so no one will know my direty Action disposer...
+        //oh if only I had Disposable.Create in this lib :)  tempted to copy it in like dragablz, 
+        //but this is an internal method so no one will know my direct Action disposer...
```
bombipappoo, line 144: I think this was "dirty".

MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs
greuelpirat: `Application.Current.Dispatcher` is not clear in applications with multiple dispatchers (I have such in my application).

I took this from the `DispatcherObject` where it gets its dispatcher with `Dispatcher.CurrentDispatcher`. But this returns a new Dispatcher of the current does not have one. See [here](https://referencesource.microsoft.com/#windowsbase/Base/System/Windows/Threading/Dispatcher.cs).

If you like I can change it to `Dispatcher.CurrentDispatcher`

MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs
greuelpirat: sure, you are right. I did not really read the comments. just picked the first spell-check suggestion :-)

MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs
greuelpirat: sure, I will change that

MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs
greuelpirat: I updated this method

greuelpirat:
Just for better understanding. Now every `SnackbarMessageQueue.Enqueue` call invokes `ShowNextAsync` on the dispatcher. First step there is to await the semaphore signal:

https://github.com/greuelpirat/MaterialDesignInXamlToolkit/blob/76212a5a2af3314f20898f9cd771f907cde2e194/MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs#L256

The semaphore is set to only have one signal at a time. When the message is completed the semaphore releases this signal for the next message. So when 10 messages are queued there a 10 tasks where each task picks the next message (not the one it triggered) to show.

bombipappoo:
If we have multiple dispatchers e.g., dispatcherA and dispatcherB, on the context of dispatcherB, the message queue created on dispatcherA can be assigned to the snackbar created on dispatcherB.
This is incorrect, so while setting `Snackbar.MessageQueue` `InvalidOperationException` should be thrown instead.

PR #1923, opened 2020-06-14
Update NumericUpDown theme to closer match other fields

yosefmah:
There were a couple of minor visual issues I noticed with the MaterialDesignNumericUpDown theme so this PR tweaks the following to closer match the behaviour of other fields:

- Change the background to Transparent  (didn't match dark or custom backgrounds)
- Also applied the transparent background to the DataGrid NumericUpDown editing template
- Remove the extra white underline when focused
- Changed the opacity of the disabled state
- Added the bottom line dashed adorner

I also thought about changing the context menu to the MDIX one, I can add this if it is not going to break anything.

Before:

d8d7816a99664bf15bf89985ea7d09ce

After:

f8813b073458fa19f22a329b4fdd773d


Keboo:
Great work. Thank you.

PR #1932, opened 2020-06-19
Make sure that SnackbarMessageQueue has the same dispatcher as Snackbar.

bombipappoo:
Prevent SnackbarMessageQueue from being assigned to a Snackbar created on different thread.

### DIFF
--- a/MaterialDesignThemes.Wpf/Snackbar.cs
+++ b/MaterialDesignThemes.Wpf/Snackbar.cs
@@ -53,7 +53,12 @@ namespace MaterialDesignThemes.Wpf
         public SnackbarMessageQueue MessageQueue
         {
             get => (SnackbarMessageQueue) GetValue(MessageQueueProperty);
-            set => SetValue(MessageQueueProperty, value);
+            set
+            {
+                if (!(value is null) && value.Dispatcher != Dispatcher)
+                    throw new InvalidOperationException("Objects must be created by the same thread.");
+                SetValue(MessageQueueProperty, value);
+            }
         }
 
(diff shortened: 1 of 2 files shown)


MaterialDesignThemes.Wpf/Snackbar.cs
```diff
@@ -53,7 +53,12 @@ private static void MessageQueuePropertyChangedCallback(DependencyObject depende
         public SnackbarMessageQueue MessageQueue
         {
             get => (SnackbarMessageQueue) GetValue(MessageQueueProperty);
-            set => SetValue(MessageQueueProperty, value);
+            set
+            {
+                if (!(value is null) && value.Dispatcher != Dispatcher)
+                    throw new InvalidOperationException("Objects must be created by the same thread.");
+                SetValue(MessageQueueProperty, value);
+            }
```
w-syss, line 61: The check if the dispatcher is correct or not has to be made in the callback. A property wrapper cannot contain additional logic. I think adding something like
```
if (messageQueue?.Dispatcher != Dispatcher)
{
    throw new InvalidOperationException("...");
}
```
to the MessageQueuePropertyChangedCallback would achieve the same thing.

bombipappoo: If an `Exception` is thrown when checked at the `MessageQueuePropertyChangedCallback`, the `MessageQueue` already has a new (invalid) value.
In such case, what is usually the right thing to do? Revert to `DependencyPropertyChangedEventArgs.OldValue` and then throw?

PR #1927, opened 2020-06-16
[DataGridAssist] Refactoring & Tests

w-syss:
- Streamlined attached properties
- Added basic tests for attached properties
- Extracted Defaults
- Simplified One-Click edit of comboboxes and checkboxes in datagrids

PR #1954, opened 2020-07-01
Fix clicks not registering on DataGridTemplateColumn

yosefmah:
Following the refactoring in #1927 the new AllowDirectEditWithoutFocus method does not allow clicks to pass through onto controls that were contained within a DataGridCellTemplate when not in focus.  These needed two clicks to register.

I have attempted a rather crude fix to use the older code in the event the click is on a DataGridTemplateColumn but I suspect there is a better way to do it!

### DIFF
--- a/MaterialDesignThemes.Wpf/DataGridAssist.cs
+++ b/MaterialDesignThemes.Wpf/DataGridAssist.cs
@@ -3,6 +3,8 @@ using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Controls.Primitives;
 using System.Windows.Input;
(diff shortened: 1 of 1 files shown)


Keboo:
Can you explain how this is fixing the issue? I am not sure I understand.

yosefmah:
Yes, quite right @Keboo .  Calling the second method is unnecessary since it doesn't actually do anything.  Instead, would it be better to just return the method when the click is on a DataGridTemplateColumn?

Keboo:
Yea I think that is better, will double check it soon and get it merged.

PR #1958, opened 2020-07-02
Add Prev and Next button for demo app items

AmRo045:
Preview

### DIFF
--- a/MainDemo.Wpf/Domain/MainWindowViewModel.cs
+++ b/MainDemo.Wpf/Domain/MainWindowViewModel.cs
@@ -25,6 +25,8 @@ namespace MaterialDesignColors.WpfExample.Domain
         private ObservableCollection<DemoItem> _allItems;
         private ObservableCollection<DemoItem> _demoItems;
         private DemoItem _selectedItem;
+        private bool _demoItemsPrevButtonStatus;
+        private bool _demoItemsNextButtonStatus;
 
 
         public string SearchKeyword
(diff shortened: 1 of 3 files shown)


MainDemo.Wpf/Domain/MainWindowViewModel.cs
```diff
@@ -25,6 +25,8 @@ public MainWindowViewModel(ISnackbarMessageQueue snackbarMessageQueue)
         private ObservableCollection<DemoItem> _allItems;
         private ObservableCollection<DemoItem> _demoItems;
         private DemoItem _selectedItem;
+        private bool _demoItemsPrevButtonStatus;
```
Keboo, line 28: Though this works, there is a much cleaners option.
1. Create a new `int` property in this view model and bind it to the `SelectedIndex` property of the `DemoItemsListBox`
2. Rather than using click handlers, and bindings for `IsEnabled` we can use commands. In the demo app there is a command implementation called `AnotherCommandImplementation`. Create one for move next, and one for move previous. The constructor will take in both an method to execute, and a method to determine if it can execute. The execute method should simply increment or decrement the int property you made above. The canExecute methods should return true if int property above is not at the end.

There is an example of doing this in `TreesViewModel.RemoveSelectedItemCommand`. That uses delegates rather than methods; either approach is fine.

AmRo045: Thanks for the recommendation.

PR #1986, opened 2020-07-22
[Expander] Rework & Performance Improvements

### LINKED ISSUE #1973: WPF UI freeze when populating listview with expanders
I have a ListView I'm populating with 40/50 items. Each item is an expander  with a body which contains a TextBox. The expanders are on IsExpanded = False.

When the list is displayed on the UI, there is a freeze of like 5/6 seconds until the items are displayed. After that, I can scroll and open them without problem.

The freeze seems to come from the TextBox content in the expander body which is a big string (between 10 and 20 lines). When I replace the text by "a", the display is instant and there is no freeze.

I tried to virtualize the listView without success...

Have you an idea what can I do to avoid the freeze ?

w-syss:
Resolves #1973

This is a quite large overhaul of the **Expander** control.
I tried to simplify templates where possible to make future contributions easier.
This pull request breaks the current api due to changing the naming of the _Down/Up/Right/Left_-Headerpaddings.

Performance evaluation for the provided project in #1973:
Old: ~ **47s**
New: ~**7s**
w/o style: ~**1s**
New w/o animations: ~**1s**


w-syss:
This is not finished yet, as i still have to include closing the expander when the bound collection gets cleared.

PR #1988, opened 2020-07-23
Bind Height of Hint and ClearButton to the text line height.

### LINKED ISSUE #1979: TextBox VerticalAlignment Bug

bombipappoo:
Fixes #1979.

Bound `Height` of `Hint` and `ClearButton` to the text line height instead of `ActualHeight`.
It seems that `ActualHeight` doesn't shrink after it expands.

### DIFF
--- a/MaterialDesignThemes.Wpf/Converters/FontToLineHeightConverter.cs
+++ b/MaterialDesignThemes.Wpf/Converters/FontToLineHeightConverter.cs
@@ -1,0 +1,22 @@
+﻿using System;
+using System.Globalization;
+using System.Windows.Data;
+using System.Windows.Media;
+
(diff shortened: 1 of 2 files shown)


Keboo:
After digging into it a little more I think a cleaner solution is to simply remove the height binding from the SmartHint altogether. I have put together a PR for it https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/pull/2005. 

Are there any other cases I should have tests for?

bombipappoo:
The height binding was added at #1900, so if simply remove it, it will be the same as before.

Specify `Margin` to `TextBox` as below, `ActualHeight` does not change at the unit test, but when you actually display it, you can see that the height changes.
```xml
    <TextBox Margin="0 0.5 0 0" Style=...
```

Issue #1883, opened 2020-06-02
TextFieldAssist.HasClearButton changes size of control


lszczygielek:
I found and issue with TextFieldAssist.HasClearButton="True". When my TextBox has it then when I type something and X icon appears, then control has a bigger height by 1 or 2 pixels. It makes size/location fluctuation on controls which are below.

This behavior is visible in demo application. Steps to reproduce:
Demo App > Fields > WiFi TextBox
1. Put value in TextBox.
2. Observe that controls below are moved a little bit down (all **Filled fields** section).

Keboo:
I can confirm that there is a sizing change occurring, but it appears to be caused by the floating hint rather than the clear button.
![TextBoxSize](https://user-images.githubusercontent.com/952248/83606446-75b38780-a52e-11ea-80ac-160c005e0642.gif)

PR #2005, opened 2020-08-01
Fixing height issue of the TextBox

### LINKED ISSUE #1979: TextBox VerticalAlignment Bug

Keboo:
Fixes #1979
Added UI tests to account for both that issue as well as #1883 which is the bug that caused the height binding to initially be added.


bombipappoo:
In demo app, remove the "WiFi" hint as below. Then enter or clear the text in the TextBox, you can see that the following controls are moved a bit.
```xml
                <TextBox Style="{StaticResource MaterialDesignFloatingHintTextBox}"
                         Text="Good stuff"
                         VerticalAlignment="Center"
                         materialDesign:TextFieldAssist.PrefixText="$"
                         materialDesign:TextFieldAssist.SuffixText="mm">
                </TextBox>
```
This tests a similar situation.
```csharp
        [Fact]
        [Description("Issue 1883")]
        public async Task OnClearButtonShown_ControlsLayoutDoesNotChange()
        {
            await using var recorder = new TestRecorder(App);

            //Arrange
            var stackPanel = await LoadXaml(@"
<StackPanel>
    <TextBox Style=""{StaticResource MaterialDesignFloatingHintTextBox}""
        materialDesign:TextFieldAssist.HasClearButton=""True""/>
    <TextBlock>Text</TextBlock>
</StackPanel>");
            var textBox = await stackPanel.GetElement("/TextBox");
            var textBlock = await stackPanel.GetElement("/TextBlock");

            var initialRect = await textBlock.GetCoordinates();

            //Act
            await textBox.SetText("a");

            //Assert
            var rect = await textBlock.GetCoordinates();
            Assert.Equal(initialRect, rect);

            recorder.Success();
        }
```

Keboo:
@bombipappoo thank you very much for your help on this. I think the current fix is likely the cleanest but I would like you feedback on it. In short I changed the SmartHint to only use the render transform instead of the layout transform so that it won't trigger size changes.

Keboo:
@bombipappoo thank you. For the clear button I have pulled in your fix for it.

PR #2016, opened 2020-08-09
Fixes validation message remains visible

### LINKED ISSUE #1969: Validation error message remains visible when moving through transitions
Posted by @ChrisCross67 [here](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/pull/1842#issuecomment-653526645)

image

I'm not sure where the problem lies, wrapping the entire control in an AdornerDecorator fixes it but would be good to sort it in the template.  I've tried but with no success.

Here is the sample XAML:

```
<materialDesign:Transitioner SelectedIndex="0">
        <StackPanel HorizontalAlignment="Center">
            <TextBlock Text="First slide" />
            <TextBox materialDesign:HintAssist.HelperText="Helper text longer than error" materialDesign:HintAssist.Hint="Name">
                <TextBox.Text>
                    <Binding Path="Name" UpdateSourceTrigger="PropertyChanged">
                        <Binding.ValidationRules>
                            <domain1:NotEmptyValidationRule ValidatesOnTargetUpdated="True" />
                        </Binding.ValidationRules>
                    </Binding>
                </TextBox.Text>
            </TextBox>

            <Button Command="{x:Static materialDesign:Transitioner.MoveNextCommand}"
                    Style="{DynamicResource MaterialDesignFloatingActionButton}">
                <materialDesign:PackIcon Kind="ArrowRight" />
            </Button>
        </StackPanel>
        <StackPanel HorizontalAlignment="Center">
            <TextBlock Text="Second slide" />
            <Button Command="{x:Static materialDesign:Transitioner.MovePreviousCommand}"
                    Style="{DynamicResource MaterialDesignFloatingActionButton}">
                <materialDesign:PackIcon Kind="ArrowLeft" />
            </Button>
        </StackPanel>
    </materialDesign:Transitioner>
```

bombipappoo:
Fixes #1969.

Made the `Opacity` of the validation error message has same value as the `Opacity` of `TransitionerSlide`, if the validatee is on the `Transitioner`.

### DIFF
--- a/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.ValidationErrorTemplate.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.ValidationErrorTemplate.xaml
@@ -18,7 +18,8 @@
             </DataTemplate>
         </ControlTemplate.Resources>
         <StackPanel>
-            <AdornedElementPlaceholder Name="Placeholder" />
+            <AdornedElementPlaceholder Name="Placeholder"
+                                       wpf:ValidationAssist.Validatee="{Binding ElementName=Placeholder, Path=AdornedElement}"/>
             <Border x:Name="DefaultErrorViewer"
                     Visibility="Collapsed"
                     Background="{Binding ElementName=Placeholder, Path=AdornedElement.(wpf:ValidationAssist.Background)}">
@@ -29,6 +30,7 @@
                            Margin="0 2"
                            TextWrapping="Wrap"
                            Text="{Binding CurrentItem.ErrorContent}"
+                           Opacity="{Binding Path=(wpf:ValidationAssist.ValidateeOpacity), ElementName=Placeholder}"
                            UseLayoutRounding="false" />
             </Border>
             <controlzEx:PopupEx x:Name="ValidationPopup"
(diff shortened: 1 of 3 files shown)


MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.ValidationErrorTemplate.xaml
```diff
@@ -29,6 +30,7 @@
                            Margin="0 2"
                            TextWrapping="Wrap"
                            Text="{Binding CurrentItem.ErrorContent}"
+                           Opacity="{Binding Path=(wpf:ValidationAssist.ValidateeOpacity), ElementName=Placeholder}"
```
Keboo, line 33: Would this make more sense to just directly bind to the element rather than introduce an additional attached property?
```suggestion
                           Opacity="{Binding ElementName=Placeholder, Path=AdornedElement.Opacity}"
```

bombipappoo: Since the `Transitioner` hides the slide by setting `TransitionerSlide.Opacity` to zero, `Opacity` of the elements on that slide does not change. `AdornedElement.Opacity` is also non-zero, so binding it does not hide the validation error message.
Therefore, `ValidationAssist.ValidateOpacity` that has the value of `TransitionerSlide.Opacity` should be used instead of `AdornedElement.Opacity`.

PR #2017, opened 2020-08-10
Fix height issue of the PasswordBox

bombipappoo:
Similar to #2005.

### DIFF
--- a/MaterialDesignThemes.UITests/WPF/PasswordBox/PasswordBoxTests.cs
+++ b/MaterialDesignThemes.UITests/WPF/PasswordBox/PasswordBoxTests.cs
@@ -22,18 +22,16 @@ namespace MaterialDesignThemes.UITests.WPF.PasswordBox
             var stackPanel = await LoadXaml(@"
 <StackPanel>
     <PasswordBox materialDesign:TextFieldAssist.HasClearButton=""True""/>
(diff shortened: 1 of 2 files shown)


Keboo:
I believe this still addressed the issue by looking only at the password box size/coordinates. GetCoordinates (probably needs a better name) returns a rect that includes the size of the PasswordBox. If the TextBox shifts down, it will be because the PasswordBox changes in size. After your fix it no longer changes size, and I think that is the key thing to have the test assert on.

bombipappoo:
In theory it is correct, but in actually it wasn't. Even before fixing it, `passwordBox.GetCoordinates()` was not asserted as it returns the same value.

Now, I found the reason:
`IVisualElement.GetCoordinates()` calls `Visual.PointToScreen()`, then `Visual.PointToScreen()` calls `ClientToScreen()` Windows API, so it always returns the rectangle with integer coordinates. As the result, the coordinate values were rounded, then same values were returned.

Therefore, I should have compared the values of `passwordBox.GetActualHeight()` instead of `textBox.GetCoordinates()` or `passwordBox.GetCoordinates()`.

Keboo:
Ah good catch, thank you. I will log a bug and get that fixed in the XAML Test library.

PR #2025, opened 2020-08-17
Alternate proposal fox fixing #1969

### LINKED ISSUE #1969: Validation error message remains visible when moving through transitions

Keboo:
@bombipappoo any thoughts on this fix vs #2016?

Fixes #1969.

### DIFF
--- a/MainDemo.Wpf/TransitionsDemo/Slide1_Intro.xaml
+++ b/MainDemo.Wpf/TransitionsDemo/Slide1_Intro.xaml
@@ -4,13 +4,26 @@
              xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
              xmlns:d="http://schemas.microsoft.com/expression/blend/2008" 
              xmlns:local="clr-namespace:MaterialDesignDemo.TransitionsDemo"
-             xmlns:materialDesign="http://materialdesigninxaml.net/winfx/xaml/themes"
+             xmlns:materialDesign="http://materialdesigninxaml.net/winfx/xaml/themes" 
+             xmlns:domain1="clr-namespace:MaterialDesignDemo.Domain"
              mc:Ignorable="d" 
              d:DesignHeight="300" d:DesignWidth="300"
              >
     <materialDesign:ColorZone Mode="Light" VerticalAlignment="Stretch" HorizontalAlignment="Stretch">
-        <StackPanel VerticalAlignment="Center">
-            <TextBlock>Welcome to the transitions showcase!  Take the red pill to begin...</TextBlock>
+        <StackPanel VerticalAlignment="Center" Margin="10">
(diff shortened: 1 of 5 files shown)


bombipappoo:
That makes sense. I didn't realize that at all.

But was there any reason why Previous was Visible?

bombipappoo:
I understood the reason. It's for fading animation.
So, I think the following is better than simply deleting it.

```xml
            <MultiDataTrigger>
                <MultiDataTrigger.Conditions>
                    <Condition Binding="{Binding State, RelativeSource={RelativeSource Self}}" Value="Previous" />
                    <Condition Binding="{Binding Opacity, RelativeSource={RelativeSource Self}, Converter={StaticResource NotZeroConverter}}" Value="True" />
                </MultiDataTrigger.Conditions>
                <Setter Property="Visibility" Value="Visible" />
            </MultiDataTrigger>
```

Keboo:
Agreed I was testing with some of transitions and coming up with something similar. Long term I would like to look at implementing [virtualization](https://docs.microsoft.com/en-us/dotnet/framework/wpf/advanced/optimizing-performance-controls#displaying-large-data-sets) which would also tie into this a little bit as well since the UI elements would not all be present, but that is a larger PBI for a different day.

PR #2029, opened 2020-08-20
Add the DialogHost close method.

HenJigg:
Add a manual method of closing the window, and in order to keep the syntax uniform, the original Close method updates the alias.

### DIFF
--- a/MaterialDesignThemes.Wpf/DialogHost.cs
+++ b/MaterialDesignThemes.Wpf/DialogHost.cs
@@ -166,19 +166,8 @@ namespace MaterialDesignThemes.Wpf
         /// <returns>Task result is the parameter used to close the dialog, typically what is passed to the <see cref="CloseDialogCommand"/> command.</returns>
         public static async Task<object> Show(object content, object dialogIdentifier, DialogOpenedEventHandler openedEventHandler, DialogClosingEventHandler closingEventHandler)
         {
-            if (content == null) throw new ArgumentNullException(nameof(content));
-
-            if (LoadedInstances.Count == 0)
-                throw new InvalidOperationException("No loaded DialogHost instances.");
-            LoadedInstances.First().Dispatcher.VerifyAccess();
-
-            var targets = LoadedInstances.Where(dh => dialogIdentifier == null || Equals(dh.Identifier, dialogIdentifier)).ToList();
-            if (targets.Count == 0)
-                throw new InvalidOperationException($"No loaded DialogHost have an {nameof(Identifier)} property matching {nameof(dialogIdentifier)} argument.");
-            if (targets.Count > 1)
-                throw new InvalidOperationException("Multiple viable DialogHosts.  Specify a unique Identifier on each DialogHost, especially where multiple Windows are a concern.");
-
-            return await targets[0].ShowInternal(content, openedEventHandler, closingEventHandler);
+            if (content is null) throw new ArgumentNullException(nameof(content));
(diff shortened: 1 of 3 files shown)


HenJigg:
Resolve naming conflicts.

MaterialDesignThemes.Wpf/DialogHost.cs
```diff
@@ -181,6 +181,22 @@ public static async Task<object> Show(object content, object dialogIdentifier, D
             return await targets[0].ShowInternal(content, openedEventHandler, closingEventHandler);
         }
 
+        /// <summary>
+        ///  Close a modal dialog.
+        /// </summary>
+        /// <param name="dialogIdentifier"> of the instance where the dialog should be shown. Typically this will match an identifer set in XAML. </param>
+        public static void CloseDailog(object dialogIdentifier)
```
bombipappoo, line 188:
```suggestion
        public static void CloseDialog(object dialogIdentifier)
```

MaterialDesignThemes.Wpf/DialogHost.cs
```diff
@@ -181,6 +181,22 @@ public static async Task<object> Show(object content, object dialogIdentifier, D
             return await targets[0].ShowInternal(content, openedEventHandler, closingEventHandler);
         }
 
+        /// <summary>
+        ///  Close a modal dialog.
+        /// </summary>
+        /// <param name="dialogIdentifier"> of the instance where the dialog should be shown. Typically this will match an identifer set in XAML. </param>
```
bombipappoo, line 187:
```suggestion
        /// <param name="dialogIdentifier"> of the instance where the dialog should be closed. Typically this will match an identifer set in XAML. </param>
```

MaterialDesignThemes.Wpf/DialogHost.cs
bombipappoo, line 188: I think a closeParameter need to be passed.

MaterialDesignThemes.Wpf/DialogHost.cs
```diff
@@ -181,6 +181,22 @@ public static async Task<object> Show(object content, object dialogIdentifier, D
...
+        public static void CloseDailog(object dialogIdentifier)
+        {
+            if (dialogIdentifier == null) throw new ArgumentNullException(nameof(dialogIdentifier));
+
+            if (LoadedInstances.Count == 0)
+                throw new InvalidOperationException("No loaded DialogHost instances.");
+
+            var targets = LoadedInstances.FirstOrDefault(dh => dialogIdentifier == null || Equals(dh.Identifier, dialogIdentifier));
```
bombipappoo, line 195: Obtaining `DialogHost` from `dialogIdentifier` is found in `Show()`, so it's better to move the duplicated code to a new method.

MaterialDesignThemes.Wpf/DialogHost.cs
```diff
@@ -181,6 +181,22 @@ public static async Task<object> Show(object content, object dialogIdentifier, D
...
+            if (targets != null)
+                CloseDialogCommand.Execute(false, null);
```
bombipappoo, line 197: You can call `Close()` instead of `CloseDialogCommand.Execute()`. Note that need to check whether `CurrentSession` is not null.

bombipappoo (commented):
I noted some thoughts.

MaterialDesignThemes.Wpf/DialogHost.cs
HenJigg: That's a big mistake.

PR #2051, opened 2020-09-06
now avoiding some missing resource warnings #2040

### LINKED ISSUE #2040: Missing DynamicResource lookups when SnackbarMessage removed from Snackbar
The [Super Quick Start](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/wiki/Super-Quick-Start) documentation says to put an instance of the `BundledTheme` in the collection of merged dictionary of `Application`.  However, I put it in the collection of merged dictionaries for my `MainWindow`, which contains a `Snackbar`.  When one of its messages disappears, WPF says there are some missing resource dictionary keys.

# Steps to reproduce
1. Fork [this repo](https://github.com/bender2k14/MDinXaml_DynamicResources_from_App)
2. Run the application
3. Wait for the snackbar message to disappear

# Expected behavior

No new lines appear in the "Output" window of Visual Studio.

# Actual behavior

The following lines appear in the "Output" window of Visual Studio.

```
System.Windows.ResourceDictionary Warning: 9 : Resource not found; ResourceKey='SecondaryAccentBrush'
System.Windows.ResourceDictionary Warning: 9 : Resource not found; ResourceKey='MaterialDesignSnackbarRipple'
System.Windows.ResourceDictionary Warning: 9 : Resource not found; ResourceKey='SecondaryAccentBrush'
```

# Workaround

1. Directly put an instance of the `BundledTheme` in the collection of merged dictionary of `Application` by commenting in the commented out line in `App.xaml`.
2. Indirectly make `System.Windows.Application.Current` have all the same resources as `MainWindow` be commenting in the commented out line in `MainWindow.xaml.cs`.

# Questions

I don't understand why these two situations have different behavior.  I expected both of them to have no missing resources.  

The part that is most confusing to me is that some resources were found.  I don't know an easy way to get the exact key, but notice that the button on the message has the text `UNDO` in the color `Lime`, which is the secondary color.

Why is the color `Lime` found by the `Snackbar` but not the color with resource key `SecondaryAccentBrush`?  Why is the resource with key `SecondaryAccentBrush` found when it is part of the `Resources` of `Application` but not when it was only part of the `Resources` of `MainWindow`, which is the parent of the `Snackbar`?

TysonMN:
Fixes #2040

This PR is based on the fix @bombipappoo suggested in https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2040#issuecomment-683248221.

To test that it works, try [this branch](https://github.com/bender2k14/MaterialDesignInXamlToolkit/tree/testing/2040_default_SnackbarMessage_resources), which adds an executable project in the first commit and then cherry picks the commit from the branch for this PR to create its second commit.  In this way, testing on the first commit exhibits the issue while testing on the second commit shows that the issue is fixed.

As I said in https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2040#issuecomment-687867884, I tried and failed to create an automated test for this issue.  I created this PR because I ran out of ideas for creating such an automated test.  If someone wants to suggest any more though, I will definitely look into them.

### DIFF
--- a/MaterialDesignThemes.Wpf/Snackbar.cs
+++ b/MaterialDesignThemes.Wpf/Snackbar.cs
@@ -31,16 +31,7 @@ namespace MaterialDesignThemes.Wpf
         }
 
         public static readonly DependencyProperty MessageProperty = DependencyProperty.Register(
-            nameof(Message), typeof(SnackbarMessage), typeof(Snackbar), new PropertyMetadata(default(SnackbarMessage), MessagePropertyChangedCallback));
-
-        private static void MessagePropertyChangedCallback(DependencyObject dependencyObject, DependencyPropertyChangedEventArgs dependencyPropertyChangedEventArgs)
-        {
-            var snackbar = (Snackbar)dependencyObject;
-            //this prevents missing resource warnings after the message is removed from the Snackbar
-            //see https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2040
-            if (snackbar.Message != null)
-                snackbar.Message.Resources = SnackbarMessage.defaultResources;
-        }
+            nameof(Message), typeof(SnackbarMessage), typeof(Snackbar), new PropertyMetadata(default(SnackbarMessage)));
 
         public SnackbarMessage Message
         {
--- a/MaterialDesignThemes.Wpf/Snackbar.cs
+++ b/MaterialDesignThemes.Wpf/Snackbar.cs
@@ -31,7 +31,16 @@ namespace MaterialDesignThemes.Wpf
         }
 
         public static readonly DependencyProperty MessageProperty = DependencyProperty.Register(
-            nameof(Message), typeof(SnackbarMessage), typeof(Snackbar), new PropertyMetadata(default(SnackbarMessage)));
+            nameof(Message), typeof(SnackbarMessage), typeof(Snackbar), new PropertyMetadata(default(SnackbarMessage), MessagePropertyChangedCallback));
+
+        private static void MessagePropertyChangedCallback(DependencyObject dependencyObject, DependencyPropertyChangedEventArgs dependencyPropertyChangedEventArgs)
+        {
+            var snackbar = (Snackbar)dependencyObject;
+            //this prevents missing resource warnings after the message is removed from the Snackbar
+            //see https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2040
+            if (snackbar.Message != null)
+                snackbar.Message.Resources = SnackbarMessage.defaultResources;
+        }
 
         public SnackbarMessage Message
         {
--- a/MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs
+++ b/MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs
@@ -352,6 +352,11 @@ namespace MaterialDesignThemes.Wpf
             //this for now...at least it is prevent extra call back hell
             await Task.Delay(snackbar.DeactivateStoryboardDuration);
 
+            //this prevents missing resource warnings after the message is removed from the Snackbar
+            //see https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2040
+            if (snackbar.Message != null)
+                snackbar.Message.Resources = SnackbarMessage.defaultResources;
+
             //remove message on snackbar
             snackbar.SetCurrentValue(Snackbar.MessageProperty, null);
 
--- a/MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs
+++ b/MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs
@@ -338,7 +338,10 @@ namespace MaterialDesignThemes.Wpf
         private async Task ShowAsync(Snackbar snackbar, SnackbarMessageQueueItem messageQueueItem, ManualResetEvent actionClickWaitHandle)
         {
             //create and show the message, setting up all the handles we need to wait on
-            var mouseNotOverManagedWaitHandle = CreateAndShowMessage(snackbar, messageQueueItem, actionClickWaitHandle);
+            var tuple = CreateAndShowMessage(snackbar, messageQueueItem, actionClickWaitHandle);
+            var snackbarMessage = tuple.Item1;
+            var mouseNotOverManagedWaitHandle = tuple.Item2;
+
             var durationPassedWaitHandle = new ManualResetEvent(false);
             StartDuration(messageQueueItem.Duration.Add(snackbar.ActivateStoryboardDuration), durationPassedWaitHandle);
 
@@ -354,8 +357,7 @@ namespace MaterialDesignThemes.Wpf
 
             //this prevents missing resource warnings after the message is removed from the Snackbar
             //see https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2040
-            if (snackbar.Message != null)
-                snackbar.Message.Resources = SnackbarMessage.defaultResources;
+            snackbarMessage.Resources = SnackbarMessage.defaultResources;
 
             //remove message on snackbar
             snackbar.SetCurrentValue(Snackbar.MessageProperty, null);
@@ -364,7 +366,7 @@ namespace MaterialDesignThemes.Wpf
             durationPassedWaitHandle.Dispose();
         }
 
-        private static MouseNotOverManagedWaitHandle CreateAndShowMessage(UIElement snackbar,
+        private static Tuple<SnackbarMessage, MouseNotOverManagedWaitHandle> CreateAndShowMessage(UIElement snackbar,
             SnackbarMessageQueueItem messageQueueItem, EventWaitHandle actionClickWaitHandle)
         {
             var clickCount = 0;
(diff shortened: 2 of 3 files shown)


MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs
```diff
@@ -352,6 +352,11 @@ private async Task ShowAsync(Snackbar snackbar, SnackbarMessageQueueItem message
             //this for now...at least it is prevent extra call back hell
             await Task.Delay(snackbar.DeactivateStoryboardDuration);
 
+            //this prevents missing resource warnings after the message is removed from the Snackbar
+            //see https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2040
+            if (snackbar.Message != null)
+                snackbar.Message.Resources = SnackbarMessage.defaultResources;
```
bombipappoo, line 358: Sorry, but it should be done in `PropertyChangedCallback` of `Snackbar.MessageProperty` instead of here.

bender2k14: No problem.

I made that change in [this commit](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/pull/2051/commits/039d49c5d8cdff629dc498eb2d39b16bac69dd8a) and verified that this also works by adding this commit to my [testing branch](https://github.com/bender2k14/MaterialDesignInXamlToolkit/tree/testing/2040_default_SnackbarMessage_resources) and running the project that I added in that branch.

Since I think I made the requested change, I will mark this conversation as resolved for convenience.  Please reopen though if there is anything else to discuss.

MaterialDesignThemes.Wpf/Snackbar.cs
```diff
@@ -31,7 +31,16 @@ static Snackbar()
...
-            nameof(Message), typeof(SnackbarMessage), typeof(Snackbar), new PropertyMetadata(default(SnackbarMessage)));
+            nameof(Message), typeof(SnackbarMessage), typeof(Snackbar), new PropertyMetadata(default(SnackbarMessage), MessagePropertyChangedCallback));
+
+        private static void MessagePropertyChangedCallback(DependencyObject dependencyObject, DependencyPropertyChangedEventArgs dependencyPropertyChangedEventArgs)
+        {
+            var snackbar = (Snackbar)dependencyObject;
+            //this prevents missing resource warnings after the message is removed from the Snackbar
+            //see https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2040
+            if (snackbar.Message != null)
+                snackbar.Message.Resources = SnackbarMessage.defaultResources;
```
bombipappoo, line 42: "UNDO" has disappeared. It should be like this.

```csharp
            if (dependencyPropertyChangedEventArgs.OldValue is SnackbarMessage oldMessage)
                oldMessage.Resources = SnackbarMessage.defaultResources;
```

Since we can reuse `SnackbarMessage` like this, "UNDO" disappears with the current approach.

```csharp
      private async void Button1_Click (object sender, RoutedEventArgs e)
        {
            var message1 = new SnackbarMessage { Content = "Hello world1", ActionContent = "UNDO" };
            var message2 = new SnackbarMessage { Content = "Hello world2", ActionContent = "UNDO" };

            MySnackbar.Message = message1;
            MySnackbar.IsActive = true;
            await Task.Delay (1000);

            MySnackbar.Message = message2;
            await Task.Delay (1000);

            MySnackbar.Message = message1;
            await Task.Delay (1000);
            MySnackbar.IsActive = false;
        }
```

Therefore, this may be better.

Snackbar:
```csharp
        private static void MessagePropertyChangedCallback(DependencyObject dependencyObject, DependencyPropertyChangedEventArgs dependencyPropertyChangedEventArgs)
        {
            (dependencyPropertyChangedEventArgs.OldValue as SnackbarMessage)?.Clear();
            (dependencyPropertyChangedEventArgs.NewValue as SnackbarMessage)?.Restore();
        }
```

SnackbarMessage:
```csharp
        private ResourceDictionary _resources;
        public new ResourceDictionary Resources
        {
            get => _resources;
            set
            {
                _resources = value;
                if (base.Resources != defaultResources)
                    base.Resources = value;
            }
        }

        public SnackbarMessage() => _resources = base.Resources;
        internal void Clear() => base.Resources = defaultResources;
        internal void Restore() => base.Resources = _resources;
```

bender2k14: Yes, you are right.  Great catch.

In the past two days, I have been trying to find a better solution.  To that end, I thought it would be helpful if I could reproduce this "Resource not found" warning using a minimal example (i.e. one that doesn't depend on MDinXAML).  I have tried, but I haven't been able to figure out how to do that.

@bombipappoo, do you know how to do this?  Said another way...do you know how to create an example like I did in https://github.com/dotnet/wpf/issues/2956#issuecomment-681189922 but one that is actually minimal (...one that doesn't have any dependencies like MDinXAML)?

bombipappoo:
@bender2k14 Try [this](https://github.com/bombipappoo/ResourceNotFoundApp.git).

TysonMN:
The main issue I raised (which I am now articulating more precisely now than ever before) was that an instance of `SnackbarMessage` _created by MDinXAML_, added to the active spot in the `SnackbarMessageQueue`, and then removed from that post causes warnings about missing resource dictionary keys.

Your [core idea](https://github.com/bombipappoo/MaterialDesignInXamlToolkit/commit/333c3d66651c6975fb05ec5e4326daf1221e472d) (original linked to in https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2040#issuecomment-683248221) is to set `Snackbar.Resources` to some default resources at an appropriate time.  The downside with this solution is having to list all the default resource keys.  I have tried...and failed...to find a solution that doesn't have this downside. Nonetheless, I think this is good pragmatic solution.

Your original implementation approach was to set the default resources in [`SnackbarMessageQueue.ShowAsync(...)`](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/5eb607529053039dad791ed44c1dd8be8144a5de/MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs#L338).  Then [you suggested](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/pull/2051#discussion_r484504241) moving this to the `PropertyChangedCallback` of `Snackbar.MessageProperty` and then further suggested a more complicated solution that would also avoid the missing resource dictionary key warnings when the user "manually" puts their own `SnackbarMessage` into the active spot.

I have been thinking about all this.  I now like your original approach better.  If the user is going to "manually" remove a message from the active spot in the queue, then they can also set the default resources to avoid the missing resource dictionary key warnings.

I pushed some commits that revert the behavior in the branch for this PR to your original implementation idea.  I also modified it slightly to make it clearer that the instance of `SnackbarMessage` having its resources set to defaults is an instance created by MDinXAML (and will never be `null`, so we don't need a `null` check like before).

@bombipappoo, what do you think about this?

bombipappoo:
Certainly, it works correctly under normal use, so my first approach may suffice.

TysonMN:
@Keboo, I think this PR is ready for your review.  No rush though.

MaterialDesignThemes.Wpf/SnackbarMessageQueue.cs
```diff
@@ -381,7 +383,7 @@ private async Task ShowAsync(Snackbar snackbar, SnackbarMessageQueueItem message
             };
             snackbar.SetCurrentValue(Snackbar.MessageProperty, snackbarMessage);
             snackbar.SetCurrentValue(Snackbar.IsActiveProperty, true);
-            return new MouseNotOverManagedWaitHandle(snackbar);
+            return Tuple.Create(snackbarMessage, new MouseNotOverManagedWaitHandle(snackbar));
```
bombipappoo, line 386: Personally, I prefer `ValueTuple` and deconstructing instead of `Tuple`.

TysonMN: Me too, but when I tried to use it, Visual Studio said...
> Predefined type 'System.ValueTuple`2' is not defined or imported

...so I decided to make a more minimal change by sticking with `System.Tuple`.

2020-09-15_08-56-55_299_devenv

bombipappoo: Add this to MaterialDesignThemes.Wpf.csproj
```xml
  <ItemGroup Condition="$(TargetFramework) == 'net45'">
    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
  </ItemGroup>
```

TysonMN: Normally I would, but MDinXAML is using paket, and I don't know how to use paket.

bombipappoo: I see, I don't know either.

TysonMN: I think using `System.Tuple` is fine for this PR.  A future PR can switch to using `System.ValueTuple` if that is desired.

bombipappoo: I got how to use paket.

Add to paket.dependencies.
```
nuget System.ValueTuple 4.5.0 framework: net45
```
Add to MaterialDesignThemes.Wpf\paket.references,
```
System.ValueTuple
```
Then run this command.
```
.paket\paket.exe install
```

Keboo:
@TysonMN I am good with the solution, however it would be nice to avoid the additional value tuple dependency. Can we simply change the return type to some POCO, or add an out parameter. Given that `CreateAndShowMessage` is private I am not too concerned if its API is a bit ugly.

TysonMN:
Sure.  What if the return type is the older tuple `System.Tuple<,>`?

TysonMN:
I removed the commit that switched from `System.Tuple<,>` to `System.ValueTuple<,>`, so no dependency is added now.

Keboo (approved):

Keboo:
Great! Thank you for your work on this.

PR #2053, opened 2020-09-09
Fix blurry text in context menus

jairbubbles:
When using High DPI screen with text scaling at 125% the text in context menu is blurry. The culprit is the effect that is used to have a drop shadow.

This is discussed on many places : https://stackoverflow.com/questions/805677/why-everything-in-wpf-is-blurry

- Adding `UseLayoutRounding=true`  did fix the sub menus
- For the root menus I did not find any other solution than using a trick. A grid is used with two child borders on top of each other, one is only for the drop shadow and does not affect the other border.

Before:  
image

After:  
image


Keboo (approved):

Keboo:
Thanks! Fix looks great!

Issue #2081, opened 2020-10-01
Move TransitionAssist to new root namespace


Keboo:
The [TransistionAssist](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MaterialDesignThemes.Wpf/Transitions/TransitionAssist.cs) class lives within the `MaterialDesignThemes.Wpf.Transitions`. This file should be moved up a directory (to be inside of the MaterialDesignThemes.Wpf directory), and its namespace changed to be `MaterialDesignThemes.Wpf`.

This will also require updating the relevant [demo app](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/tree/master/MainDemo.Wpf/TransitionsDemo) slides to reference to new namespace.

PR #2089, opened 2020-10-01
Moved TransitionAssist up one level

theramis:
Fixes https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2081

I didn't need to update the path in the demo app here as suggested in the issue https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MainDemo.Wpf/Fields.xaml#L66 as it seemed to just work.


Keboo:
Thanks but this was previously addressed.

Issue #2083, opened 2020-10-01
Rename ColorHelper to ColorAssist


Keboo:
By default all of the extension method classes in this library end with the "Assist" suffix. We should update [ColorHelper](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MaterialDesignColors.Wpf/ColorManipulation/ColorHelper.cs) to be named `ColorAssist`.

4Rebin:
Hey Keboo,
I like to work with this issue.

PR #2091, opened 2020-10-01
Rename ColorHelper to ColorAssist #2083

alexeykap:


Keboo:
Thanks but this has already been addressed.

Issue #2082, opened 2020-10-01
Rename validation theme resources


Keboo:
Most of the theme resources are prefixed with "MaterialDesign" exception for `ValidationErrorBrush` and `ValidationErrorColor`. We should add this prefix to all usages of both the brush and the color.

[Usages of ValidationErrorBrush](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/search?q=ValidationErrorBrush)
[Usages of ValidationErrorColor](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/search?q=ValidationErrorColor)

trevorcarter83:
@Keboo I can take this if you like.

Keboo:
@trevorcarter83 thanks but it looks like there is already a PR for it

PR #2092, opened 2020-10-01
Rename validation theme resources #2082

alexeykap:


Keboo:
Thanks but this was already addressed.

PR #2096, opened 2020-10-01
Rename ColorHelper to ColorAssist in ColorManipulation

4Rebin:
Hey there,
The ColorHelper class has renamed to ColorAssist in ColorManipulation folder.


Keboo:
Thanks but this was previously addressed

Issue #2076, opened 2020-10-01
Renamed filled TextBox style


Keboo:
The style `MaterialDesignFilledTextFieldTextBox` should be renamed to `MaterialDesignFilledTextBox`

The style can be found [here](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.TextBox.xaml)

Update all usages in the demo app to use the new style name as well.

abuZayed15:
Can I give this issue a try? It's my first time contributing to an open source project

Keboo:
@abuZayed15 yes go for it. If you have questions feel free to post them here, or on the [gitter chat](https://gitter.im/ButchersBoy/MaterialDesignInXamlToolkit)

abuZayed15:
Great. I will take it. Is there documentation that would help me get the  MaterialDesignInXAMLToolkit up running in an IDE, like Visual Studio? Thank you for letting me have this issue

PR #2101, opened 2020-10-02
Renamed filled TextBox style #2076

mohammed-ijas:


Keboo:
Thank you but this has already been addressed

PR #2122, opened 2020-10-11
Use paket as dotnet tool, fix #2118

### LINKED ISSUE #2118: Update paket to use the dotnet version
We are currently still using the old [Paket bootstrapper](https://fsprojects.github.io/Paket/get-started.html) we should remove this and instead update to using paket as a dotnet local tool. (docs [here](https://fsprojects.github.io/Paket/get-started.html)).

For this I would expect to see:
- [ ] the existing .paket directory removed. 
- [ ] A new dotnet local tool manifest created for paket

menishmueli:
fix #2118

- Removed .paket folder exe related files that is now a dotnet tool
- Also removed .Nuget folder, I did not see any usages of the nuget tool directly (the scripts using the dotnet cli instead)
- Updated azure-pipelines.yml to do restore tools and paket restore

Misc:
- .paket folder could not be removed completely because of the Paket.Restore.targets file that is needed when building because of all the .csproj file referances it
- Removed dotnet restore from azure-pipelines.yml, it is redundent becuase of the "dotnet paket restore" command
- In MahMaterialDragablzMashUp/MahAppsDragablzDemo.csproj there is a package referance to Dragablz, I'm not sure why but it was added recently. I think it should be removed (maybe in another issue?)
- I did not find documentation for dev environment startup. because of the usage of tools every developer need to run a "dotnet tool restore" command to restore paket (See https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools)
If the dev environment startup do not exist it should be created and contain the "dotnet tool restore" instruction.


Keboo (approved):
Great work, thank you

PR #2123, opened 2020-10-11
Updated ShowMeTheXAML to version 2.0.0.0 and fixed breaking changes, fix #2119

### LINKED ISSUE #2119: Update ShowMeTheXAML to version 2.0.0
ShowMeTheXAML just released a new 2.0.0 with some breaking changes.
The key differences is the `Key` properties on the XamlDisplay class was renamed to `UniqueKey`. We should update the version we are using to 2.0.0 and update all of the usages to using the new property. 

These dependencies are managed with [Paket](https://fsprojects.github.io/Paket/learn-how-to-use-paket.html) not NuGet directly so you will need to use paket to update and restore the dependency. Keep in mind there is a separate issue for updating paket (#2118) so the exact command needed will depend on if that work has been done.

menishmueli:
fix #2119

- Updated ShowMeTheXAML to version 2.0.0.0
- Fix breaking changes:
- Renamed "Key" property to "UniqueKey"

I runned the Demo app and the ShowMeTheXAML fearure is working as expected


jespersh:
Looks good to me

PR #2141, opened 2020-10-20
Update README.md

payal006:


Keboo:
Thanks but I don't believe we want this verbiage change.

PR #2164, opened 2020-11-08
Allow changing expander header background through ExpanderAssist

ncksol:
I needed this functionality for myself, ExpanderAssist seemed like the right place to implement it with

### DIFF
--- a/MaterialDesignThemes.Wpf/ExpanderAssist.cs
+++ b/MaterialDesignThemes.Wpf/ExpanderAssist.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Windows;
 using System.Windows.Controls;
+using System.Windows.Media;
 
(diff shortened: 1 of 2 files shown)


jizc (changes requested):
I would change the property type to `Brush` instead of `SolidColorBrush`, to allow for more flexibility.

You have only changed the horizontal style (`MaterialDesignHorizontalHeaderStyle`). If this change is to be made, it needs to change the vertical style as well (`MaterialDesignVerticalHeaderStyle`).

However, a better solution would be to revert the change to the horizontal style, and instead set the `Background` property on the `ToggleButton` named `HeaderSite`. As both toggle button styles get their background from `TemplateBinding`.

ncksol:
@jizc thanks for the feedback, that does make sense. I'll update the PR.

Keboo (approved):
Great work! Thank you

PR #2176, opened 2020-11-20
WIP: First round of adding a new project template

Keboo:
First round of changes to include a new project template to make getting started easier.


Templates/Project/MainWindow.xaml
```diff
@@ -0,0 +1,23 @@
...
+        xmlns:local="clr-namespace:mdix"
+        xmlns:materialDesign="http://materialdesigninxaml.net/winfx/xaml/themes"
+        mc:Ignorable="d"
+        TextElement.Foreground="{DynamicResource MaterialDesignBody}"
+        TextElement.FontWeight="Regular"
+        TextElement.FontSize="13"
+        TextOptions.TextFormattingMode="Ideal" 
+        TextOptions.TextRenderingMode="Auto"
+        Background="{DynamicResource MaterialDesignPaper}"
+        FontFamily="{materialDesign:MaterialDesignFont}"
```
campersau, line 15: Mabye use the new `Style="{StaticResource MaterialDesignWindow}"` from https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/pull/2174 instead?

Keboo: I was considering that, but it would have to be conditional based on the NuGet package version. But probably still work doing.

Templates/Project/App.xaml
```diff
@@ -1,4 +1,4 @@
-﻿<Application x:Class="mdix.App"
+﻿<Application x:Class="MaterailDesign.WpfApp.App"
```
MichelMichels, line 1: Small type in namespace, "MaterailDesign"-> "MaterialDesign"

Templates/Project/MainWindow.xaml.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace MaterailDesign.WpfApp
```
MichelMichels, line 1: Namespace typo

Keboo:
Scaping this in favor of a more generic approach that will be handled here: https://github.com/Keboo/DotnetTemplates

PR #2181, opened 2020-11-27
Update MaterialDesignTheme.TabControl.xaml

stanley8615:
Fix when TabStripPlacement = Top / Bottom , tabItem in headerPanel  not Horizontal


HClausing:
Actually, when I created this style, I was thinking only on [NavigationRail ](https://material.io/components/navigation-rail/) component, that don't have horizontal placement. There's a WIP style for TabControl (commented on the same XAML resource dictionary).

PR #2201, opened 2020-12-29
Added vertical offset to properly align submenu

### LINKED ISSUE #2190: Submenus (MenuItem) are not properly aligned
Hello,
As you can see on my screenshot, submenus (MenuItem) are not properly aligned. 
I don’t know if it’s intentional, but I think it's an issue.

image

jizc:
Fixes #2190


Keboo (approved):

josh2112:
This has the side-effect of pushing the top-level menus (those attached to a `<Menu>` up by 16 pixels as well:

image

Which just looks weird. Is there an easy solution to this? Perhaps use a control trigger to set the vertical alignment back to 0 if the popup's placement target is a Menu instead of a MenuItem?

josh2112:
Fixed! Instead of setting the `VerticalAlignment` on the `Popup` itself, set it in the first `ControlTempate` `Trigger` (the one for a `Role` of `SubmenuItem`.

PR #2204, opened 2021-01-03
Outlined ComboBox style

### LINKED ISSUE #1613: Outline Combobox?
So I have decided to go the route using outlined form fields. As of today this library only has Outlined/Filled features for Textboxes, PasswordBoxes and DatePickers. Is there anyway we could get one for ComboBoxes? I'm trying to do it custom myself but I'm just not quite advanced enough in XAML to quite figure it out :/. As of right now I just set BorderThickness to 2 and added some padding as well. It looks a little awkward with the Pretty outlined controls that already have the feature

jizc:
Fixes #1613

This PR adds a new `ComboBox` style with the key `MaterialDesignOutlinedComboBox`.

image

### Spec
https://material.io/components/menus#exposed-dropdown-menu

### Notes
- This combines the two existing `ComboBox` templates and makes use of attach properties instead, so the same template could also be used for the outlined style. I used the `MaterialDesignTextBoxBase` style as a guide for this.
- In the spec, when the `ComboBox` has no selection, but the drop down menu is open, the hint/label is supposed to be moved up. I tested with this by changing `ComboBoxHintProxy`, but it caused issues with the regular `ComboBox` style.


Keboo (approved):
Really great work! Thank you

Keboo:
Looks like the demo app clean up caused a conflict. Do you mind rebasing this onto master?

jizc:
On it! 😄

jizc:
Not that experienced with rebasing, so I opted to merge master into the feature branch instead. You are welcome to squash the PR if you prefer to keep things a little tidier.

PR #2242, opened 2021-02-12
Implementing a character counter when MaxLength is set

Keboo:
This replaces #1607


om226:
How do I hide this "character counter"

Keboo:
@om226 see #2247 for details

PR #2244, opened 2021-02-14
Slider spec update

### LINKED ISSUE #1137: Slider visualizes selection range always in slider center
I tried to use a slider with the standard feature IsSelectionRangeEnabled="true". As long as I am not using material design package the selection range will be shown properly, but when I use material design the selection range color visualization is only in the middle of the slider, although the range ticks are shown correctly.

See following images
defaultslider
slidermaterialdesign

 
How to reproduce. Use the following code in a view and use material design package.
 
```
         <Slider
            x:Name="PowerSlider"
            Margin="8"
            DockPanel.Dock="Left"
            IsSelectionRangeEnabled="True"
            Maximum="28.8"
            Minimum="0"
            SelectionEnd="22.3"
            SelectionStart="12.5"
            TickPlacement="BottomRight"/>
```

### LINKED ISSUE #1782: Update discrete slider appearance to match latest material design specification
Currently our discrete slider style does not match the [latest material design spec](https://material.io/components/sliders#discrete-slider).
image

Specifically the popup that is shown as the value changes should be updated to match the new style.
This requires updating the existing `MaterialDesign*DiscreteSliderThumb` templates.

jizc:
Fixes #1782
Fixes #1137

### Summary
This PR updates all the slider styles to the current MD spec. It also fixes the selection range position and style.

As I had to use the demo app a lot during this work, I couldn't stop myself from improving it a bit. The main changes are the addition of `ViewModelBase` and that the demo pages are now created on-demand, instead of all at once during start-up. The latter vastly improves the start-up time of the demo. The sliders demo now has a lot more examples.

### MD Spec
[Sliders - Material Design](https://material.io/components/sliders)

### Behavior notes
To enable the "Click and arrow" and the "Tab and arrow" behaviors in the spec, part of the style was changed to trigger on `IsFocused` instead of `IsDragging`. This results in having to click away the focus indicator, which may be a little jarring for some, but this is the correct behavior, that follows the MD spec. This vastly improves the tab experience, as the control now has a proper visual focus indicator.

### Selection range notes
The WPF slider selection range feature is not covered in the MD spec, so I had to improvise a bit, but I'm happy with how it turned out. The feature is easily confused with range selection where you have 2 grips, which is covered by the spec, but the WPF feature is very different.

### Breaking changes
- Horizontal track height increased from 18px to 20px
- Vertical track width increased from 18px to 20px
- Discrete slider styles no longer include extra whitespace for the label popup, they are the same size as continuous versions. This results in a height/width decrease from 60px to 20px
- `IsMoveToPointEnabled` is `true` by default. This is to enable the "Click jump" behavior in the spec
- `RepeatButton` styles are replaced by a single style; `MaterialDesignRepeatButton`. Removed styles: `MaterialDesignHorizontalTrackRepeatButton`, `MaterialDesignHorizontalTrackDiscreteRepeatButton`, `MaterialDesignVerticalTrackRepeatButton`, `MaterialDesignVerticalTrackDiscreteRepeatButton`, `MaterialDesignHorizontalColorSliderTrackRepeatButton`

### Screenshots
screen-sliders

image image


Keboo:
This is really great work! 
Initially I thought the discrete slides popup staying visible was a bug, but as you indicated this matches the spec. I am going to merge this, but I think it might be nice to create an attached property that allows for only showing the numeric value label only while dragging. That way is people want something closer to the original behavior there is a way to get something similar. Thoughts?

jizc:
I had the same thought about an attach property when working on this, so let's do it 😃 I'll submit a PR for it. Do you have a suggestion on the name of the property?

PR #2288, opened 2021-03-23
Don't disable primary content when dialog visible to workaround #2285

### LINKED ISSUE #2285: Missing DialogHostContent when opening DialogHost for the second time in a large application
If the XAML contains a sufficiently large number of elements with certain types, then the content of a `DialogHost` is not displayed when opening the `DialogHost` for the second time (as well as all times after that).  An additional constraint is that the `DialogHostContent` must bind its `Visibility` and have the value match the `IsOpen` property of the `DialogHost` (i.e. `true` means `Visible` and `false` means `Collapsed`).

Although this incorrect behavior involves `DialogHost`, my guess is that neither `DialogHost` nor any other part of Material Design in XAML Toolkit is to blame.  This feels like a bug in WPF to me.  I did not try to create a reproduction that does not involve `DialogHost`, but that is an interesting direction to pursue.  See the section title "Additional information" below.

## Steps to reproduce
1. Create a local fork of [TysonMN/MissingDialogHostContent](https://github.com/TysonMN/MissingDialogHostContent)
2. Checkout one of the following branches: [`CheckBox`](https://github.com/TysonMN/MissingDialogHostContent/tree/CheckBox), [`ComboBox`](https://github.com/TysonMN/MissingDialogHostContent/tree/ComboBox), [`RadioButton`](https://github.com/TysonMN/MissingDialogHostContent/tree/RadioButton), or [`TextBox`](https://github.com/TysonMN/MissingDialogHostContent/tree/TextBox)
3. Run the application
4. Click the button (that says `Open Dialog`)
5. Observe the text in the dialog that says `Hello World`
6. Click on the dimmed background (but not too close in order to avoid bug #2279)
7. Click the button again

## Expected behavior
The text `Hello World` should be visible (again, just like in step 5).  It should look like this animated GIF, which I created using the branch [`main`](https://github.com/TysonMN/MissingDialogHostContent/tree/main).

## Actual behavior
The text `Hello World` is not visible.  The whole screen is dimmed.  This behavior continues each time the dialog is closed and reopened.  This animated GIF shows the actual behavior when using the branch `CheckBox`.

## Additional information
To help quantify how large is "sufficiently large", my reproduction repository contains several branches named after a type that extends `FrameworkElement`.  Each branch forks from `main` and has two commits of its own.  In the first comment, many of the corresponding element are added to the application.  On these first commits, following the reproduction steps leads to the "expected behavior".  In the second commit, an additional element of the same type is added.  On these second commits, following the reproduction steps leads to the "actual behavior".  For each element type, here are the quantities that give the expected and actual behaviors.

| Element Type | Expected Behavior | Actual Behavior |
| --------------- | -------------------- | ----------------- |
| [`CheckBox`](https://github.com/TysonMN/MissingDialogHostContent/tree/CheckBox) | 144 | 145 |
| [`ComboBox`](https://github.com/TysonMN/MissingDialogHostContent/tree/ComboBox) | 48 | 49 |
| [`RadioButton`](https://github.com/TysonMN/MissingDialogHostContent/tree/RadioButton) | 144 | 145 |
| [`TextBox`](https://github.com/TysonMN/MissingDialogHostContent/tree/TextBox) | 72 | 73 |

I tried reproducing the actual behavior using other elements types, but the other cases were more confusing.  I have a private reproduction in my application at work that involves `Button`s with its `Style` set to some `StaticResource` key, but I was unable to achieve the actual behavior via `Button`s without modifying their `Style`.

## Workaround

I have found one workaround.  See the branch [workaround_GridWithHeightWidth](https://github.com/TysonMN/MissingDialogHostContent/tree/workaround_GridWithHeightWidth).  It wraps the `DialogHostContent` in a `Grid` that has a fixed `Height` and `Width`.

TysonMN:
Resolves #2285 

I continue to believe that issue #2285 is not due to a bug in the code of Material Design in XAML.  Instead, I think it is a bug in the code of WPF.  See https://github.com/dotnet/wpf/issues/4319.

Nonetheless, it is reasonable to consider working around the deficiencies of dependencies.  This PR makes such a workaround, which was identified in https://github.com/dotnet/wpf/issues/4319 (see the third workaround).  Specifically, the change is to keep the primary content enabled (i.e. don't set `IsEnabled` to `false` for the element with name `ContentPresenter`).  This is a reasonable change because the `Grid` with name `PART_ContentCoverGrid` already obscures the view of the element with name `ContentPresenter` and intercepts all user interaction.

On the other hand, I also think it is reasonable to leave the code as is since this bug probably only affects a small number of users.  Instead, those affected can wait for a fix in WPF.  While this bug does affect me, I have already created custom NuGet packages of `MaterialDesignThemes` and `MaterialDesignColors` that includes this change, so I am not blocked in any way by this PR.

If the decision is made to merge this PR, it might also be a good idea to comment out that line instead of deleting it and add a comment nearby that links to #2285.

### DIFF
--- a/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.DialogHost.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.DialogHost.xaml
@@ -189,7 +189,6 @@
                     </Grid>
                     <ControlTemplate.Triggers>
                         <Trigger Property="IsOpen" Value="True">
-                            <Setter TargetName="ContentPresenter" Property="IsEnabled" Value="False" />
                             <Setter TargetName="PART_ContentCoverGrid" Property="IsHitTestVisible" Value="True" />
                         </Trigger>
                     </ControlTemplate.Triggers>
@@ -377,7 +376,6 @@
                     </Grid>
                     <ControlTemplate.Triggers>
                         <Trigger Property="IsOpen" Value="True">
-                            <Setter TargetName="ContentPresenter" Property="IsEnabled" Value="False" />
                             <Setter TargetName="PART_ContentCoverGrid" Property="IsHitTestVisible" Value="True" />
                         </Trigger>
                     </ControlTemplate.Triggers>


Keboo:
Thank you for the write up and time spent on this. I want to spend some time reviewing the implications of this change, but I do agree that it would be nice to work around.

Keboo:
@TysonMN ok, I think this is a reasonable change. However, it does cause a slight UI difference with the controls no longer longer being disabled. For that reason can you please update so the Embedded dialog host has the same change, so both visually look the same.

TysonMN:
Done

Keboo:
Looks like the latest commit, actually undoes the first one. I was proposing that on both the embedded and default dialog hosts, we remove the IsEnabled trigger for the content presenter.

TysonMN:
Woops!  That is definitely not what I meant to do.  Sorry about that.

I just force pushed a single comment that removes boths lines.

Keboo (approved):

PR #2290, opened 2021-03-26
Clicking on dialog shadow now dismisses dialog

### LINKED ISSUE #2279: DialogHost CloseOnClickAway region should include more space next to the dialog
The `MaterialDesignDemo` project contains five `DialogHost` samples.  The fourth one supports closing when "clicking away", which means clicking on the dimmed background.  However, not all locations cause the `DialogHost` to close when clicked.  The set of such locations might even be inconsistent across executions though it does seem consistent during a particular execution.

This is not blocking me in any way.  Not only am I not using this feature in my application at work but it even worked correctly in that application after I added it out of curiosity.

Keboo:
Fixes #2279

### DIFF
--- a/MaterialDesignThemes.Wpf/Converters/ShadowConverter.cs
+++ b/MaterialDesignThemes.Wpf/Converters/ShadowConverter.cs
@@ -7,7 +7,7 @@ namespace MaterialDesignThemes.Wpf.Converters
 {
     public class ShadowConverter : IValueConverter
     {
-        public static readonly ShadowConverter Instance = new ShadowConverter();
+        public static readonly ShadowConverter Instance = new();
 
         public object? Convert(object? value, Type targetType, object? parameter, CultureInfo culture)
         {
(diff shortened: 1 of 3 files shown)


TysonMN:
Good solution.

I expect this to only work for left mouse clicks.  What about a touch event?
https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/020cea36821fb2f998a9992a5a1b7b6d6919f61a/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.DialogHost.xaml#L145-L154

TysonMN:
And should this same change be made for the style with key `MaterialDesignEmbeddedDialogHost`?

Keboo:
I will test it on my surface later today when i get some time, but I don't expect it to behave any different. I figured basing it on the mouse event event would bring it inline with what the [DialogHost control is doing for the overlay](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MaterialDesignThemes.Wpf/DialogHost.cs#L504). In many cases touch events get raised as mouse events, but it is a good call out, and I will test it.

As for the embedded dialog host. No, it is not needed since the dialog is within the visual tree, while the other is in a Popup which has its own tree. From my testing it appeared that the dialog's margin was forcing the popup larger, and the popup was catching and preventing the click from reaching the window below it. This is not an issue when the dialog content and the dialog  host are in the same visual tree.

PR #2301, opened 2021-04-13
Fix submenus of top-level MenuItems pushed up by 16 pixels

josh2112:
[Pull request 2201](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/pull/2201) has the side-effect of pushing up submenus of a menu bar:

image

That pull request was intended only to fix submenu alignment when a submenu opened to the right of another submenu by raising the popup 16 pixels, but by applying the `VerticalOffset` directly to the popup it changed it everywhere.

This pull request applies the `VerticalOffset` via a control trigger only to submenu headers (menu items that are not part of a `Menu`.


Keboo (approved):
Awesome! Thank you for the contribution

PR #2331, opened 2021-05-25
Proposed Fix for 2330, ProgressBar clipping inside the Button

### LINKED ISSUE #2330: Button's ProgressBar is not clipped and the ProgressBar Border is stuck to PrimaryHueLightBrush
p8Oi0hU9fk

As you can see the progressbar spills outside the Button's Border and has a weird border color, which is inhereted, and therefore hardcoded, by the Progressbar control, where it is set by default to PrimaryHueLightBrush

In my opinion the ProgressBar Border should follow the color of "materialDesign:ButtonProgressAssist.IndicatorBackground"

StefanoRivolta-Previero:
Fixes #2330 

Clips the ProgressBar the same way the Ripple effect gets clipped

BorderBrush of the ProgressBar  now uses IndicatorBackground property instead of using PrimaryHueLightBrush

end result:

V23MMFvOwt


Keboo (approved):
Great work, thank you!

PR #2333, opened 2021-05-25
Proposal: add CornerRadius properties to the DrawerHost

### LINKED ISSUE #2332: Update DrawerHost to accept a CornerRadius as per Spec
https://material.io/design/shape/about-shape.html#shape-customization-tool

Material Spec says that drawers could have a radius, like this:

firefox_rR0vAvTQSY

StefanoRivolta-Previero:
fixes #2332 

Result:

o2o1NHPer4
0NnD4N5Mzh
YOaLU7IclC

It just adds four properties for the CornerRadius of each Drawer. To make it appear like in the Spec and pictures you'll need to specify each corner radius in the property, like this:

`        <materialDesign:DrawerHost  LeftDrawerCornerRadius="0 20 20 0" BottomDrawerCornerRadius="25 25 0 0">`

Maybe it could be improved to accept only one number and automatically figure out which two corners to smooth, although that would remove some customization. Not sure what's the right way

### DIFF
--- a/MaterialDesignThemes.Wpf/DrawerHost.cs
+++ b/MaterialDesignThemes.Wpf/DrawerHost.cs
@@ -172,6 +172,15 @@ namespace MaterialDesignThemes.Wpf
             get => (bool)GetValue(TopDrawerCloseOnClickAwayProperty);
             set => SetValue(TopDrawerCloseOnClickAwayProperty, value);
         }
+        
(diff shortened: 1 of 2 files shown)


Keboo:
I think the CornerRadius is the right call here. Nice addition!

Keboo (approved):

PR #2356, opened 2021-06-22
Update Calendar to specs for more customization

StefanoRivolta-Previero:
I've added a few properties to allow for more customization of the calendar control and to better match specs. Final result:

immagine

### Changes:

- Foreground property now changes the foreground of all text in the bottom half of the control (instead of some parts being hardcoded to MaterialDesignBody)
- Added properties to change the color of the header (Background and Foreground)
- Added properties to change the color of the selection (Background and Foreground)
- Added new style that removes the header Color, which looks more similar to what Google themselves use
- The button for month and year is now pill shaped, [to match the specs](https://material.io/components/date-pickers#mobile-pickers)
- Added property to change the orientation to horizontal, [like in the specs](https://lh3.googleusercontent.com/khzlTy4nEbmuJojqCZ68uLySrO-B9ZqSoik3q8NnSux32WS5pT2GtxmEaqdaHAraSlOq3zNlHa7GxLhvyINlGgzTQQKaBawseAdwQws=w1064-v0)
- Added possibility to control the Shadow directly from the Calendar

### DIFF
--- a/MaterialDesignThemes.Wpf/CalendarAssist.cs
+++ b/MaterialDesignThemes.Wpf/CalendarAssist.cs
@@ -1,13 +1,62 @@
 using System.Windows;
+using System.Windows.Media;
 
 namespace MaterialDesignThemes.Wpf
 {
+    public enum CalendarOrientation
+    {
+        Vertical,
+        Horizontal
+    }
     public static class CalendarAssist
     {
+        #region Header Visibility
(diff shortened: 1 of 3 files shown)


Keboo:
Really great work! If you want to add those samples from your screenshots to the demo app feel free. Always nice to have examples of what is possible.

PR #2359, opened 2021-06-30
Update Treeview to offer new options

StefanoRivolta-Previero:
# Result:
MqzA2EN6iK

## Features:

- Added a property to change the expander size in the TreeView (especially useful for touch oriented interfaces)
- Added a property to hide the selection effects, like the ripple and the selection and mouse over borders. Personally I find that if the selection state is not important, having selected items is quite distracting (and useless)


Keboo (approved):
Again, really great work!

PR #2367, opened 2021-07-07
Fix some functions with wrong parameters in SnackBarMessageQueue.cs

liufangchen:
微信图片_20210707171526

- Try to call Enqueue method in SnackbarMessageQueue，you will find that the value of promote parameter  is always false
-  Wrong function
```
public void Enqueue(object content, object? actionContent, Action? actionHandler, bool promote)
            => Enqueue(content, actionContent, _ => actionHandler?.Invoke(), promote, false, false);
```


Keboo (approved):
Great catch, thank you!

PR #2387, opened 2021-07-28
Add ShowSelection Property to ListBoxItem

StefanoRivolta-Previero:
Simple addition of a property to control the selection, click and hover highlights of ListBoxItem, similarly to the one added for TreeViewItem. Speaking of which, I've noticed an error in my last commit for TreeViewAssist, added the fix.

### DIFF
--- a/MainDemo.Wpf/Lists.xaml
+++ b/MainDemo.Wpf/Lists.xaml
@@ -128,7 +128,7 @@
 
             <smtx:XamlDisplay
                 UniqueKey="list_3"
-                Grid.Column="4"
+                Grid.Column="3"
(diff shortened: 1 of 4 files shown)


MainDemo.Wpf/Lists.xaml
```diff
@@ -107,7 +128,7 @@
 
             <smtx:XamlDisplay
                 UniqueKey="list_3"
-                Grid.Column="2"
+                Grid.Column="4"
```
campersau, line 131: Should this be `3` instead?

```suggestion
                Grid.Column="3"
```

Keboo (approved):

PR #2389, opened 2021-08-01
Add settings for color adjustment to demo application

### LINKED ISSUE #2361: Color adjustment in light mode with light colors doesn't work properly
When selecting light colors like yellow and teal in Light Mode, the colors become distorted like so:
RMXqEH4Urb

While, when I remove the `ColorAdjustment="{materialDesign:ColorAdjustment}"` from App.xaml, the colors work properly:
MaterialDesignDemo_2OPUzzu19Y

Is this an intended behavior?

FantasyTeddy:
This adds the possibility to enable/disable the automatic color correction in the demo application (resolves #2361).
Additionally it lets the user try out different values for the color adjustment.

Palette

### DIFF
--- a/MainDemo.Wpf/Domain/PaletteSelectorViewModel.cs
+++ b/MainDemo.Wpf/Domain/PaletteSelectorViewModel.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Windows.Input;
 using MaterialDesignColors;
 using MaterialDesignThemes.Wpf;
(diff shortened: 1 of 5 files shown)


Keboo:
This is great work. Thanks you!

FantasyTeddy:
Thank you for fixing the tests, I forgot to run those.
The setting is currently only available under the "Palette" page, should it also be made available for the "Color Tool" page? (Would probably require some refactoring to prevent a lot of copy&paste)

Keboo:
No problem, it was a good change and I was wanting to get it in.

It would be nice to also include it on the color tool page, not a big deal either way.

PR #2392, opened 2021-08-02
Use color adjustment settings on "Color Tool" page too

FantasyTeddy:
This is a continuation of #2389 and adds the possibility for color adjustment also on the "Color Tool" page.


Keboo (approved):
Awesome! Thank you for doing this.

Issue #2350, opened 2021-06-13
RangeSlider does not match latest slider spec update


Yoooi0:
In 4.1.0 the range slider style does not match the #2244 pr:
![2021-06-13_22-12-19](https://user-images.githubusercontent.com/63662834/121820851-36695c00-cc95-11eb-8ad5-12eaaa73fb9a.png)
Visual Studio also shows a bunch of style binding errors.

4.0.0:
![2021-06-13_22-19-20](https://user-images.githubusercontent.com/63662834/121820880-6add1800-cc95-11eb-84eb-a960ada68d41.png)

Keboo:
Hi @Yoooi0 can you provide the XAML used to generate the above screenshots?
I am also not able to replicate the VS style binding errors, are you able to provide those as well?

Thanks

Yoooi0:
XAML:
```xml
<Slider Style="{StaticResource MaterialDesignSlider}"/>
<metro:RangeSlider Style="{StaticResource MaterialDesignRangeSlider}"/>
```

I tested with compiled demo app from latest master commit.
I had to add a reference to `MaterialDesignThemes.MahApps` project, edit the `App.xaml` file as in wiki: https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/wiki/MahApps.Metro-integration, and add a RangeSlider style `<ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.MahApps;component/Themes/MaterialDesignTheme.MahApps.RangeSlider.xaml" />` 

Binding errors:
```
Severity	Count	Data Context	Binding Path	Target	Target Type	Description	File	Line	Project
Error	1	null	Orientation	MetroThumb.NoTarget, Name='PART_RightThumb'	Object	Cannot find source: RelativeSource FindAncestor, AncestorType='System.Windows.Controls.Primitives.Track', AncestorLevel='1'.			
Error	2	null	(0)	Ellipse.Effect, Name='grip'	Effect	Cannot find source: RelativeSource FindAncestor, AncestorType='System.Windows.Controls.Slider', AncestorLevel='1'.			
Warning	1	null	IsFocused	MetroThumb.NoTarget, Name='PART_LeftThumb'	Object	Cannot find source: RelativeSource FindAncestor, AncestorType='System.Windows.Controls.Slider', AncestorLevel='1'.			
[...]

PR #2449, opened 2021-09-30
Fix RangeSlider style to match Slider

Yoooi0:
Fix for #2350

There is no RangeSlider demo in the demo app. Adding RangeSlider demo would require referencing `MahApps.Metro` and `MaterialDesignThemes.MahApps.csproj` so I didnt include it in this PR.

It probably still doesnt match 100% in style/behavior, but this is how it looks now:

### DIFF
--- a/MaterialDesignThemes.MahApps/Themes/MaterialDesignTheme.MahApps.RangeSlider.xaml
+++ b/MaterialDesignThemes.MahApps/Themes/MaterialDesignTheme.MahApps.RangeSlider.xaml
@@ -124,25 +124,25 @@
                         Background="Transparent"
                         Orientation="Horizontal">
                 <RepeatButton x:Name="PART_LeftEdge"
-                              Foreground="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=Foreground}"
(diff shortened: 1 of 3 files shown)


MaterialDesignThemes.MahApps/Themes/MaterialDesignTheme.MahApps.RangeSlider.xaml
```diff
@@ -121,25 +124,25 @@
                         Background="Transparent"
                         Orientation="Horizontal">
                 <RepeatButton x:Name="PART_LeftEdge"
-                              Foreground="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=(controls:SliderHelper.TrackFillBrush)}"
+                              Foreground="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=Foreground}"
```
Keboo, line 127: What is the reason for changing away from the attached properties to Foreground? Typically this results in less customizability.

Yoooi0: Yes, not sure what I was thinking there.
Reverted.

Keboo (approved):

PR #2469, opened 2021-10-16
Default the TextBoxBase/PasswordBox SelectionBrush

JeffBarnard:
Instead of using SystemColors which will rarely match the users material theme, let's consider defaulting the TextBoxBase/PasswordBox SelectionBrush property to PrimaryHueLightBrush as per the material spec.
https://material.io/design/color/text-legibility.html#text-types
image


PentaTech:
Is it possible to add this to other controls like pickers and editable comboboxes?

PR #2481, opened 2021-11-02
Add Progressbar to Flat and Outlined buttons

### LINKED ISSUE #2478: `ProgressBar` is not implemented for flat and outlined buttons
Currently `ProgressBar` is available in raised buttons, but not in flat or outline buttons. Was this a design choice, or did we simply forget to implement it for non-raised buttons?

StefanoRivolta-Previero:
fixes #2478 

result:

immagine

(up left with custom colors)


database64128:
The implementation has some bugs. When multiple buttons in a stack panel enter indeterminate state at the same time, the  2nd button is distorted:

image

image

This doesn't happen when both buttons are raised buttons.

The buttons in question are https://github.com/database64128/youtube-dl-wpf/blob/8cee23b31be37215ad336934b510b3e8195b1644/YoutubeDl.Wpf/Views/HomeView.xaml#L64-L70.

Issue #2495, opened 2021-11-18
SelectedDate Isn't Null After Clicking ClearButton


MohammadHadi2031:
Hi, there is a attached property called TextFieldAssist.HasClearButton. I enabled it for a DatePicker. When I click the clear button the text is cleared but the SelectedDate property is not null.

[github repo](https://github.com/MohammadHadi2031/MaterialDesignClearButton.git)

PR #2498, opened 2021-11-23
Set DatePicker SelectedDate to null on clicking the TextFieldAssist ClearButton

ahmed-abdelrazek:
that should fix https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2495


Keboo:
Unfortuently this doesn't appear to actually null out the SelectedDate property (though it does clear the text). After digging into it a bit more I have opened an alternate PR (#2502) to address it, since it is a bit more complete. Due to how the DatePicker's template is setup the clear button is not actually part of its template so it fails to find the `PART_ClearButton` element.

PR #2509, opened 2021-11-30
Fix border in buttons with progressbar

StefanoRivolta-Previero:
Fixed bug introduced in #2481 where in outlined and flat buttons with progress bar and no fixed width, the button would go to infinity.

before:
nEZddjXjAL

after:
HSRTNvFmv1

Also fixed progress bar for secondary colored buttons


Keboo (approved):

database64128:
Before we had progressbar in outlined buttons, a disabled outlined button looked like this:

image

Now with the progressbar, they look like this:

image

Maybe the background and foreground of the progress bar should be the other way around?

database64128:
Update: I opened #2530 to fix the problem mentioned in the above comment.

Issue #2316, opened 2021-05-04
Add PlacementMode to DialogHost


Drise13:
Conversation context: https://gitter.im/ButchersBoy/MaterialDesignInXamlToolkit?at=609183bc963923553ae2d4d4

I'm looking for a way to control the placement of the dialog popup. Currently it seems to always center. For my application I'm looking for a `Top` mode, but I image other standard modes would be useful as well.

MichelMichels:
Resolved with PR.

PR #2520, opened 2021-12-08
Add Placement Property to DialogHost

ahmed-abdelrazek:
should address https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2316

### DIFF
--- a/MaterialDesignThemes.Wpf/DialogHost.cs
+++ b/MaterialDesignThemes.Wpf/DialogHost.cs
@@ -365,6 +365,15 @@ namespace MaterialDesignThemes.Wpf
             set => SetValue(IsOpenProperty, value);
         }
 
+        public static readonly DependencyProperty PlacementProperty = DependencyProperty.Register(
+            nameof(Placement), typeof(PlacementMode), typeof(DialogHost), new PropertyMetadata(PlacementMode.Center));
+
+        public PlacementMode Placement
+        {
+            get => (PlacementMode)GetValue(PlacementProperty);
+            set => SetValue(PlacementProperty, value);
+        }
+
         public static readonly DependencyProperty DialogContentProperty = DependencyProperty.Register(
             nameof(DialogContent), typeof(object), typeof(DialogHost), new PropertyMetadata(default(object)));
 
--- a/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.DialogHost.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.DialogHost.xaml
@@ -139,6 +139,7 @@
                             </VisualStateGroup>
                         </VisualStateManager.VisualStateGroups>
                         <wpf:PopupEx PlacementTarget="{Binding ElementName=DialogHostRoot, Mode=OneWay}"
+                                     Placement="{TemplateBinding Placement}"
                                      x:Name="PART_Popup" 
                                      Style="{TemplateBinding PopupStyle}"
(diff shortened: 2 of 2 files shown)


Keboo:
This mostly works, however there is some weirdness with some of the values (Absolute, AbsolutePoint, Custom). It also does not work for the embedded DialogHost style. 

I am thinking that perhaps a cusom enum that only contains a few of the values (Top, Bottom, Left, Right, Center), with some value converters to convert to the Popup's PlacementMode and to handle the embedded DialogHost case. 

Thoughts?

ahmed-abdelrazek:
@Keboo is there something wrong with (Absolute, AbsolutePoint, Custom)? i think they work as excepted when i tested them
right now i can't see an obvious way to make this work with the embedded DialogHost style i will try and debug it with snoop

Keboo:
@ahmed-abdelrazek not really anything wrong with it, as it does work (at least in the regular popup-based template). My thought with using a custom enum for the values was to try and reduce the [leaky abstraction](https://medium.com/young-coder/what-are-leaky-abstractions-an-illustrated-guide-f2982ff21cae). This is certainly a judgement call, and not a hard rule, but in this case, I think hiding the fact that the dialog is shown in a popup is a detail that is best hidden, especially since there is also the embedded style that does not use the popup. 

Sorry for not being clear with what I was thinking with the value converter. What I was thinking was something like this (pseudo code; not tested).

```C#
public enum DialogPosition
{
    Center,
    Left,
    Top,
    Right,
    Bottom
}

public class PopupPlacementValueConverter : IValueConverter
{
    ....
    public object Convert(object value, ....)
    {
        if (value is DialogPosition dialogPosition)
        {
            return dialogPostion switch
            {
                DialogPosition.Left => PlacementMode.Left,
                _ => PlacementMode.Center
            }
        }
        return Binding.DoNothing;
    }
    ....
}

public class DialogHorizontalValueConverter : IValueConverter
{
    ....
    public object Convert(object value, ....)
    {
        if (value is DialogPosition dialogPosition)
        {
            return dialogPostion switch
            {
                DialogPosition.Left => HorizontalAlignment.Left,
                _ => HorizontalAlignment.Center
            }
        }
        return Binding.DoNothing;
    }
    ....
}
```

And then in the XAML it would the popup would look something like....

```xaml
Placement="{Binding Placement, RelativeSource={RelativeSource TemplatedParent}, Converter={StaticResource PopupPlacementValueConverter}}"
```

I would then imagine something similar for the horizontal/vertical alignment for the dialog in the embedded case.

Does that make sense?

ahmed-abdelrazek:
yeah i got your point about the Converter i just thought there is something wrong with how the placement property worked

ahmed-abdelrazek:
@Keboo while the converter approach will work for both cases it will be limiting at least in my use case where i want to show it under the mouse curser or when i want to show it on top left or something
i think the current one works better for normal `DialogHost` with the full range of Placement options but when it come to `MaterialDesignEmbeddedDialogHost` they can just use `HorizontalContentAlignment` and `VerticalContentAlignment` to control the dialog position or maybe `MaterialDesignEmbeddedDialogHost` need to be rewritten to support `Placement`

Keboo:
Ah, I had not considered that those values were actually desirable.
Perhaps then for now, let's leave the property as you have it, but add in converters for the embedded dialog case. I am thinking that it really only needs to support Left, Top, Right, Bottom, and Center. All of the other ones should probably be ignored and just continue to center the dialog.

PR #2525, opened 2021-12-13
Fixed the MaterialDesignFilledComboBox style

MichelMichels:
The Filled ComboBox style was broken by previous code.

Broken style: 
image

Fixed style: 
image

I changed the order of the controls inside the "MaterialDesignFloatingHintComboBoxTemplate" ControlTemplate. In the AdornerDecorator there's a Grid in which the first item was the ToggleButton (the little triangle) that opens or closes the ComboBox. The TemplateBinding Background was applied to this item, but this control doesn't apply the CornerRadius property. So I changed the order of the controls and applied the Background TemplateBinding to the Border.

Current order inside Grid:
1. Border
2. wpf:Underline
3. Canvas
4. wpf:ComboBoxPopup
5. ToggleButton
6. Button (the clear button)

Previous order inside Grid: 
1. ToggleButton
2. Border
3. wpf:Underline
4. Canvas
5. wpf:ComboBoxPopup
6. Button (the clear button)

### DIFF
--- a/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.ComboBox.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.ComboBox.xaml
@@ -304,19 +304,14 @@
         </Setter>
     </Style>
 
-    <ControlTemplate x:Key="MaterialDesignFloatingHintComboBoxTemplate" TargetType="{x:Type ComboBox}">
+    <ControlTemplate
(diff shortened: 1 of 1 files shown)


MichelMichels:
PS: This style isn't quite up to spec atm. The ToggleButton should be centered inside the Grid. This is quite easy to achieve, but the clear button also has to be centered and I don't quite see how I can easily achieve this.

The Hint should also be centered before it gets selected.

Keboo (approved):
Great work. Thank you!

PR #2591, opened 2022-03-02
Menu improvements

Xaalek:
I made two improvements for menus :
- Menu headers are now transparent
- Menu headers now fill the available height

I also added some examples in the demo

image

image

Please note that since the menus no longer have a fixed height, this change could affect the layout of the controls on a page. So the people who use the menus will have to manually set a fixed height on the menus to get the same layout.


Keboo (approved):
Great work! Thank you

Issue #2578, opened 2022-02-14
Update to MaterialDesign 3 ListBox


StefanoRivolta-Previero:
I've updated the ListBox style to match the new MaterialDesign 3 guidelines. You can find the changes in this fork:

https://github.com/StefanoRivolta-Previero/MaterialDesignInXamlToolkit/tree/listbox  

## [Material Design Guidelines](https://m3.material.io/components/navigation-rail/specs):

<img width="676" alt="Light" src="https://user-images.githubusercontent.com/58171461/153855441-c0479ee3-1d48-4830-8e22-674c4ba3a49a.png"><img width="1137" alt="Dark" src="https://user-images.githubusercontent.com/58171461/153855444-3043c813-532e-4651-bc40-1148cf9fc772.png">

## Result:
![vcADXe0g21](https://user-images.githubusercontent.com/58171461/153855650-2e80bfaf-ae0d-4d50-9c82-5da913a11864.png)

The process to create an Item is a bit complex, since it needs two icons (one filled for when it's selected, and one outlined for when it's unselected) and a text under the selection border. 
I've opted for the content to only be the text part, while the icons get set trough a style property, like this:

```
<ListBoxItem>
    <ListBoxItem.Style>
        <Style TargetType="ListBoxItem" BasedOn="{StaticResource MaterialDesign3SmallNavigationRailPrimaryListBoxItem}">
            <Setter Property="materialDesign:ListBoxItemAssist.SelectedIcon" Value="Circle"/>
            <Setter Property="materialDesign:ListBoxItemAssist.UnselectedIcon" Value="CircleOutline"/>
        </Style>
    </ListBoxItem.Style>
    <TextBlock Text="Test 1"/>
</ListBoxItem>
```
Or like this with ItemSource Binding
```
<ListBox.ItemContainerStyle>
    <Style TargetType="ListBoxItem" BasedOn="{StaticResource MaterialDesign3NavigationRailPrimaryListBoxItem}">
        <Setter Property="materialDesign:ListBoxItemAssist.SelectedIcon" Value="{Binding SelectedIcon}"/>
        <Setter Property="materialDesign:ListBoxItemAssist.UnselectedIcon" Value="{Binding UnselectedIcon}"/>
    </Style>
</ListBox.ItemContainerStyle>
<ListBox.Resources>
    <DataTemplate DataType="{x:Type local:ItemSample}">
        <TextBlock Text="{Binding Title}"/>
[...]

PR #2628, opened 2022-03-30
NavigationRail, NavigationBar, NavigationDrawer controls for MaterialDesign3

StefanoRivolta-Previero:
# Summary

From issue #2578 

I've added the Navigation Rail, Navigation Bar and Navigation Drawer controls as styles for the ListBox. This is probably not ideal, as it had already been pointed out, as it limits some of the things that can be done, but it was also somewhat easier for me to do rather than creating a completely new CustomControl. In the future that may be a better way to implement the controls though.

The messy part has to do with the Icons, since it needs both a selected version and an unselected (outlined) version. The best way I came up with was to make them a part of the Style, so it works like this:

```
<ListBox.ItemContainerStyle>
    <Style TargetType="ListBoxItem" BasedOn="{StaticResource MaterialDesign3NavigationDrawerPrimaryListBoxItem}">
        <Setter Property="materialDesign:NavigationDrawerAssist.SelectedIcon" Value="{Binding SelectedIcon}"/>
        <Setter Property="materialDesign:NavigationDrawerAssist.UnselectedIcon" Value="{Binding UnselectedIcon}"/>
    </Style>
</ListBox.ItemContainerStyle>
```
Or this without MVVM:

```
<ListBoxItem>
    <ListBoxItem.Style>
        <Style TargetType="ListBoxItem" BasedOn="{StaticResource MaterialDesign3NavigationRailPrimaryListBoxItem}">
            <Setter Property="materialDesign:NavigationRailAssist.SelectedIcon" Value="Circle"/>
            <Setter Property="materialDesign:NavigationRailAssist.UnselectedIcon" Value="CircleOutline"/>
            <Setter Property="materialDesign:NavigationRailAssist.IconSize" Value="16"/>
        </Style>
    </ListBoxItem.Style>
    <TextBlock Text="text 1"/>
</ListBoxItem>
```

I still have to add the Badge to the items, as per spec. Waiting for feedback before I do. I also haven't re-implemented the TabControl style for NavigationRail, should I?

# Navigation Rail

Specs: https://m3.material.io/components/navigation-rail/overview

sQvUiT4VkE

# Navigation Bar

Specs: https://m3.material.io/components/navigation-bar/overview

EASLg7jgj8

# Navigation Drawer

Specs: https://m3.material.io/components/navigation-drawer/overview

7DRbW1TXJH

# Demo App

I've updated the DemoApp to use NavigationRail, NavigationBar and NavigationDrawer in relation to the size of the window, similarly to this: https://m3.material.io/foundations/adaptive-design/overview

devenv_WyfVKEp0fH

I'll wait for feedback. 
By the way, should we differentiate the *Assist.cs files between MD2 and MD3 or should we leave them the same?


MaterialDesignThemes.Wpf/MaterialDesignThemes.Wpf.csproj
```diff
@@ -27,4 +27,17 @@
...
+      <SubType>Designer</SubType>
+    </Page>
+    <Page Update="Themes\MaterialDesign3.NavigationDrawer.xaml">
+      <XamlRuntime>$(DefaultXamlRuntime)</XamlRuntime>
+      <SubType>Designer</SubType>
+    </Page>
+    <Page Update="Themes\MaterialDesign3.NavigationRail.xaml">
+      <XamlRuntime>$(DefaultXamlRuntime)</XamlRuntime>
+    </Page>
+  </ItemGroup>
```
Keboo, line 42: I believe these lines can be omitted.

MaterialDesign3.Demo.Wpf/Domain/SampleItem.cs
```diff
@@ -0,0 +1,10 @@
+﻿using MaterialDesignThemes.Wpf;
+
+namespace MaterialDesign3Demo.Domain;
+
+public class SampleItem
+{
+    public string Title { get; set; }
```
Keboo, line 7: Because nullable reference types is enabled, this should be declared as a nullable string (this is the build failure on the pipeline; only a warning when compiled locally)
```suggestion
    public string? Title { get; set; }
```

Keboo (commented):
First of all, let me start by saying thank you. This work looks amazing!

I left a few small comments, but I think this is very close to being merged. 
As for your question on the *Assist classes. My preference would be to keep them together. Attached properties are difficult to discover as it is, so I would be hesitant to separate them out. With that said, I suspect there may be cases where an attached property only applies to MD2 or MD3. For those I think we will just need to evaluate on a case-by-case basis.

StefanoRivolta-Previero:
I've fixed the errors and added the badge to the navigation controls. 
To pass through the properties of the badge, I had to create a badge assist class, not sure if it's the right way to do it though.
I also added a new style for the badged control, so that it only shows the small dot, no text.

Some changes also to the Demo App. To showcase the badge I've added a Notifications property to the DemoItem class and a little control to increase or dismiss the notifications. Also new an extended FAB and a collapse button in the drawer when in standard mode.

devenv_JxZR7mTPYK

BigBadBleuCheese:
Where is this groovy new demo? I downloaded the 4.5.0 demo and it seems mostly as it was.

Keboo:
@BigBadBleuCheese it lives under the [MaterialDesign3.Demo.Wpf directory in the repo](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/tree/master/MaterialDesign3.Demo.Wpf). However your comments reminds me that it has not been added to the pipeline which is why it is not currently being included in the release artifacts. I will work on getting that added, because it is pretty cool.

Issue #2616, opened 2022-03-23
TabControl multi-row broken after update 4.3.0 -> 4.4.0


JanZeman:
Hello,

I haven't touched UI part of my project for many months. I do however regularly update NuGet packages including MaterialDesignThemes - most of the time without any problems. Today I updated from 4.3.0 -> 4.4.0 and noticed the Settings page broken. It is one huge TabControl with many tabs - intended to show two rows of tabs. After the update it shrank to one row only which makes it user unfriendly. It is couple of years back I designed this page, I do not remember (and I do not see it in the current code) I had to do anything special to get multi-row support.

Please see the two images.

4.3.0:
![image](https://user-images.githubusercontent.com/1670470/159800286-96a77d1b-f71a-455d-9144-071d003a3245.png)

4.4.0:
![image](https://user-images.githubusercontent.com/1670470/159800314-d8b78aec-e44b-40dc-baf5-35f9b0d7be1b.png)

Is this an intended change and I have to do anything special in my client code or is it a bug?

Awsmolak:
Found this while checking for a different issue. But the toolkit _just_ introduced the tabcontrol style. So now that is being set as the default style for TabControl. And the tabitems are in a uniform grid that can't wrap. I'm not exactly sure how to do this, but to go back to what you had you need to set the style back to the WPF default for TabControl

JanZeman:
Hi @Awsmolak, thank you for the tip. It's been a while I touched the UI but I remembered it is possible to create default styles by right-clicking, Edit Template > Edit the copy... That gave me slightly different version of the sample you posted.

Applying this manually created default style made it back to normal - multi-row displayed nicely.

I however hesitate closing the bug as multi-row should be supported as a standard feature imho. Let you guys to decide...

ahmed-abdelrazek:
https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2603

Keboo:
So for the material design, multiple rows of tabs are not expected (that is a WPF design feature), so if the original WPF look and behavior is desired going back to the original WPF style is the way to go. The above mentioned options certainly work, however, there is an even easier option. Simply declare empty implicit styles targeting both the `TabControl` and the `TabItem`.
```xaml
<Style TargetType="TabControl" />
<Style TargetType="TabItem" />
```
[...]

Issue #2603, opened 2022-03-18
TabControl - all TabItems Headers width is always set to the widest tab


ahmed-abdelrazek:
beside #2602 using the example from the demo (changing HorizontalContentAlignment to left or right) but changing the second header to add more text will result in both tabs to have the second tab width (the widest tab)
`
<TabControl HorizontalContentAlignment="Left" materialDesign:ColorZoneAssist.Mode="SecondaryMid">
                        <TabItem Header="TAB1">
                            <TextBlock Margin="8" Text="Full-width secondary example tab 1" />
                        </TabItem>
                        <TabItem Header="TABBBBBBBBBBBBBBBBBBBB 2">
                            <TextBlock Margin="8" Text="Full-width secondary example tab 2" />
                        </TabItem>
                    </TabControl>
`
![image](https://user-images.githubusercontent.com/7858946/158997610-a06214d8-89ef-4cfc-8a78-8d49fd3a642a.png)

also it's doesn't handle overflow 
![image](https://user-images.githubusercontent.com/7858946/158997937-e658e405-918b-4eaf-83fd-4e6c58d64dd7.png)

ahmed-abdelrazek:
changing this to stackpanel or wrappanel should fix the issue https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/24c6468b8e93211e1283ba4f4cf8765baa81486e/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.TabControl.xaml#L54-L61

veiny:
UniformGrid`s children elements are same width/height.

ahmed-abdelrazek:
@veiny yeah that's what i found after looking into the code to track the issue changing it to any panel i suggested should fix the size thing if it was not intentional but we will still have the overflow thing to handle do we need them stacked on top of each other do we need a scrollbar martial design seems to implement hidden scrollbar https://material.io/components/tabs#interactive-demo

MichelMichels:
I would go with the scrollbar as per spec.

As for the UniformGrid, I updated the TabControl a while ago and I can't remember if I put the UniformGrid there or if it was already present. Anyhow, the Material spec is not clear in wether this should be all the same width or not. All the examples have a same-width header so I did not think further than that. I think UniformGrid is used to achieve a full-width (accross the screen) TabControl and have the headers spaced evenly.

ahmed-abdelrazek:
[...]

PR #2634, opened 2022-04-01
Fix Tabs Width and Overflowing issues

ahmed-abdelrazek:
Fix for https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2603 https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2616

it might be a better way to do this 
also you can't scroll horizontal tabs with the mouse wheel -i think it's a wpf limitation (there are workarounds)- when your horizontal tabs are overflowing you can still navigate them with the keyboard arrow keys

MaterialGif


Keboo:
This is great. Thank you!

PR #2644, opened 2022-04-07
allow setting background of menu item

krmr:
With the current version I unfortunately could not easily change the background of a menu. Setting the background property on a MenuItem simply had no effect. Here the result of the included sample code where the foreground change is applied but not the background which results in white on white:
image

With the changes in this pull request the result changes to this:
image
_A common use case would use the same color for both the main and child menu items but I used different colors for illustration._

### DIFF
--- a/MainDemo.Wpf/MenusAndToolBars.xaml
+++ b/MainDemo.Wpf/MenusAndToolBars.xaml
@@ -93,6 +93,11 @@
                         <MenuItem Header="Test 1"/>
                         <MenuItem Header="Test 1"/>
                     </MenuItem>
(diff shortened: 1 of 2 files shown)


krmr:
I did not create a separate issue for this. Feel free to let me know if you want me to create one.

@Xaalek I guess this is related to your recent work on the menus.

Keboo (approved):
Great work! Thank you for the contribution!

Xaalek:
The background of the nested menu of the toolbar has become white in the demo. The background should be transparent. Is this a bug? @krmr

Issue #2626, opened 2022-03-30
Better control of menu height


Xaalek:
As this gitter conversation specifies, we need to change the behaviour of the Menu control : 
https://gitter.im/ButchersBoy/MaterialDesignInXamlToolkit?at=623d5f32c61ef0178e8d09df

> Replace the height binding that was added in the PR linked above to instead point to a new MenuAssist.TopLevelMenuItemHeight attached property. In the Menu style the default value for that property should be set back to 48. In the demo app for the large and small menu examples, we should include an example of setting that to use the menu size. Something similar to:
```
<Menu md:MenuAssist="{Binding ActualHeight, RelativeSource={RelativeSource Self}} ... >
...
</Menu>
```
Linked PR : https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/commit/d306a56abfbbab5f5f54f2432735fda7a6b93fda#diff-fd12eb1f1fae6617ce2e8d03c219b772139250243fea685188868b656fd8069aL337

Xaalek:
I tried to create the property but I have a problem. Does anyone know where this could be coming from?

![image](https://user-images.githubusercontent.com/54487782/164743068-04ad7a5e-9a99-4e81-a489-92eae356dd93.png)


```
System.Windows.Data Information: 41 : BindingExpression path error: '(wpf:MenuAssist.TopLevelMenuItemHeight)' property not found for 'object' because data item is null.  This could happen because the data provider has not produced any data yet. BindingExpression:Path=(0); DataItem=null; target element is 'MenuItem' (Name=''); target property is 'Height' (type 'Double')
System.Windows.Data Information: 20 : BindingExpression cannot retrieve value due to missing information. BindingExpression:Path=(0); DataItem=null; target element is 'MenuItem' (Name=''); target property is 'Height' (type 'Double')
System.Windows.Data Information: 21 : BindingExpression cannot retrieve value from null data item. This could happen when binding is detached or when binding to a Nullable type that has no value. BindingExpression:Path=(0); DataItem=null; target element is 'MenuItem' (Name=''); target property is 'Height' (type 'Double')
System.Windows.Data Information: 10 : Cannot retrieve value using the binding and no valid fallback value exists; using default instead. BindingExpression:Path=(0); DataItem=null; target element is 'MenuItem' (Name=''); target property is 'Height' (type 'Double')
```

This is the class : 
```
    public static class MenuAssist
    {
[...]

PR #2657, opened 2022-04-25
Better control of menu height

Xaalek:
I made the changes mentioned here :
https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2626

- Add attached property `TopLevelMenuItemHeight`
- Set back the default height to 48
- Restructuring and adding examples to the demo app

image

image

### DIFF
--- a/MainDemo.Wpf/Buttons.xaml
+++ b/MainDemo.Wpf/Buttons.xaml
@@ -1091,5 +1091,39 @@
                 </materialDesign:PopupBox>
             </smtx:XamlDisplay>
         </StackPanel>
+
+        <Rectangle
+            Margin="0 24 0 0"
+            Height="1"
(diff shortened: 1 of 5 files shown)


Xaalek:
The default Menu has now a height of 48.
To get an adaptive menu we have to do something like this :
```
<materialDesign:Card>
   <Menu materialDesign:MenuAssist.TopLevelMenuItemHeight="{Binding RelativeSource={RelativeSource FindAncestor, AncestorType={x:Type materialDesign:Card}},Path=ActualHeight}">
       ...
   </Menu>
</materialDesign:Card>
```

Is that what we wanted?

MichelMichels:
This seems alright, the change is much appreciated! Thank you.

Keboo (approved):

Keboo:
Awesome! Thank you for doing this.

PR #2662, opened 2022-05-02
adding ComboBoxAssist.MaxLength property

Xaalek:
Adding ComboBoxAssist.MaxLength property as mentioned here : https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2442

Issue #2442, opened 2021-09-27
Add ComboBoxAssist.MaxLength attached property in ComboBox to control MaxLength when IsEditable is set to True


jamesport079:
ComboBox with IsEditable set to true essentially behaves as a TextBox with an autosuggest dropdown list. A silly ommission in WPF is that MaxLength cannot be set in XAML for an editable ComboBox. 

I created an attached property to address this shortcoming. I strongly feel this can be integrated in this library in the form of  materialDesign:ComboBoxAssist.MaxLength="___" helper class. 

The code to implement it is pretty straightforward, essentially creating an attached property MaxLength and finding the TextBox part in ComboBox and setting MaxLength on TextBox. FindChild<T> does the magic to find the TextBox part of the editable ComboxBox. 

    private static void OnMaxLenghtChanged(DependencyObject obj, DependencyPropertyChangedEventArgs args)
    {
        if (obj is ComboBox comboBox)
        {
            comboBox.Loaded +=
              (s, e) =>
              {
                  DependencyObject? textBox = comboBox.FindChild<TextBox>("PART_EditableTextBox");
                  if (textBox == null)
                  {
                      return;
                  }

                  textBox.SetValue(TextBox.MaxLengthProperty, args.NewValue);
              };
        }
    }

If you think this is beyond the scope of this library I would at least suggest to add TextFieldAssist.CharacterCounterVisibility to be accessible in an editable ComboBox for the sake of consistency.

Xaalek:
`FindChild` does not exist in the current context. Do you mean `FindName` ?

jamesport079:
> 
sorry, here's the code for FindChild...

`public static class UIChildFinder
{
    /// <summary>
    /// Finds a Child of a given item in the visual tree. 
    /// </summary>
    /// <param name="parent">A direct parent of the queried item.</param>
    /// <typeparam name="T">The type of the queried item.</typeparam>
    /// <param name="childName">x:Name or Name of child. </param>
    /// <returns>The first parent item that matches the submitted type parameter. 
    /// If not matching item can be found, 
    /// a null parent is being returned.</returns>
    public static T? FindChild<T>(this DependencyObject parent, string childName)
       where T : DependencyObject
    {
        // Confirm parent and childName are valid. 
        if (parent == null) return null;

        T? foundChild = null;

        int childrenCount = VisualTreeHelper.GetChildrenCount(parent);
[...]

PR #2664, opened 2022-05-03
Continuation of #2662

Keboo:
Uses relative source binding to avoid looking up elements by name.
Added example to the demo project.

Related: #2442


Keboo:
FYI @Xaalek

Xaalek:
Thanks, I thought so.

Issue #2475, opened 2021-10-29
Material Design 3 Roadmap


Keboo:
Google has release Material Design 3. The intent here is to outline the needed work to support it in this library.

You can find the latest details [here](https://m3.material.io/).

All of the tasks listed here can be considered up for grabs and I am happy to accept PRs for any of them.

New control resource dictionaries should be named as:
`MaterialDesign3.<ControlName>.xaml`

## High level changes
At a minimum these tasks and the color changes need to be implemented to start releasing these new styles.

#### Tasks

- [x] Create a copy of the `MaterialDesignTheme.Defaults.xaml` resource dictionary, it should be called `MaterialDesign2.Defaults.xaml`. 
- [x] Create a new top level resource dictionary along side `MaterialDesignTheme.Defaults.xaml` it should be called `MaterialDesign3.Defaults.xaml`. For all of the new M3 styles, they should get referenced by this resource dictionary. 

### [Color](https://m3.material.io/styles/color/overview)

This library already support some amount of the new Dynamic Colors, will need to evaluate the M3 implementation to determine how viable it will be to implement here as well.

Because the color brushes are all going in the same code path (ie will be treated and available at the app level), there will be some un-used brushes depending on which control styles are being used.

#### Tasks

- [ ] Add tertiary color to the theme classes
- [ ] Add container brushes for primary, secondary, and tertiary colors
- [ ] Add neutral color brushes. Evaluate if the "Surface" ones are actually needed (I suspect we can simply use the background brushes instead).

### [Typography](https://m3.material.io/styles/typography/overview)

There is another font Noto Sans that could be included in this library. Right now the fonts make up a significant percentage of size, this would add approximately 1.3MB. Alternatively we could pull fonts out, and make them opt-in with a csproj flag or similar. I am interested in hearing feedback on this, not sure if file size of the assembly matters much to people. 

#### Tasks

- [ ] Create a copy of the existing `MaterialDesignFontExtension` and name it `RobotoFontExtension`. `MaterialDesignFontExtension` should be marked as obsolete.
- [ ] Create a new `NotoFontExtension` (similar to the one above) and include the Noto Sans font files. 

### [Buttons](https://m3.material.io/components/all-buttons)
This will be the largest portion of the work as there is quite a bit to do.

#### Tasks

- [ ] Create style for Elevated button
- [ ] Create style for Filled button
[...]

PR #2748, opened 2022-06-19
Added MD3 version of MaterialDesignSwitchToggleButton

jizc:
Part of issue #2475

This PR adds a Material Design 3 version of the `MaterialDesignSwitchToggleButton` style.

Md3Switch
Md3Switch

### Specs
https://m3.material.io/components/switch/specs

### Notes
- The other styles in the `MaterialDesign3.ToggleButton.xaml` resource dictionary are just copies of the ones in `MaterialDesignTheme.ToggleButton.xaml`
- The colors don't match the specs exactly. This is the closest I could come with the current color system. We should give it another pass when/if a more complete MD3 color system is implemented.

### DIFF
--- a/MaterialDesign3.Demo.Wpf/App.xaml
+++ b/MaterialDesign3.Demo.Wpf/App.xaml
@@ -44,7 +44,7 @@
                             <ResourceDictionary
                                 Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Button.xaml" />
                             <ResourceDictionary
-                                Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.ToggleButton.xaml" />
+                                Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesign3.ToggleButton.xaml" />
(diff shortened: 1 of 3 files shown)


Keboo (approved):
Great work thank you!

jizc:
@soghomonyansargis It hasn't been released yet. Look for it in the upcoming [4.6.0 release](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/milestone/38).

Nextra:
These look fantastic, but they are absolutely massive compared to the MD2 variant. The box size is pretty much double, and without the borders the MD2 variant feels much smaller visually on top of that.

How difficult would it be to have a separate style that more closely matches the MD2 size, or even make this one adjustable? Simply changing the height breaks the styling of the MD3 variant unfortunately.

Oszoou:
Indeed, they are massive. Here is a temporary solution.

```xaml
<ToggleButton Style="{StaticResource MaterialDesignSwitchToggleButton}">
    <ToggleButton.RenderTransform>
        <ScaleTransform CenterX="0" CenterY="0" ScaleX="0.8" ScaleY="0.8" />
    </ToggleButton.RenderTransform>
</ToggleButton>
```

Issue #2658, opened 2022-04-25
Outlined card style is missing


Xaalek:
As you can see in the specs, there is an outlined card style : 
https://material.io/components/cards

We don't have this style in the library. It would be nice to implement it.

**- Edit -**
- [ ] Create an Elevated style that matches the current style
- [ ] Create an Outlined style that matches the specs

Keboo:
Agreed, this would be a nice addition.

Xaalek:
So we have to create two styles for the card? One for the Elevated style and one for the Outlined style, right?

Keboo:
Correct

PR #2757, opened 2022-07-02
Fix2658

nicolaihenriksen:
Fix for #2658

Split Card template into 2 styles: Elevated and Outlined.

For the outlined style, I was unsure whether to add a new color or use an existing one. It was not 100% clear where I should add that color if I wanted it, so I opted to reuse the MaterialDesignDivider brush which seems to resemble the dimmed look from the material.io design pretty well.

Added unit test for newly added converter and UI tests for newly added card style.

In order to get the UI tests to run, I had to include Generic.xaml in the merged resources dictionary. I think this "happens automatically" when running a WPF app, but for the test it seemed necessary to add manually. Perhaps this is something that should be included in the XAMLTest library instead?

### DIFF
--- a/MainDemo.Wpf/Cards.xaml
+++ b/MainDemo.Wpf/Cards.xaml
@@ -84,6 +84,68 @@
         </smtx:XamlDisplay>
 
         <smtx:XamlDisplay
+            UniqueKey="cards_1b"
+            Margin="4 4 0 16"
+            VerticalContentAlignment="Top">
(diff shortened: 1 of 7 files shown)


nicolaihenriksen:
I was a little unsure of whether I should merge two nested Borders into one, but I left it as 2 which were "copied" from the Elevated style and modified to not use the drop shadow, but have a solid border instead.

Keboo (approved):
This is really great work! I appreciate the additional tests that you added as well.

Xaalek:
@nicolaihenriksen Thank you for your contribution, it's really appreciated!

Issue #2689, opened 2022-05-11
Request: Add CornerRadius to Flipper


mikechristiansenvae:
Requesting that a `CornerRadius` property be added to `Flipper`.

PR #2758, opened 2022-07-02
Added UniformCornerRadius to Flipper

nicolaihenriksen:
Fix for #2689

The new DP is only used in the MaterialDesignCardFlipper style. For the default style, calling code can simply add rounded corners to the content that is put in.

Kind of feels like there there should be a UniformCornerRadius attached property for the Card instead. This would avoid exposing a DP on Flipper.cs that is not used in all styles? Would this go into a CardAssist.cs file or would the attached property go directly in the Card.cs file alongside the "real" DP?

### DIFF
--- a/MainDemo.Wpf/Cards.xaml
+++ b/MainDemo.Wpf/Cards.xaml
@@ -551,6 +551,62 @@
                 </materialDesign:Flipper.BackContent>
             </materialDesign:Flipper>
         </smtx:XamlDisplay>
+
+        <StackPanel>
+            <smtx:XamlDisplay
+                UniqueKey="cards_10"
+                Margin="4 4 0 0"
+                VerticalContentAlignment="Top">
+                <materialDesign:Flipper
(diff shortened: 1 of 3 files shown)


Keboo (approved):
Great work, thank you!

nicolaihenriksen:
> Great work, thank you!

What are your thoughts on the attached property approach? Do you think it is better with a DP that only applies to one of the styles (ie. what this PR includes) as opposed to not having a DP on the Flipper, but instead an attached property that can be applied when using the "card-specific" style?

Keboo:
I think for properties that only sometimes apply, the attached property is preferable to a full DP.

nicolaihenriksen:
> I think for properties that only sometimes apply, the attached property is preferable to a full DP.

In that case, shouldn't I have refactored into the attached property approach before this PR was merged? And also, there currently isn't a CardAssist class as far as I can tell, would you put the attached property directly in the Card class (next to the full DP)?

PR #2759, opened 2022-07-04
Add ExpanderAssist.CanToggle, which enables/disables expanding/collapsing an expander

binarycow:
This attached property, `ExpanderAssist.CanToggle`, gives you the ability to control whether or not an `Expander` can be expanded/collapsed.

- Default value: `true` (i.e., no change in behavior from current)
- Implementation: When set to `false`, sets `IsHitTestVisible` to `false` on the `HeaderSite` `ToggleButton` (the one that encompasses the _entire_ header - not just the arrow/chevron)
- The user cannot expand/collapse the expander by clicking on the header.
- The user does not get any mouseover indications they can click on the header

----

While this behavior may seem counter-intuitive at first, it has its uses.  

I recently developed an `Accordion` control, where each item was represented by an `Expander`.  When the user clicks the header of a collapsed item, the currently opened item will close, and the clicked-on item will open.  With the built-in `Expander` control, the user would be able to collapse the current item, and then all items would be closed.  I did not want to allow this behavior.

Either way, it's a very low-impact change.  Implementing it in the library would mean I don't need to re-implement the `ControlTemplate`.

Thanks in advance!

### DIFF
--- a/MainDemo.Wpf/Expander.xaml
+++ b/MainDemo.Wpf/Expander.xaml
@@ -2,14 +2,14 @@
     x:Class="MaterialDesignDemo.Expander"
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
     xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
     xmlns:materialDesign="http://materialdesigninxaml.net/winfx/xaml/themes"
     xmlns:smtx="clr-namespace:ShowMeTheXAML;assembly=ShowMeTheXAML"
-    mc:Ignorable="d" 
+    mc:Ignorable="d"
     d:DesignHeight="400"
     d:DesignWidth="600">
-    
+
     <UserControl.Resources>
         <Style TargetType="{x:Type TextBlock}" x:Key="HorizontalExpanderContentTextBlock">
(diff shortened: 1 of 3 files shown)


Keboo (commented):
This mostly works, however, if a user is using the keyboard, they are still able to tab into the header content (or the chevron) and use the spacebar to toggle the expander.

Thinking out loud here. It probably would make more sense to conditionally bind the IsChecked property of the ToggleButton base on the attached property. That way, it does not matter how the user interacts with the ToggleButton, it will only be bound to the IsExpanded property when the attached property is true.
In addition, I think it would make sense to hide the chevron if the user cannot interact with it. I suspect that leaving it visible will cause more confusion.

One final suggestion, can we rename the attached property to better reflect its purpose. Perhaps something like `IsUserInteractionEnabled`

binarycow:
> In addition, I think it would make sense to hide the chevron if the user cannot interact with it. I suspect that leaving it visible will cause more confusion.

In the general case, you probably are correct.  But, in my use case, I want the chevron to be visible.

My use case is an accordion control (each `ItemContainer` is an `Expander`).  

- There will always be exactly one expanded item
- I don't want the user to be able to collapse the currently expanded item
- I want the chevron to be visible on all items, even the currently expanded item.

_But..._ you may be right. 

Additionally, I realized that maybe `ToggleButton` isn't the right control in my use case (for the `Accordion`).  [Microsoft docs](https://docs.microsoft.com/en-us/dotnet/api/system.windows.controls.radiobutton?view=windowsdesktop-6.0) say that a `RadioButton` "represents a button that can be selected, but not cleared, by a user.".  This is _exactly_ the behavior I'm looking for on my accordion control.

So - ultimately, I think, for my use case, to get the right control, I'm gonna end up reimplementing the `ControlTemplate` _anyway_.

Issue #2255, opened 2021-02-19
Breaking 5.0.0 Changes


Keboo:
There is an existing `5.0.0` branch where these changes are being made. Once we are a little closer to the release this branch will be merged into `master`. Anyone is welcome to pick up tasks here and PR into the `5.0.0` branch.

This is a list of proposed breaking changes for the 5.0.0 release. Feedback welcome!

- [x] Remove `DrawerHostOpenMode.Model` (Added in 4.0.0, Obsolete in 4.1.0)
- [x] Remove `MaterialDesignScrollBarThumbVertical` and `MaterialDesignScrollBarThumbHorizontal` replace with `MaterialDesignScrollBarThumb`
- [x] Consider making [recommended theme color classes](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/tree/master/MaterialDesignColors.Wpf/Recommended) static. 
  - Currently not possible because they implement the `ISwatch` interface.
- [x] Update ListBoxItem to use theme brush for hover color rather than deriving from the Foreground.
- [x] Update MenuItem to use theme brush for hover color rather than deriving from the Foreground.
- [x] There is a both usage of "Assist" and "Extensions" used. Consider standardizing on naming
- [x] Remove `MaterialDesignRaisedAccentButton` in favor of `MaterialDesignRaisedSecondaryButton`
- [x] Remove `MaterialDesignFlatAccentButton` in favor of `MaterialDesignFlatSecondaryButton`
- [x] Remove `MaterialDesignFlatAccentBgButton` in favor of `MaterialDesignFlatSecondaryMidBgButton`
- [x] Remove `MaterialDesignFloatingActionMiniAccentButton` in favor of `MaterialDesignFloatingActionMiniSecondaryButton`
- [x] Remove `MaterialDesignFloatingActionAccentButton` in favor of `MaterialDesignFloatingActionSecondaryButton`
- [x] Remove `ComboBoxPopupPlacement`
- [x] Remove obsolete members from `ComboBoxPopup`
- [x] Remove obsolete templates `PopupContentUpTemplate` and `PopupContentDownTemplate`
- [x] Remove obsolete brushes #2435 
- [x] Replace usages of "Accent" with "Secondary"
- [x] Remove "MaterialDesignNavigatilRailTabControl" it is replaced by "MaterialDesignNavigationRailTabControl"
- [x] Remove RatingBarButton.IsWithinSelectedValue property
[...]

PR #2762, opened 2022-07-05
Refactor fix for 2689 to use attached properties

nicolaihenriksen:
Refactoring of previous fix for #2689

Added attached properties for UniformCornerRadius and Style. These properties are inherited and can thus be used to manipulate a Card located inside of another UIElement (eg. the Flipper). I decided to put them in a CardAssist file to align with the other attached properties in the toolkit. It could be moved to the Card type if that is more desirable.

Added more UI tests for this as well. Also fixed a bug in the original elevated card style (UniformCornerRadius binding) caught by these tests.

### DIFF
--- a/MaterialDesignThemes.Wpf/CardAssist.cs
+++ b/MaterialDesignThemes.Wpf/CardAssist.cs
@@ -1,0 +1,28 @@
+﻿namespace MaterialDesignThemes.Wpf;
+
+public static class CardAssist
+{
+    #region AttachedProperty : UniformCornerRadiusProperty
+    /// <summary>
+    /// Controls the (uniform) corner radius of the contained card
+    /// </summary>
+    public static readonly DependencyProperty UniformCornerRadiusProperty
+        = DependencyProperty.RegisterAttached("UniformCornerRadius", typeof(double), typeof(CardAssist),
+            new FrameworkPropertyMetadata(default(double), FrameworkPropertyMetadataOptions.AffectsRender | FrameworkPropertyMetadataOptions.Inherits));
+
+    public static void SetUniformCornerRadius(DependencyObject element, double value) => element.SetValue(UniformCornerRadiusProperty, value);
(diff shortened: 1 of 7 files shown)


MaterialDesignThemes.Wpf/CardAssist.cs
```diff
@@ -0,0 +1,34 @@
+﻿namespace MaterialDesignThemes.Wpf;
+
+public static class CardAssist
```
Keboo, line 3: I would rename this to `FlipperAssist` since the consumption is designed for flippers not cards, even though the target of the property is a card.

MaterialDesignThemes.Wpf/Flipper.cs
```diff
@@ -143,18 +143,6 @@ public bool IsFlipped
             instance.RaiseEvent(args);
         }
 
-        public static readonly DependencyProperty UniformCornerRadiusProperty = DependencyProperty.Register(
-            nameof(UniformCornerRadius), typeof(double), typeof(Flipper), new PropertyMetadata(default(double)));
-
-        /// <summary>
-        /// Gets or sets the (uniform) corner radius applied the the <see cref="Flipper"/> when the MaterialDesignCardFlipper style is applied.
-        /// </summary>
-        public double UniformCornerRadius
```
Keboo, line 152: Rather than delete this, lets mark it with `[Obsolete]` and I can add it to the [list of breaking changes for the 5.x release](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2255).

Keboo (changes requested):
Looks pretty close, just a couple changes.

MaterialDesignThemes.Wpf/CardAssist.cs
nicolaihenriksen: I agree, I'll rename it.

MaterialDesignThemes.Wpf/Flipper.cs
Keboo: O opps, this was my mistake.

Issue #2073, opened 2020-09-30
Rating Bar support fractional values


devriesb13:
Hi there,

Can the rating bar handle displaying an average rating value (doubles) but let a user vote (integers) similar to how the SharePoint list items work?

Cheers Ben

jespersh:
You mean like this?
![image](https://user-images.githubusercontent.com/1004198/94658390-06467c80-0303-11eb-88a5-a14637f9aba2.png)

devriesb13:
Yes exactly, see my gif attachment for SharePoint version. Ignore the number 2, this is just a number of how many users have voted.

My thoughts around this would be. By default you see the average rating bound to some (double) value. When you hover over the rating bar, it turns on  eg "rating mode", then you can select a rating (integer) you want to give.

![2020-10-01_8-47-16](https://user-images.githubusercontent.com/10214916/94747760-0f971e00-03c3-11eb-990d-fcaa0dd81f50.gif)

jespersh:
To answer your question then, that's a no sadly:
https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/1873f0d58929ededc8724bac3a68fa260dfd498a/MaterialDesignThemes.Wpf/RatingBar.cs#L65-L69

But I don't see why someone couldn't modify this to support halves

Keboo:
Currently there is not support for fractional amounts. If someone would like to try and tackle this, I would be happy to accept a PR for it.

Christiankoo:
Can i take this issue, please?

Keboo:
@Christiankoo go for it. Let me know if you have any questions

stojy:
hi @Christiankoo did you make any progress on fractional values?

stojy:
After taking a quick look through the RatingBar code, perhaps a better solution might be a solution on the Slider control instead?  
The basic functionality I believe is already there with the left/right slider and discrete steps functionality.  You'd lose the funky animation of course, but I suspect this isn't essential.

PR #2764, opened 2022-07-06
Allow "half" rating selection

nicolaihenriksen:
@Keboo Thanks for the feedback on this PR. I have reworked this PR based on your comments, and as such the changes are a bit more extensive; so grab a Mountain Dew before you read any further :-)

The PR now covers this:
- Coercing Min/Max values. I don't see this as a breaking change since wrong values would yield an unusable rating bar.
- Added opt-in functionality for previewing a value (before selecting it). Inspiration taken from `Slider` but works on hover in this case.
- Added opt-in functionality for fractional values. This is done by adding a `ValueIncrements` property (`double`) which can have a value between `double.Epsilon` and 1.0. A value of 1.0 - which is the default - effectively means that fractional values are disabled. Specifying a value where less than 1.0 where 1 is not a multiple of said value (e.g. 0.75) works, but is not really useful. Perhaps the coercion of `ValueIncrements` should ensure that the 1 is a multiple of the value being set. Thoughts?
- The two opt-in options are independent of each other; ie. one can be enabled without the other and vice versa.
- Added a 3 purpose-built `IMultiValueConveters`, and placed them as nested classes in the `RatingBar` class because they are specifically made for that control. Let me know if you normally do it in a different manner.
- Added unit tests for converters
- Implementation respects `RatingBar.Orientation` and `RatingBar.FlowDirection`

**Funny note:**
The original `RatingBar` does not coerce its value, and thus you can set anything for the `Value` property. This is probably also needed, because with Min=1 and Max=5 you can effectively not (with the mouse at least) set a rating of "0 stars".
With fractional values enabled, I had to change this a bit, so in cases like that you would typically set Min=0 (instead of 1), which then means that you can partially select the first star; i.e. set a value between 0 and 1. In this case you **can** give a rating of "0 stars" because the coercion will eventually give you the value 0 if your mouse is close enough to the left edge of the first `RatingBarButton`.

Screen captures (GIFs) of the new opt-in features can be seen below:

2022-07-13_21h52_49

2022-07-13_21h54_30

2022-07-13_21h56_43

Original PR text:
---

Partial fix for #2073

**IMPORTANT:**
Correct theming colors need to be applied in MaterialDesignTheme.RatingBar.xaml resources defined in the top of the file. @Keboo do you have good suggestions on how to get these colors (NOT brushes). And note that I also need some colors with a specific alpha-channel applied to them (matching 26% opacity). I hope something is already available to get this, otherwise a custom `MarkupExtension` could be an option.

**Changes**:
Changes `RatingBar.Value` to a double which is coerced into a multiple of 0.5 and forced to stay within the Min/Max bounds of the `RatingBar`. The "half selected" visual appearance is accomplished by using a (gradient) `Brush` on `TextBlock.Foreground` which is either fully opaque (selected), a mix of fully opaque and semi-transparent (half-way selected), or completely semi-transparent (not selected). These gradients/brushes are where I need to access theming colors.

Implementation respects `RatingBar.Orientation`, however I am not entirely sure (ie. I haven't tested) whether it works in RTL layout.

image


MaterialDesignThemes.Wpf.Tests/RatingBarTests.cs
```diff
@@ -0,0 +1,28 @@
+﻿using Xunit;
+
+namespace MaterialDesignThemes.Wpf.Tests;
+
+public class RatingBarTests
+{
+    [StaTheory]
+    [InlineData(-5, 0.0)]
```
Keboo, line 8: Should the expected value here be 1.0 to match the Min?

Keboo (commented):
This is great work, thank you for taking it on. A couple of suggestions.

First, I think we should make it an opt-in option for users clicking on the rating bar and getting fractional values. I think one way to accomplish this would be a property similar to the `TickFrequency` of the slider where the developer can indicate the level of precision that is desired.

As for getting the colors and setting it up. I think we might be able to support more than just 0.5 by leveraging some attached properties and a custom `MultiValueConverter`.

Here is a small sample I put together using a slider.

This XAML is using the Tag property, but we could create any number of internal attached/dependency properties to support this.
```xaml
<Slider
    Minimum="0"
    Maximum="50"
    Value="35"
    Tag="{DynamicResource PrimaryHueMidBrush}">
    <Slider.Foreground>
        <MultiBinding Converter="{StaticResource LinearBackgroundConverter}">
            <Binding Path="Tag" RelativeSource="{RelativeSource Self}"/>
            <Binding Path="Value" RelativeSource="{RelativeSource Self}"/>
        </MultiBinding>
    </Slider.Foreground>
</Slider>
```
Then the converter would look something like this:
```c#
public object? Convert(object?[]? values, Type targetType, object parameter, CultureInfo culture)
{
    //TODO add orientation as another parameter (might also need min/max to derive percentage)
    if (values?.Length == 2 &&
        values[0] is SolidColorBrush brush &&
        values[1] is double value)
    {
        value /= 50;
        return new LinearGradientBrush()
        {
            StartPoint = new Point(0, 0.5),
            EndPoint = new Point(1, 0.5),
            GradientStops = new()
            {
                new GradientStop { Color = Colors.Fuchsia, Offset = value },
                new GradientStop { Color = brush.Color, Offset = value }
            }
        };
    }
    return null;
}
```
I think if we do something like this, it could also support more than just half values. Thoughts?

nicolaihenriksen:
> I think if we do something like this, it could also support more than just half values. Thoughts?

Great idea with the `MultiValueConverter` and the trick of injecting the current brush via the `Tag` (or an attached property). That should work nicely and will - as you say - allow for "any" step size (ie. `TickFrequency`).

My only concern about opening up for `TickFrequency` being very small increments, is how this is "supported" in the UI (from the users perspective). Currently there is no visual feedback on hover, so lets say we use 0.1 increments, it for one can be really difficult to actually hit those with the mouse, but I think we should definitely add a hover effect which then "previews" which value you're about to apply (perhaps both with coloring in the `RatingBarButtons` up to the desired value, but maybe also a numeric indication of the value itself? Thoughts?

MaterialDesignThemes.Wpf.Tests/RatingBarTests.cs
nicolaihenriksen: Well hidden in my commit comments, I left this little comment:

    "It is worth noting that valid values are actually values between Min-1 and Max (in multiples of 0.5)"

I agree this is somewhat strange, and I will work on changing the behavior.

It should however be noted that the implementation before this PR **did not** have any coercion (which is strange for a "range bound" control IMO), and thus allowed you to set **any** value.

In any case, the first button represents the `Min` value today, and that is a little strange because that actually means you should not be able to set a rating less than 1 star when `Min`=1; why should it not be allowed to set 0 stars (although difficult to do with the mouse!)? Anyway, I will see if I can rework this into something a little more intuitive while preserving back-compat.

Keboo:
> My only concern about opening up for `TickFrequency` being very small increments, is how this is "supported" in the UI (from the users perspective). Currently there is no visual feedback on hover, so lets say we use 0.1 increments, it for one can be really difficult to actually hit those with the mouse, but I think we should definitely add a hover effect which then "previews" which value you're about to apply (perhaps both with coloring in the `RatingBarButtons` up to the desired value, but maybe also a numeric indication of the value itself? Thoughts?

I agree with the "supported" problem. I suspect that most consumers that use small increments will likely also set them to readonly (and I suspect bind the value to a TextBlock or similar to show a numeric value). 

I do like the idea of allowing the user to see the bound value on hover or press. I think some interaction that is similar to how the slider does it might be nice. Though I would say that displaying the value should be an opt-in behavior.

nicolaihenriksen:
> > My only concern about opening up for `TickFrequency` being very small increments, is how this is "supported" in the UI (from the users perspective). Currently there is no visual feedback on hover, so lets say we use 0.1 increments, it for one can be really difficult to actually hit those with the mouse, but I think we should definitely add a hover effect which then "previews" which value you're about to apply (perhaps both with coloring in the `RatingBarButtons` up to the desired value, but maybe also a numeric indication of the value itself? Thoughts?
> 
> I agree with the "supported" problem. I suspect that most consumers that use small increments will likely also set them to readonly (and I suspect bind the value to a TextBlock or similar to show a numeric value).
> 
> I do like the idea of allowing the user to see the bound value on hover or press. I think some interaction that is similar to how the slider does it might be nice. Though I would say that displaying the value should be an opt-in behavior.

I have reworked the PR a bit to include opt-in features for fractional values (freely set), and preview of value being selected. See the updated PR description for details.

Keboo (approved):
This is really awesome work! Thank you

stojy:
Not sure if this is the right place to ask, but do you know when 4.6.0 will be released so we can start using the 1/2 rating selection?

Issue #1181, opened 2019-01-30
Setting DialogHost.IsOpen to false does not trigger DialogClosing event


pfeigl:
The DialogClosing event is emitted in the `Close` Method, however this method is never called, when the dialog is closed via the IsOpen property.

Thus the `DialogClosing` event is never emitted.

Is this expected behavior or a bug?

Keboo:
@pfeigl somewhat expected. The `DialogClosing` event [allows for cancelation](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MaterialDesignThemes.Wpf/DialogClosingEventArgs.cs#L35-L43). If it were raised when `IsOpen` change that could then mean that when you set the property to false that if you immediately read it again, it would still be false. 
```C#
DialogHost dialogHost = ...;
dialogHost.IsOpen = false;
bool isOpen = dialogHost.IsOpen;
//isOpen could be true at this point if closing was cancelled.
```

With that said, I do agree that it is a bit confusing. The DialogHost is one area where I think the API could use some work. I am open to suggestions on how it should work.

pfeigl:
I need to dig a little more into the code to give some good feedback. Some things that come into my mind:

* Have an `IsCancelable` property on the closing event and throw if a non cancelable event is beeing canceled
* Keep the ClosingEvent as is, put meaningful documentation (comment) into the code but additionally emit an DialogClosed event which is always thrown. Except from stopping the closing, for most situations it's probably irrelevant whether people register on Closing or Closed, they probably just want to do something after the dialog is "gone".

Keboo:
I think adding in a new event for `DialogClosed` makes the most sense here. This event should always be raised.

Keboo:
Resolved by #2767

PR #2767, opened 2022-07-07
Added DialogHost.DialogClosed event

nicolaihenriksen:
Fix for #1181 

Adds DialogClosing event to DialogHost with the same hooks as the DialogOpening and DialogClosing events.

Works the same way as DialogClosing, except it cannot be cancelled, and it is only fired once IsOpen changes to False.

Updated samples to include the DialogClosing event hooks and also updated/added some unit tests to ensure the event is fired correctly.

**IMPORTANT**:
There is a slight risk that the `DialogHost.Show(...)` and `DialogHostEx.ShowDialog(...)` overloads is a breaking change. This could potentially cause ambiguity in existing code using the toolkit, or even worse it could end up calling another overload than the intended one. It might be better to remove the overloads with fewer parameters and thus only extending with the new overload taking the extra "closed handler" parameter?

### DIFF
--- a/MainDemo.Wpf/Dialogs.xaml
+++ b/MainDemo.Wpf/Dialogs.xaml
@@ -64,6 +64,7 @@
                 Grid.Row="1">
                 <materialDesign:DialogHost
                     DialogClosing="Sample1_DialogHost_OnDialogClosing"
+                    DialogClosed="Sample1_DialogHost_OnDialogClosed"
                     DialogTheme="Inherit">
                     <materialDesign:DialogHost.DialogContent>
                         <StackPanel Margin="16">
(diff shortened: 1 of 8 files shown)


Keboo (commented):
This is really great work! I think you are correct about the potential ambiguity in the calling code. I think for now, let remove the overloads that might conflict (specifically the ones that only differ by closing vs closed). Given how old the issue is, I suspect people who want the feature will be happy to use the bigger overload and specify all parameters.

nicolaihenriksen:
> This is really great work! I think you are correct about the potential ambiguity in the calling code. I think for now, let remove the overloads that might conflict (specifically the ones that only differ by closing vs closed). Given how old the issue is, I suspect people who want the feature will be happy to use the bigger overload and specify all parameters.

Sounds reasonable. I'll change it so only a single new overload is added; the one with the additional closed handler in it.

Keboo (approved):

PR #2778, opened 2022-07-13
Fix TabControl style

### LINKED ISSUE #2660: TabControl - DefaultStyle and FilledStyle are the same?
The default style and filled style of the `TabControl` look the same in the demo app and when used in an application. I don't think this is the expected behavior?

From version 4.4.0

image

From version 4.5.0

image

binarycow:
Fixes #2660


Keboo:
Thanks you. Closing this in favor of #2779 which covers a few more use cases.

Issue #2666, opened 2022-05-03
TabControl - Option to get v4.4.0 default look in v4.5.0


avanbremen:
First of all, thanks for creating an awesome toolkit!

I noticed that the default look of the TabControl changed from version 4.4.0 to version 4.5.0. So far I did not find a way to replicate the 'old' design in the latest version. Setting `materialDesign:ColorZoneAssist.Background` and `Foreground` only affects the full-width fill. Setting `wpf:ColorZoneAssist.Mode` to for example `Standard` will result in a black text and underline. There is no Mode to replicate the v4.4.0 look.

As I already released my application with the 'old' v.4.4.0 design, it would be awesome if I could replicate this design in v.4.5.0 without overriding the entire TabControl Style. For example by being able to set a `HeaderForeground`.

![mdix-tabcontrol-v4-4-0](https://user-images.githubusercontent.com/25454736/166520627-991ba3a8-350f-471d-9f12-6084dc6243a0.jpg)
_Default TabControl in v4.4.0_

![mdix-tabcontrol-v4-5-0](https://user-images.githubusercontent.com/25454736/166520657-90afaca7-9889-480b-8e3c-c7fa2dec09e6.jpg)
_Default TabControl in v4.5.0_

Thanks and all the best,
Arjan

mikechristiansenvae:
FYI: I think this is a duplicate of #2660

PR #2779, opened 2022-07-14
Fix TabControl styles

Xaalek:
Fix https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2666 and https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2660

This task was painful because of the ColorZone. I think it would have been easier to use attached properties (for example: Background, Foreground and ShadowDepth).
In the end, I kept the ColorZone system and managed to fix the problems. I also modified a few things. Here are the changes:
- The background color was applied to the TabControl header and the TabItem header. I corrected this so that the color is only applied on the TabControl header.

- Concerning the style, I remake the default style (not filled)

- As you can see in the specs, the colored foreground in only apply on the focused item.
image
It was not the case in MDIX, so I fixed that.

- I added the possibility to use custom colors

- I reduced the shadow depth

- I added examples in the demo

# Final result
[Result](https://www.youtube.com/watch?v=YVzpWNdc-TI "Result")

### DIFF
--- a/MainDemo.Wpf/Tabs.xaml
+++ b/MainDemo.Wpf/Tabs.xaml
@@ -433,7 +433,7 @@
                             <TextBlock Margin="8" Text="Not filled example tab 1" />
                         </TabItem>
                         <TabItem Header="TAB 2">
-                            <TextBlock Margin="8" Text="No filled example tab 2" />
+                            <TextBlock Margin="8" Text="Not filled example tab 2" />
                         </TabItem>
                     </TabControl>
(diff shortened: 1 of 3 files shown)


Xaalek:
Thanks for your help and explanations in your last stream!

mikechristiansenvae:
For what it's worth, I would prefer having separate properties for `TabItemHeaderForeground`, `TabItemHeaderBackground`, etc.

Issue #2713, opened 2022-05-27
Editable ComboBox behaviour broken since version 4.4


hfann:
### Bug explanation

When a `IsEditable = True` combo box with text is clicked, the drop down opens instead of the underlying `TextBox` gaining focus. This behaviour is broken since version 4.4. Version 4.3 does have the correct behaviour.

This can be reproduced using the demo app. 
1. Open the ComboBoxes section.
2. Click on any editable ComboBox and either enter some text or select an item.
3. Focus on another control.
4. Click that ComboBox again.
5. The desired behaviour (like Microsoft standard ComboBox) is to gain focus for editing. However, in version 4.5, it will display the drop down list. In order to edit the content, you must use cursor key while .

### Version

4.5.0

nicolaihenriksen:
It seems @Keboo fixed a similar issue (#2277) by moving the `ToggleButton` down in the Z-order. However, those changes seem to have been reverted by @MichelMichels as part of this PR #2525 (particularly this commit 2c75cbcda1a3a12b2df6a2c5eb36c52cf36316c2). Nearly all other changes seem to be whitespace adjustments.

I will play around with the ordering of the elements to see if I can land it such that both the rounded corners and the editable cursor stuff works at the same time.

nicolaihenriksen:
Oh I see that @Keboo already has a branch for this, but it seems unfinished. I will add another branch with my (similar) changes, then you can pick the one you like.

PR #2789, opened 2022-07-16
(editable) ComboBox: Moved ToggleButton down in Z-Order

nicolaihenriksen:
Fix for #2713 

Sorry @Keboo I did not see you already had a branch where you are working on this. Your changes seem similar.

I added a UI test that initially failed, and then I moved the `ToggleButton` down in the Z-Order which allows the editable `TextBox` to capture the mouse interaction and grab focus. After this change, the UI test passes.

I also made sure that the rounded corners where still preserved. There was a fix to add these rounded corners which actually reintroduced this issue (see my comments on issue #2713). So I just wanted to make sure the changes in PR #2525 still worked as expected; although I only verified this visually - not with a UI test.


nicolaihenriksen:
It seems the `MaterialDesignThemes.UITests.WPF.DrawerHosts.DialogHostTests.DrawerHost_OpenAndClose_RaisesEvents()` UI test failed, but I believe my changes are unrelated so that is a bit strange...

Keboo (commented):
This is great, thank you!

Issue #1293, opened 2019-06-21
Remove border around the selected cell in DataGrid


DenZuck:
Is it possible to remove the black border around the selected cell in DataGrid control ?

<img width="684" alt="image" src="https://user-images.githubusercontent.com/7376037/59915651-6d496600-941d-11e9-91b9-b3ab875f2c71.png">

AlbertAlonso:
I believe the only solution is to override the DataCell Style, as the border is hardcoded in the template.

Xaalek:
I see that there are a lot of thumbs up. It might be interesting to create an attached property to display or not the border cell.

Keboo:
I think an attached property makes a lot of sense for this!

I would propose that the attached property should be of type `Brush` with a default of `null`. We will want to keep the default of `MaterialDesignTextBoxBorder` but allow for people to set it to something like `Transparent` as a means of disabling it or changing its color. I would expect that this attached property would need to be set on the `DataGrid` itself.
This likely will mean expanding out the current Trigger into two triggers (one for when the attached property is null, and one where it isn't), or having one `MultiDataTrigger` with some bindings.


In any case I would expect the following cases to work:
1. If the user does nothing, they should get the border with the existing `MaterialDesignTextBoxBorder` brush
2. If the user sets the attached property on the DataGrid, they get whatever brush is set on the attached property. 

In the meantime, there is sort of a workaround (this same trick works for most styles). If you create a `Brush`, put it in the `DataGrid`'s resources, and give it a key of `MaterialDesignTextBoxBorder`, it will override the value for everything within the `DataGrid`. The obvious drawback here is that it may override more than you want (but that will depend on the types of columns being used). For example, here is the demo app using `Fuschia`. You can see it override the border color as well as some elements in the `ComboBoxColumn`.
```xaml
<DataGrid.Resources>
    <SolidColorBrush x:Key="MaterialDesignTextBoxBorder" Color="Fuchsia" />
</DataGrid.Resources>
```

![image](https://user-images.githubusercontent.com/952248/167533660-464351c6-f9d8-45d8-a63f-61e93e7f2a5d.png)

its-soheyl:
'& .MuiDataGrid-cell:focus': {
            outline: ' none'
          },

PR #2800, opened 2022-07-28
Add SelectedCellBorderBrush attached property

Xaalek:
As suggested here https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/1293, I implemented an attached property to easily change the selected cell border color.
I also added an example in the demo app:

image


MaterialDesignThemes.Wpf/DataGridAssist.cs
```diff
@@ -136,6 +137,17 @@ public static void SetCellPadding(DataGrid element, Thickness value)
             => element.SetValue(CellPaddingProperty, value);
         #endregion
 
+        #region AttachedProperty : SelectedCellBorderBrushProperty
+        public static readonly DependencyProperty SelectedCellBorderBrushProperty
+            = DependencyProperty.RegisterAttached("SelectedCellBorderBrush", typeof(Brush), typeof(DataGridAssist),
+                new FrameworkPropertyMetadata(null, FrameworkPropertyMetadataOptions.Inherits));
```
Keboo, line 143: It doesn't look like we need this property to inherit down, so there is a small performance gain by removing it
```suggestion
                new PropertyMetadata(null));
```

Xaalek:
Indeed you are right. I have corrected this.

Keboo (approved):

Issue #2729, opened 2022-06-01
Broken Flipper


ElieTaillard:
### Bug explanation

The flipper seems broken in the 4.5.0.

Here are the bugs I noticed:

- The Flipper has lost its "Card" appearance: missing elevation, missing border
- The Filpper has a strange behavior in the demo app like mention here: https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2577
- The background Flipper does not change color when changing theme (dark and light) like mention here: https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/694

These bugs can be observed in the demo application.

![image](https://user-images.githubusercontent.com/54487782/171449704-8c54ad6c-ee69-43d9-98cf-221d56c648b4.png)

![image](https://user-images.githubusercontent.com/54487782/171449815-169dc8bc-d783-4971-8b08-ed58909e66f1.png)

https://user-images.githubusercontent.com/54487782/171450322-950edc9f-8e27-47db-839d-d8b3143ff5de.mp4

It would be nice to take a look at these different problems because currently the Flipper is not very usable.

I quickly checked the code but I didn't see anything shocking. The only thing I noticed were the following: 
- The Flipper style is not dynamic in the demo app https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MainDemo.Wpf/Cards.xaml#L341

### Version

4.5.0

Erapchu:
Seems like the same as

https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2577

JeanPhilippeLux:
Hi, any idea when this will be fixed? Flipper is a nice feature but the color problem make controls not readable sometimes :(.

Keboo:
@JeanPhilippeLux not sure I have taken a couple stabs at fixing it but it is difficult with the flip animation and control sizing with the WPF 3d controls.

ElieTaillard:
This issue will be fixed with https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/pull/3431

PR #2801, opened 2022-07-28
Fix default elevation of Card Flipper

nicolaihenriksen:
Partial fix for #2729

This fixes the broken default elevation of the (Card) Flipper. It was a simple matter of a default elevation not being set on the MaterialDesignCardFlipper style.

I added a UI test to ensure this default is applied, but it fails because of the static constructor in FlipperAssist. I was unable to get this to work and may need some input on how to get it to locate the resource with the Uri when executing in the MaterialDesignThemes.UITests context.


Keboo:
Thank you. I fixed the issue by moving out the static loading of the resources into a separate class. This effectively defers loading of the resources until `GetDropShadow` is invoked

PR #2802, opened 2022-07-31
Restored inactive calendar buttons

### LINKED ISSUE #2799: Missing years in the calendar
### Bug explanation

As you can see on the screenshots below, two years are missing in the calendar

image

image

If we look at the calendar with the default style, we notice that there are two more years.

image

### Version

4.5.0

jizc:
Fixes #2799

This PR restores the missing calendar buttons. Both years that are not within the selected decade and days that are not within the selected month are restored by this fix. WPF refers to these buttons as Inactive, but are meant to be clickable.

The cause of this "bug" can be traced all the way back to #142, where a choice was made to hide the inactive buttons because they can be confused with [Blackout Dates](https://docs.microsoft.com/en-us/dotnet/api/system.windows.controls.calendar.blackoutdates). I opted to instead use different levels of foreground opacity, where the blackout dates appear disabled. To make the blackout dates behave more like disabled elements I also set `Cursor` to `Arrow` and hid the highlight background. In my opinion these changes make inactive and blackout dates different enough, especially if we can assume that the blackout dates feature is not used very often.

An option would be to hide the inactive buttons on the month view, and only show them in the year view. Personally I like instead opting for the default WPF behavior. The [MD specs](https://material.io/components/date-pickers) seem to be inconsistent on this, as some screenshots show days in neighboring months and others don't.

CalendarYears CalendarDays

### Notes
- I switched to using template triggers instead of visual states for the text opacity, to fix an issue where blackout dates would in a few cases get the wrong opacity
- Changed trigger property from `IsFocused` to `HasSelectedDays` to fix issue where the wrong months text would be changed
- I fixed a `DatePicker` test that failed on my PC, due to my local culture settings using `dd.MM.yyyy` instead of `MM.dd.yyyy`, it now works regardless of local settings

### DIFF
--- a/MaterialDesignThemes.UITests/WPF/ComboBoxes/ComboBoxTests.cs
+++ b/MaterialDesignThemes.UITests/WPF/ComboBoxes/ComboBoxTests.cs
@@ -114,7 +114,7 @@ public class ComboBoxTests : TestBase
 
         IVisualElement<ContextMenu>? contextMenu = await comboBox.GetContextMenu();
         Assert.True(contextMenu is not null, "No context menu set on the ComboBox");
-
+        await Wait.For(async () => await contextMenu!.GetIsVisible());
         await AssertMenu("Cut");
         await AssertMenu("Copy");
         await AssertMenu("Paste");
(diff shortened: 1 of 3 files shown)


jizc:
The `OnEditableComboBox_WithDefaultContextMenu_ShowsCutCopyPaste` test failed. It doesn't fail when I run it locally, and my changes should not affect it.

Issue #2772, opened 2022-07-12
Add a materialdesign:dialoghost overlay CornerRadius attribute


wj8400684:
![image](https://user-images.githubusercontent.com/39239954/178483468-bffea392-f5ae-438f-9e9e-1e69d129d857.png)

Xaalek:
Hello, please make an effort on the writing

wj8400684:
Add an overlay corner radius attribute to dialoghost！

wj8400684:
When the dialog box opens, the border can display rounded corners！

PR #2803, opened 2022-08-05
Added CornerRadius to DialogHost

nicolaihenriksen:
Fix for #2772 

Introduces a `CornerRadius` DP on `DialogHost` which is then used to round the corners of the `PART_ContentCoverGrid` (ie. the grid which shows a dimmed overlay covering the area that should be disabled while the dialog is open).

In order to support a `CornerRadius`, I introduced a `Border`. Ideally I guess the `PART_ContentCoverGrid` could have been **changed** to a `Border` but that could potentially break back-compat in case someone is doing their own template and providing a `Grid` named `PART_ContentCoverGrid`. To (hopefully) avoid introducing a breaking change, I simple placed a `Border` inside of the `PART_ContentCoverGrid`, and then use the `OpacityMask` on the `PART_ContentCoverGrid` to enforce the masking out of content that is not within the bounds of the contained `Border`.

My initial approach was to wrap `PART_ContentCoverGrid` in a `Border`, and the set `ClipToBounds` on the `Border`, but that **does not work**; the `PART_ContentCoverGrid` is still allowed to render its background in the corner area which is "trimmed away" by the `Border`, and thus it basically has no effect.

I added a simple UI test just to verify that the DP is correctly applied to the newly introduced `Border` (named `ContentCoverBorder`).

**UPDATE**
I did not actually update the demo application to showcase this, because I didn't really think it made a lot of sense in the uses that are showcased. But to test it out, you can just set the DP on any of the `DialogHost` instances in there.

### DIFF
--- a/MaterialDesignThemes.UITests/WPF/DialogHosts/DialogHostTests.cs
+++ b/MaterialDesignThemes.UITests/WPF/DialogHosts/DialogHostTests.cs
@@ -26,7 +26,7 @@ namespace MaterialDesignThemes.UITests.WPF.DialogHosts
             await using var recorder = new TestRecorder(App);
 
             IVisualElement dialogHost = await LoadUserControl<WithCounter>();
-            var overlay = await dialogHost.GetElement<Border>("ContentCoverBorder");
+            var overlay = await dialogHost.GetElement<Grid>("PART_ContentCoverGrid");
 
             var resultTextBlock = await dialogHost.GetElement<TextBlock>("ResultTextBlock");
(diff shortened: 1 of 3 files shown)


Keboo:
This looks great. Would you mind adding one more UI test that uses the embedded dialog host style? Just to ensure that both styles leverage the property.

Keboo (approved):

nicolaihenriksen:
> This looks great. Would you mind adding one more UI test that uses the embedded dialog host style? Just to ensure that both styles leverage the property.

Sure, no problem 👍

Keboo (approved):

Issue #2737, opened 2022-06-08
Changing border thickness of "MaterialDesignOutlinedDatePicker"


washburnit:
### Bug explanation

HI,

how could I change the border thickness of the control in the subject?
I've set the control as follows:

<DatePicker Name="dtLavo" Width="180" Style="{StaticResource MaterialDesignOutlinedDatePicker}" HorizontalAlignment="Left" Margin="20" mdix:HintAssist.Hint="Data Creazione" FontFamily="Century Gothic" FontSize="15" BorderThickness="2"/>

but is seems that it has no effect (probably I'm not doing in the correct way).
Thank you, regards.

Roberto

### Version

4.5.0

Xaalek:
I just checked and indeed, the feature is missing for `MaterialDesignOutlinedTimePicker` and `MaterialDesignOutlinedDatePicker`.

washburnit:
Are you going to include that feature in a future update?

Xaalek:
Maybe, I don't plan to work on it right now. I think it's an easy fix so if you ever want to work on it would be cool.

PR #2805, opened 2022-08-05
Added BorderThickness attached DP for outlined date/time pickers

nicolaihenriksen:
Fix for #2737

This PR introduces 2 new assists types (`DatePickerAssist` and `TimePickerAssist`) which both contain attached DPs which can be used to set the inactive and active `BorderThickness` on the `DatePicker`/`TimePicker` using the outlined styles (`MaterialDesignOutlinedDatePicker`/`MaterialDesignOutlinedTimePicker`). "Inactive" is when there is no mouse/keyboard interaction with the element, and "Active" is when the element is either hovered or has keyboard focus.

For the `TimePicker`, I could have just added the new attached DPs directly as normal DPs on the `TimePicker` type, but since the properties only apply for the outlined style, I thought it was better to keep them in a `TimePickerAssist` type.

I updated the demo app to include some samples of this because using `Thickness` values that vary from the default values may require the caller to also set the `HintAssist.FloatingOffset` in order for the hint to be correctly placed.

The `IMultiValueConverter` that I added is needed to calculate a "margin diff" which ensures the UI does not "jump" between the inactive/active states. This was previously hardcoded to -1 because the of the defaults being inactive=1 and active=2.

**NOTE**: In the UI tests I have left a **TODO** because I would have liked to be able to mark a `ValidationError` on the contained `TextBox` without too much overhead. Perhaps this is something that could be added as a feature in XAMLTest? For now, the tests include an actual `ValidationRule` and the value is changed during the test to force the validation error. I added a [PR](https://github.com/Keboo/XAMLTest/pull/83) with proof-of-concept for this in the XAMLTest project.

### DIFF
--- a/MaterialDesignThemes.UITests/WPF/DatePickers/DatePickerTests.cs
+++ b/MaterialDesignThemes.UITests/WPF/DatePickers/DatePickerTests.cs
@@ -79,17 +79,21 @@ namespace MaterialDesignThemes.UITests.WPF.DatePickers
     <DatePicker Style=""{{StaticResource MaterialDesignOutlinedDatePicker}}""
       materialDesign:DatePickerAssist.OutlinedBorderInactiveThickness=""{expectedInactiveBorderThickness}""
       materialDesign:DatePickerAssist.OutlinedBorderActiveThickness=""{expectedActiveBorderThickness}""/>
+    <Button x:Name=""Button"" Content=""Some Button"" />
 </StackPanel>");
             var datePicker = await stackPanel.GetElement<DatePicker>("/DatePicker");
             var datePickerTextBox = await datePicker.GetElement<TextBox>("PART_TextBox");
+            var button = await stackPanel.GetElement<Button>("Button");
 
             // Act
+            await button.MoveCursorTo();
+            await Task.Delay(50);   // Wait for the visual change
             var inactiveBorderThickness = await datePickerTextBox.GetProperty<Thickness>(Control.BorderThicknessProperty);
             await datePickerTextBox.MoveCursorTo();
-            await Task.Delay(100);   // Wait for the visual change
+            await Task.Delay(50);   // Wait for the visual change
             var hoverBorderThickness = await datePickerTextBox.GetProperty<Thickness>(Control.BorderThicknessProperty);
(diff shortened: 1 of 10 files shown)


MaterialDesignThemes.UITests/WPF/DatePickers/DatePickerTests.cs
```diff
@@ -64,5 +64,75 @@ public async Task OnDatePicker_WithClearButton_ClearsSelectedDate()
...
+            await Task.Delay(50);   // Wait for the visual change
+            var inactiveBorderThickness = await datePickerTextBox.GetProperty<Thickness>(Control.BorderThicknessProperty);
+            await datePickerTextBox.MoveCursorTo();
+            await Task.Delay(50);   // Wait for the visual change
+            var hoverBorderThickness = await datePickerTextBox.GetProperty<Thickness>(Control.BorderThicknessProperty);
+            await datePickerTextBox.LeftClick();
+            await Task.Delay(50);   // Wait for the visual change
+            var focusedBorderThickness = await datePickerTextBox.GetProperty<Thickness>(Control.BorderThicknessProperty);
+
+            // TODO: It would be cool if a validation error could be set via XAMLTest without the need for the Binding and ValidationRules elements in the XAML above.
```
Keboo, line 108: This is a good request. I will have to look into what that API could look like. I typically have done that by building out a custom control in the test project (as this then supports things like code behind, view models, etc). But I can certainly see a desire to not have to jump through all of those hoops.

Keboo (approved):

MaterialDesignThemes.UITests/WPF/DatePickers/DatePickerTests.cs
nicolaihenriksen: > This is a good request. I will have to look into what that API could look like. I typically have done that by building out a custom control in the test project (as this then supports things like code behind, view models, etc). But I can certainly see a desire to not have to jump through all of those hoops.

Cool! As mentioned in the PR, I already have a proof-of-concept PR in the XAMLTest library with one approach to adding this type of functionality. Looking forward to seeing how you would approach it.

Issue #2596, opened 2022-03-13
Error message for MaterialDesignOutlinedTextBox should has same margin/padding as HelperText


lszczygielek:
Error message for `MaterialDesignOutlinedTextBox` should has same margin/padding as `HelperText`.

![helper-text](https://user-images.githubusercontent.com/41156511/158069693-2124370f-4597-42f7-81d3-04b23e3f946c.png)
![validation-error](https://user-images.githubusercontent.com/41156511/158069535-ceb2ec9c-b1f3-461e-8b7d-487371a49dcf.png)
```
<TextBox Style="{StaticResource MaterialDesignOutlinedTextBox}"
		 materialDesign:HintAssist.Hint="Category name*"
		 materialDesign:HintAssist.HelperText="*Required"/>
```

Side note: I think that the issue is connected with PR https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/pull/2217 that fixed the padding specification on Filled or Outlined TextBox style. Unfortunately it doesn't cover validation template.

Xaalek:
The margin should also be corrected for other types of control. The margin is currently only correct on the textbox control. When we look at the password control, we notice that the margin is not good
![image](https://user-images.githubusercontent.com/54487782/159238564-75f8e1f1-c23e-413f-9664-40ea12b802c6.png)

Xaalek:
The HelperText issue has been fixed https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/pull/2648

Now we need to solve the problem with the validation template.

- [x] HelperText template
- [ ] Validation template

Xaalek:
The task is available if someone wants to solve the issue with the validation style

nicolaihenriksen:
@Xaalek I believe the linked PR should solve the validation error margin issue.

Xaalek:
@nicolaihenriksen From what I remember, the styling was incorrect for the outlined and filled fields. This was for the TextBox, Password, ComboBox, DatePicker and RickTextBox controls.
We should check if all the controls are in accordance with the specs https://material.io/components/text-fields

nicolaihenriksen:
@Xaalek Indeed you are correct. I will push more commits to the PR fixing the remaining ones...

Issue #2812, opened 2022-08-10
Add a revealed PasswordBox style


Drise13:
**Describe the solution you'd like**
Looking for a toggle-based password box reveal style. Something similar to [MahApp's revealed password box](https://github.com/MahApps/MahApps.Metro/blob/4bc68c7d37b11827319e5eaba9dfadba9ea6e80d/src/MahApps.Metro/Styles/Controls.PasswordBox.xaml#L340). MahApps seems to require mouse-down to reveal. I believe the correct Material interaction would be to toggle states.

**Describe alternatives you've considered**
We currently have a custom control that just switches between a textbox and password box, but the toggle button is separate from the input control and seems awkwardly placed.

![image](https://user-images.githubusercontent.com/10517125/183972573-bf922a68-5d59-41bd-9dc3-a1857707c5dd.png)

**Additional context**
Possibly additional nice to haves would be able to directly control the icons used for each state.
Also would be nice if it supported validation rules. (Think WIFI's 8-63 character requirement)

nicolaihenriksen:
@Drise13 This seems like a valid request, but I am wondering whether there are some security concerns to be aware of here. The `PasswordBox` in WPF stores the entered password in a `SecureString` (whether that is secure or not is another discussion). Opting to reveal the password would require pulling the password out of the `SecureString` and into a clear-text string which naturally will be stored in memory. This can be picked up by a process dump and memory inspection. @Keboo thoughts?

To address the issue you have with the awkwardly placed reveal button, I think you could solve that quite easily. I would do something akin to this:
```
<Grid>
  <PasswordBox Padding="0,0,30,0" />
  <TextBlock Padding="0,0,30,0" />
  <Button Width="24" HorizontalAlignment="Right" />
</Grid>
```

The `Padding` on the `PasswordBox`/`TextBlock` ensure there is room for the button "inside" the `PasswordBox`. `Padding` and `Width` of course need to be adjusted to the values that match your button.

If the above approach is not feasible in your control, you could also apply the `Padding` on the `PasswordBox` like above, and then simply use a negative `Margin.Left` property on your button.

Keboo:
I think it is reasonable to implement this within the library. A few details on what I would expect for the implementation. 
[...]

PR #2820, opened 2022-08-20
Fix for 2596 - Align left margin for error message on outlined TextBox with helper text left margin

nicolaihenriksen:
Fix for #2596

The error message and helper text should have the same Margin.Left value. Updated control template trigger to use the 16px left margin like the helper text. This is applied for _Outlined_ and _Filled_ styles for the following controls:
* TextBox
* PasswordBox
* ComboBox
* DatePicker
* TimePicker

Added UI tests to ensure these stay in sync for the _Outlined_ and _Filled_ TextBoxes (which should cover all controls above).

### DIFF
--- a/MainDemo.Wpf/Domain/PasswordHelper.cs
+++ b/MainDemo.Wpf/Domain/PasswordHelper.cs
@@ -1,0 +1,69 @@
+﻿namespace MaterialDesignDemo.Domain;
+
+/// <summary>
+/// Simple helper class to enable data-binding for <see cref="PasswordBox.Password"/> which is not a <see cref="DependencyProperty"/>
+/// </summary>
+internal static class PasswordHelper
+{
+    public static readonly DependencyProperty PasswordProperty =
+        DependencyProperty.RegisterAttached("Password",
+        typeof(string), typeof(PasswordHelper),
+        new FrameworkPropertyMetadata(string.Empty, OnPasswordPropertyChanged));
+
+    public static readonly DependencyProperty AttachProperty =
+        DependencyProperty.RegisterAttached("Attach",
+        typeof(bool), typeof(PasswordHelper), new PropertyMetadata(false, Attach));
+
+    private static readonly DependencyProperty IsUpdatingProperty =
+       DependencyProperty.RegisterAttached("IsUpdating", typeof(bool),
+       typeof(PasswordHelper));
+    
+    public static void SetAttach(DependencyObject dp, bool value) => dp.SetValue(AttachProperty, value);
+
+    public static bool GetAttach(DependencyObject dp) => (bool)dp.GetValue(AttachProperty);
+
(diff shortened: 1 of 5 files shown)


MainDemo.Wpf/Domain/PasswordHelper.cs
```diff
@@ -0,0 +1,69 @@
+﻿namespace MaterialDesignDemo.Domain;
+
+/// <summary>
+/// Simple helper class to enable data-binding for <see cref="PasswordBox.Password"/> which is not a <see cref="DependencyProperty"/>
+/// </summary>
+internal static class PasswordHelper
```
Keboo, line 6: I looked at this after writing up my comment on #2812.  I think for this class we should end up making it public (and moved into the main project) and renamed to `PasswordBoxAssist` (for naming consistency) to support that use case down the road.

MainDemo.Wpf/Domain/PasswordHelper.cs
```diff
@@ -0,0 +1,69 @@
...
+{
+    public static readonly DependencyProperty PasswordProperty =
+        DependencyProperty.RegisterAttached("Password",
+        typeof(string), typeof(PasswordHelper),
+        new FrameworkPropertyMetadata(string.Empty, OnPasswordPropertyChanged));
+
+    public static readonly DependencyProperty AttachProperty =
```
Keboo, line 13: I think if we create a behavior for doing the syncing (like what is in #2812), I suspect we will not need this property as the registration will be done in a behavior, so attaching the behavior to the PasswordBox is the "trigger" for registering for the needed events.

MainDemo.Wpf/Domain/PasswordHelper.cs
nicolaihenriksen: I agree. I'll try to work on a PR for doing the "reveal PasswordBox" when I have the time for it. Once that is in place, it could definitely replace this class here.

MainDemo.Wpf/Domain/PasswordHelper.cs
nicolaihenriksen: I agree that a behavior is a better approach. However, a behavior is typically a dedicated type. I still think we should create a `PasswordBoxAssist` type which can hold attached properties for the style and such, but I don't think that class itself should be the behavior. What do you think is the best approach? Should we create another type alongside of it (e.g. `PasswordBoxBindingBehavior` or similar)? Or perhaps create it as a nested class under the `PasswordBoxAssist` (e.g. `PasswordBoxAssist.BindingBehavior`)?

Keboo (approved):

MainDemo.Wpf/Domain/PasswordHelper.cs
Keboo: I am good with that.

Issue #2722, opened 2022-05-29
DataGrid unintuitive behaviour


ThatGuyMike7:
### Bug explanation

When I click twice on a field, its entire row becomes grey and my clicks on the field are ignored (can't edit the value), until I click on another field in another row. It seems to be very unintuitive.

I don't expect two clicks on a field to result in not being able to edit that field anymore.

I use the DataGrid like this:
```xaml
<Grid>
    <TabControl TabStripPlacement="Top">
        <TabItem Header="Produkte" MaxWidth="10000">
            <StackPanel Orientation="Vertical">
                <ScrollViewer Height="300" Margin="10">
                    <DataGrid x:Name="ProductDataGrid" MinColumnWidth="50" ItemsSource="{Binding ProductData}"></DataGrid>
                </ScrollViewer>
            </StackPanel>
        </TabItem>
    </TabControl>
</Grid>
```

### Version

4.5.0

Xaalek:
I noticed the same behavior in the Demo App.

We have to clic outside the textbox do be able to edit the cell.

https://user-images.githubusercontent.com/54487782/170958020-53f7b14e-71bb-4ff9-a289-41cfb7909621.mp4

soi013:
I had the behavior in my app. I found a workaround.
Set the `Focusable` of the `ScrollViewer` above the `DataGrid` to `False`.

```xaml
<ScrollViewer 
   Focusable="False">
...
   <DataGrid>
...
```

In the Demo App, I fix this line.
https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/a4a5cd21e328ce1e473811ad330a40e5f9ac1d55/MainDemo.Wpf/MainWindow.xaml#L220

However, I am not sure what the cause is.

jbooth88:
I think setting DataGridAssist.EnableEditBoxAssist="False" resolves the issue? Unfortunately it re-establishes the double-click to edit (default) behaviour.

My uneducated guess would be that the AllowDirectEditWithoutFocus() method appears to work the first time when clicking directly on an element inside a datagrid cell, but won't work again unless the user either clicks outside of the contained element but still in the target cell, OR they click on another element in a different row.

Please confirm.

nicolaihenriksen:
[...]

PR #2821, opened 2022-08-20
Let DataGridCell handle mouse events itself when already editing

nicolaihenriksen:
Possible fix for #2722.

**UPDATE** There is another PR (#2824) which refactors this part. This PR only addresses the single problem mentioned in issue #2722 but @Xaalek pointed out that the `DataGrid` mouse behavior in general was kinda strange. Thus I ended up doing the other PR, just to see if I could make the UX of the `DataGrid` mouse/keyboard interactions a little smoother and intuitive. Have a look at the other one first, and if that is accepted, this PR should not be needed.

---

@Xaalek Could you please verify if this is the desired behavior?

The change is quite simple, we bail out of the (tunneling) event `PreviewMouseLeftButtonDown` in case the `DataGridCell` receiving the click is already editing. This way the cell itself handles the (bubbling) event `MouseLeftButtonDown` and does what it needs to. I think this is what is desired.

The effect you get when you single click the column is that cell editing starts and the I-cursor is placed at the position of the click. If you double-click, it first starts the editing and then the "second click" (handled by the cell itself) ends up highlighting/selecting the word at the position of the mouse. Seems intuitive to me, but let me know what you think.

### DIFF
--- a/MaterialDesignThemes.Wpf/DataGridAssist.cs
+++ b/MaterialDesignThemes.Wpf/DataGridAssist.cs
@@ -242,6 +242,11 @@ namespace MaterialDesignThemes.Wpf
                     {
                         return;
                     }
+                    if (dataGridCell.IsEditing)
+                    {
+                        // If the cell is already being edited, we let the cell itself handle the mouse event
+                        return;
+                    }
 
                     dataGrid.CurrentCell = new DataGridCellInfo(dataGridCell);
                     dataGrid.BeginEdit();


nicolaihenriksen:
@Keboo I also added you as a reviewer just to see if you have any feedback about potential side-effects of this.

Xaalek:
It's better than before but we still have something weird. Depending on where we click on the cell the behavior is different.
I also noticed something weird with the style MaterialDesignDataGridTextColumnPopupEditingStyle.
To open the edition we have to click on the edge of the cell. Moreover the pop up doesn't open at the right place.
https://youtu.be/pWXAf7tRK1Q

Unfortunately, there are no specs for editable datagrids. However, we could be inspired by this library to do the job https://mui.com/x/react-data-grid/ (especially for the style with the pop up)

Keboo:
No objection from me. @Xaalek it looks like you might have had feedback. Any issues with this one?

Keboo (approved):

Xaalek (approved):

nicolaihenriksen:
> No objection from me. @Xaalek it looks like you might have had feedback. Any issues with this one?

@Keboo The issues that @Xaalek mentioned he actually did a screen capture of. You can see the link in his comment [here](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2722#issuecomment-1221619244). The other PR (#2824) mentioned above is an attempt to address these issues.

PR #2824, opened 2022-08-22
Refactoring DataGridAssist input handling

nicolaihenriksen:
While working on #2821 @Xaalek indicated that the mouse handling was a bit strange. Indeed it was/is.

This PR refactors the mouse handling (and keyboard handling) of the DataGridAssist to try to sort out the strange behavior.

The key changes are that DataGridAssist:
* No longer handles the tunneling `PreviewMouseLeftButtonDown` event, but rather handles the bubbling event `MouseLeftButtonDown`. This event is already handled by the cell, so `UIElement.AddHandler()` is needed in order for the delegate to be invoked on already handled events (the last bool parameter of the method signature). This sort of flips the problem "upside down", but it seems more natural to do it this way; in the existing code we're raising bubbling events of the same type (mouse left button down) from inside a tunneling event, that just seems wrong to me.
* No longer handles the bubbling `KeyDown` event, but rather handles the tunneling `PreviewKeyDown` event. This is merely used to start editing a cell, and thus the cell itself does not need to process the key event first. I was a little worried this might kill the use of spacebar inside of a `TextBox` edit control, but it does not (probably because the `e.OriginalSource` is not a `DataGridCell` in that scenario).

I suggest you checkout the branch and play around with it. Based on the history of the `DataGridAssist` type I could see that there previously was some issues with correctly displaying/highlighting the selected cell/row depending on the `DataGrid.SelectionUnit` property, so I added options to change that from the UI of the demo tool and also ensured that things work the way I think the user would expect it to.

There is a slight "funkyness" to the behavior that we start editing on single-click. This means that once editing is started and you press Tab to jump between the cells, the next cell also enters edit mode. However, you can use ESC to exit edit mode, and then you can use Tab to simply jump between the cells without activating edit mode. It is quite obvious why it works that way seen from the code, I am just wondering about the UX perspectives of those two different experiences using the Tab key.

Give it a spin and let me know what you think.


Xaalek:
Ok I'll check it tomorrow

Xaalek (approved):

Xaalek (approved):
I just tested it and it works perfectly. I don't notice any strange behavior anymore. There is only the pop up that does not open at the right place in my case.
image

Good job anyway!

nicolaihenriksen:
> Ok I'll check it tomorrow

@Xaalek Does it work the way you would expect?

Xaalek:
@nicolaihenriksen Yeah, perfect. Just the pop up thing is weird but that's another issue.

Keboo (approved):

PR #2828, opened 2022-08-23
New feature - PasswordBox styles with "reveal password" functionality

nicolaihenriksen:
As discussed in issue #2812 it could be useful to have a `PasswordBox` with "reveal password" toggle functionality.

This PR adds 4 styles for that specific purpose (a default style, and 3 derived styles; same concept as the normal `PasswordBox` styles).

### Details
The "reveal" styles include a "raw" `TextBox` in the same place as the "password content", and then visibility of those elements are controlled by the `PasswordBoxAssist.IsPasswordRevealed` attached property. They of course also contain the "reveal button" located to the right of the password. There is also some focus stuff and whatnot that was needed in order to get the "right feel".

I actually opted **not** to go for the behavior approach as discussed, because to my knowledge I would then need to add a new nuget dependency (Microsoft.Xaml.Behaviors.Wpf) which I would rather not. I might be wrong, in which case just let me know. So I ended up doing something similar but only using attached properties. It is still quite simple to actually bind to the `PasswordBox.Password` property using this approach, and since it is simply using `PasswordBoxAssist` attached properties, it is a concept that the users will be familiar with. You can see examples of this in the samples I added in the demo tool. Reveal button icons and "reveal mode" are all publicly available attached properties on `PasswordBoxAssist` allowing the user full flexibility.

### Testing
I added a UI test to ensure the "3-way" `Binding` works as expected. The 3-way `Binding` can be illustrated like this:
`VM` <--> `PasswordBox` <--> `RevealedPasswordTextBox`. A change in any of the 3 should trigger an update in the other 2 (depending on the `Binding.UpdateSourceTrigger` value).

### Issue(s)
There is some vertical alignment of the "reveal" button that annoys my eye, but I chose to align it with the "clear text" button which is an opt-in feature for `TextBox`/`PasswordBox` and can be enabled if one so desires (the samples I added all have it enabled). For 2 of the styles, the buttons hover a bit too high compared to the baseline of the text in my opinion. It does however align with the [material.io](https://material.io/components/text-fields) specs, so I guess it should be that way.

image

There might also be some cosmetic things around the width of the IBeam and the `Foreground` color. But those also exist on the normal `TextBox`/`PasswordBox` which I think has to do with the 150% scaling I have on my monitors. I have tried setting `UseLayoutRounding` and `SnapToDevicePixels` to true, but the issue still persists. Look at how the width of the IBeam changes depending on where it is shown (zoomed to 200%):

ScreenGrab


Xaalek:
Good job! Just have to fix the hint bug and it will be perfect.

nicolaihenriksen:
@Keboo I resolved the conflicts in this branch after you merged the other PRs, however one thing is a bit weird to me. I have replaced the usage of the `MainDemo.Wpf.Domain.PasswordHelper` (temporarily introduced in another PR) with the use of `PasswordBoxAssist` approach, and delete the `PasswordHelper.cs` file in this branch. The file is not present on my branch, but the deletion does not seem to be part of the changes here?! Did I possibly do something wrong when resolving the conflicts?!

Keboo:
Looks like you did it fine. Even though the deletion does not show in the change set it is because the other PR already did it, and you also took care of it in [this commit](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/pull/2828/commits/a7f23a4818870a30d1437c8bbcf71295a8f00db6) when resolving conflicts.

I pushed a commit to your branch for the hint issue. I believe it should now be resolved. The issue was that the password box hint proxy only checked if the password box itself had focus, not if the focus was within a child, so I adjusted that (IMO it also makes more sense). I believe the behavior is correct, but let me know what you think,

nicolaihenriksen:
> only checked if the password box itself had focus, not if the focus was within a child, so I adjusted that

Makes a lot of sense. Thanks for the help!

MainDemo.Wpf/Fields.xaml
```diff
@@ -37,6 +37,63 @@
...
+                <RowDefinition Height="70" />
+                <RowDefinition Height="80" />
+            </Grid.RowDefinitions>
+
+            <TextBlock Grid.Column="0" Grid.Row="0" Style="{StaticResource MaterialDesignBody2TextBlock}" VerticalAlignment="Top" Text="MaterialDesignRevealPasswordBox" />
+            <TextBlock Grid.Column="2" Grid.Row="0" Style="{StaticResource MaterialDesignBody2TextBlock}" VerticalAlignment="Top" Text="MaterialDesignFloatingHintRevealPasswordBox" />
+            <TextBlock Grid.Column="4" Grid.Row="0" Style="{StaticResource MaterialDesignBody2TextBlock}" VerticalAlignment="Top" Text="MaterialDesignFilledRevealPasswordBox" />
+            <TextBlock Grid.Column="6" Grid.Row="0" Style="{StaticResource MaterialDesignBody2TextBlock}" VerticalAlignment="Top" Text="MaterialDesignOutlinedRevealPasswordBox" />
+
+            <PasswordBox
```
Keboo, line 65: I do think pulling the password box into its own section would be a good idea. Also we need to wrap each of these in a `<smtx:XamlDisplay>` element so that they get the nice XAML preview.

nicolaihenriksen: I moved the section down a bit and modified it slightly. It now uses ShowMeTheXAML wrappers and has an option for disabling the `PasswordBox`.

In doing the latter part, I actually found a bug. The opacity of the `PART_ClearButton` was not consistent across the styles due to double opacity being applied in certain cases. I fixed it so it is now consistent, and did the same for the "normal" `PasswordBox` styles and `TextBox` styles where the same bug was present.

The disabled states now look like this (with the clear button visible):
image

Keboo (approved):

jamesport079:
image

When using MaterialDesignFloatingHintRevealPasswordBox, reveal button position is not alinged properly. It is in line with the floating hint instead of on the textbox.

jamesport079:
It is not following the spec as such... 

material.io center justifies the icon in line with label when the field is empty (see snaphot).
image

This is the MaterialDesign equivalent...

image

nicolaihenriksen:
I agree there is a misalignment, but not only related to the reveal button then (same for the clear button). This should be created as a new issue instead of a discussion here in a closed PR.

The "misalignment" in material.io shows when you enter some text in the field (on my phone now, so screenshot not possible)

jamesport079:
Yes agreed, clear button has the same problem. I will bring up issue in a new thread. Thanks!

Issue #2826, opened 2022-08-23
DialogHost background fade out is to slow when using MaterialDesignEmbeddedDialogHost and TransitionAssist.DisableTransitions="True"


SeynalKim:
### Bug explanation

I'd like to disable the transition of the `DialogHost` and at the same time, use the style `MaterialDesignEmbeddedDialogHost`.

```
<md:DialogHost
                Grid.Row="1"
                md:TransitionAssist.DisableTransitions="True"
                Identifier="ContainerDialogHost"
                SnackbarMessageQueue="{Binding ElementName=MySnackbar, Path=MessageQueue}"
                Style="{StaticResource MaterialDesignEmbeddedDialogHost}">

</md:DialogHost>
```

Together with `md:TransitionAssist.DisableTransitions="True"` and `MaterialDesignEmbeddedDialogHost` everything Is fast when opening the dialog. However, upon closing it, I noticed that the `Background` is slowly fading out. Are there any other settings I am missing?

### Version

using version 4.5.0

Erapchu:
I'm just using that style

https://github.com/Erapchu/SinglePass.WPF/blob/develop/SinglePass.WPF/Themes/DialogHostTemplates.xaml

You can just remove PART_ContentCoverGrid and animations for it if you want the same effect but without that bug

Erapchu:
https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/pull/2829

Erapchu:
Of course I just want to leave it for future changes

nicolaihenriksen:
Closing this issue because I believe it was included in the 4.6.0 release; see PR #2829.

PR #2829, opened 2022-08-24
Fixed embedded dialog host style (#2826)

Erapchu:
Using materialDesign:TransitionAssist.DisableTransitions="true" duration of animation for content cover opacity should be set to 0 to skip animations at all on close dialog host


nicolaihenriksen (approved):
Looks good to me. Tested it and it does what is expected.

Keboo (approved):

Issue #2831, opened 2022-08-26
Cursor state on textbox selection


Erapchu:
### Bug explanation

Mouse cursor has incorrect state when click and hold on TextBox and RichTextBox.

https://user-images.githubusercontent.com/37536541/186819982-1772148c-7f04-450b-8466-839126428a2c.mp4

 

### Version

4.5.0

PR #2832, opened 2022-08-26
Fix for issue 2831

nicolaihenriksen:
Fix for #2831 

Setting the `Cursor` on the style as the old code did causes that cursor to be applied when selecting text.

This PR changes the behavior such that a default is not set on the style, but rather on the contents - `Arrow` is used in general, but `IBeam` on the text content itself (as it was before). An `IValueConverter` was added in order to respect if a `Cursor` was explicitly set from the caller, in which case that cursor should be used. I initially tried simply using a regular binding (with relative source=templated parent) with a TargetNullValue/FallbackValue, but that only partially fixes the problem; this approach seems to work better.

**NOTE:** Using that converter for the text content (`PART_ContentHost`) is a minor change of behavior being that if an explicit cursor is set from the outside, it **also** affects the text content where the `IBeam` was previously hardcoded. I think this is a fair assumption to make when a user explicitly sets the cursor on a `TextBox`. I would assume the user wants that ´Cursor´ on "all" the `TextBox` including all the "edges" around the actual text area. But I'll let you be the judge of that.

### DIFF
--- a/MaterialDesignThemes.Wpf/Converters/CursorConverter.cs
+++ b/MaterialDesignThemes.Wpf/Converters/CursorConverter.cs
@@ -1,0 +1,17 @@
+﻿using System.Globalization;
+using System.Windows.Data;
+using Cursor = System.Windows.Input.Cursor;
+
+namespace MaterialDesignThemes.Wpf.Converters;
+
+/// <summary>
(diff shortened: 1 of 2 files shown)


Keboo (approved):

Keboo:
I agree, I think that is a reasonable assumption.

Issue #2835, opened 2022-08-31
Cursor state on PasswordBox selection


Erapchu:
### Bug explanation

Please, lets implement the same fixes for PasswordBox like

https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/2831

### Version

4.5.0

nicolaihenriksen:
Fix applied to `PasswordBox` styles as well.

PR #2836, opened 2022-08-31
Fix for 2835 - Apply cursor fix for PasswordBox the same way it was done for TextBox

nicolaihenriksen:
Fix for #2835 

Same cursor fix as was done for `TextBox` (PR #2832)  is added to `PasswordBox` with this PR.


Keboo:
This looks like it was already completed as part of the other password PR

nicolaihenriksen:
> This looks like it was already completed as part of the other password PR

Correct. I cherrypicked the commit and applied similar changes to the "reveal" style in the other PR.

PR #2846, opened 2022-09-10
Fix SetPasswordCommand message

nicolaihenriksen:
Makes the password being set better reflect where it originates from. Not sure why I had "code-behind" there in the first place...


Keboo (approved):
:)

Issue #2844, opened 2022-09-09
DialogHost and Multithreading


JDeuker:
### Bug explanation

I have two UI Threads, both having a DialogHost Control.

At the moment I only get a `System.InvalidOperationException: 'The calling thread cannot access this object because a different thread owns it.'` as soon I call Show() on DialogHost.

I tried different dialogIdentifier names, but without success. Then I looked at the code of `DialogHost.cs` at the method  ` private static DialogHost GetInstance(object? dialogIdentifier)`.

I have two LoadedInstances, one for each Thread.
Everytime `GetInstance `gets called, the foreach loop will iterate through all instances and will call      `dialogInstance.Dispatcher.VerifyAccess();` which will obviously fail.
Wouldn't it be smarter to move the Dispatcher Verification into the  `if (Equals(dialogIdentifier, dialogInstance.Identifier))`?
So that only on a matched dialogIdentifier the Thread is relevant?

```
private static DialogHost GetInstance(object? dialogIdentifier)
        {
            if (LoadedInstances.Count == 0)
                throw new InvalidOperationException("No loaded DialogHost instances.");
            List<DialogHost> targets = new();
            foreach (var instance in LoadedInstances.ToList())
            {
                if (instance.TryGetTarget(out DialogHost? dialogInstance))
                {
                    dialogInstance.Dispatcher.VerifyAccess();
                    if (Equals(dialogIdentifier, dialogInstance.Identifier))
                    {
                        targets.Add(dialogInstance);
                    }
                }
                else
                {
                    LoadedInstances.Remove(instance);
                }
            }
}
```

### Version

4.4.0

nicolaihenriksen:
@JDeuker your suggestion seems reasonable at first glance, however it will not work with such a simple change. The reason being that `dialogInstance.Identifier` is a `DependencyProperty` and accessing the value of that is **also** only allowed from the owning thread.

I have created a PR #2849 which works around this by extracting the identifier using the `Dispatcher` associated with the owning thread. However, I am unsure whether you will just run into problems elsewhere. Perhaps this is something @Keboo could shed some light on? You could of course also just build the library yourself with the changes from the PR and see if that solves your issue.

PR #2849, opened 2022-09-14
Possible fix for multi-thread usage of DialogHost.GetInstance()

nicolaihenriksen:
Possible fix for #2844 

I added a unit test which initially failed with the same exception as reported in the issue. Slight modifications to the `DialogHost.GetInstance()` were added in order to retrieve the identifier using the `Dispatcher` associated with the owning thread. This makes the test run green. This change effectively replaces the `Dispatcher.VerifyAccess()` call by using the owning thread when needed.

As mentioned in the bug, the `Dispatcher.VerifyAccess()` guard was to some extent unnecessary because the following `dialogInstance.Identifier` DP-getter would result in the same exception being thrown.

I am a bit unsure whether there could be other areas where a similar approach would be needed.

### DIFF
--- a/MaterialDesignThemes.Wpf.Tests/DialogHostTests.cs
+++ b/MaterialDesignThemes.Wpf.Tests/DialogHostTests.cs
@@ -1,5 +1,6 @@
 ﻿using System.ComponentModel;
 using System.Threading;
+using System.Windows.Threading;
 using Xunit;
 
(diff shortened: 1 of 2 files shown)


nicolaihenriksen:
@Keboo I don't understand the latest failures of this PR verification. Is it the double-source generator issue coming into play again?

Keboo (approved):

Issue #2893, opened 2022-10-04
Regression in Datagrid behaviour


MarkusKgit:
### Bug explanation

Hi there,

after upgrading from 4.5.0 to 4.6.1 I noticed a change in behaviour of the Datagrid:
I have a `DataGrid` with a `DataGridTemplateColumn` containing a Button that is bound to a command of the child VM. Before 4.6.0/4.6.1 command execution was working fine on first click of the Button but now the first click is ignored (as if the row needs to be selected first). You can see it here:
Previous behaviour (which is also vanilla WPF behaviour):
![450](https://user-images.githubusercontent.com/13160892/193819771-b23b20d1-d868-48f7-bc64-cb0e8223cd53.gif)
Current behaviour:
![461](https://user-images.githubusercontent.com/13160892/193819804-7cfcd347-50c2-4d03-940e-f55bc73a7180.gif)

I created a [(somewhat) minimal repro](https://github.com/MarkusKgit/MaterialDesignDataGridButton)

### Version

4.6.1

PR #2898, opened 2022-10-04
Fix DataGridTemplateColumn bug

amescodes:
Add type check DataGrid columns for DataGridTemplateColumn in DataGridAssist. Fix bug (#2893) where any controls within a cell of a DataGridTemplateColumn would not respond to the first click.

### DIFF
--- a/MaterialDesignThemes.Wpf/DataGridAssist.cs
+++ b/MaterialDesignThemes.Wpf/DataGridAssist.cs
@@ -233,6 +233,12 @@ public static class DataGridAssist
         if (dataGridCell.Content is UIElement element)
         {
             var dataGrid = (DataGrid)sender;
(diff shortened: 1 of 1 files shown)


Keboo (approved):

csutcliff:
Commenting in case this helps anyone else. As well as the extrac click needed this was causing the following exception to be thrown for me when refreshing CollectionViews backing DataGrids with DataGridTemplateColumn(s) which had been clicked into by a user.

`System.InvalidOperationException: 'Refresh' is not allowed during an AddNew or EditItem transaction.`

Any eta on when this fix makes it into a release?

nicolaihenriksen:
@csutcliff It made it into the release (4.7.0) today 😃

Issue #2923, opened 2022-10-27
Scroll of root panel is not working when hover over other panels


yarus:
### Bug explanation

In MaterialDesign Demo application (WPF) open Tabs page - it has vertical scrollbar since there are a lot of content inside.
No move your mouse to hover over any Tab header and try to use mouse wheel to scroll.

Expected Result:
Scrolling works as normal.

Actual Result:
Scrolling is not working at all.

If I move my mouse out from Tab header and hover over other content, scrolling seems to be working fine.

I noticed the same behavior for Expander control. If I hover over Expander header, scroll wont work.

### Version

4.6.1

nicolaihenriksen:
Seems like a bug, as long as the mouse is not hovering (nested) scrollable content, scrolling should work as you expect in your description. Don't you agree @Keboo?

I will take a look at this.

PR #2924, opened 2022-10-28
Fix for 2923: Ignore vertical scroll inside TabControl header.

nicolaihenriksen:
Fix for #2923 

This PR introduces `ScrollViewerAssist.IgnoreVerticalScroll` attached property which effectively allows a user to opt-in on disabling vertical scrolling in a `ScrollViewer`. It works by registering for the already handled `MouseWheel` event, and then bubbling a copy of that event on its parent element.

A ScrollViewer will by default - as far as I know - consume (handled=true) the `MouseWheel` event, even if there is nothing for that particular `ScrollViewer` to scroll in (i.e. all content is displayed and visible in the viewport). The problem in the issue mentioned above was that the `TabControl` uses a `ScrollViewer` inside its header to enable _horizontal_ scrolling (if there is not enough room for the tabs I presume), but that `ScrollViewer` consumes the _vertical_ `MouseWheel` events as well; effectively disabling scrolling all together when the mouse hovers the `ScrollViewer`.

In scenarios like these where you want "horizontal" scrolling only, you can apply the attached property mentioned above. You can argue that you could end up with double-scrolling if the nested `ScrollViewer` (and the outer `ScrollViewer`) actually CAN scroll vertically AND you apply this attached property. However I see that as a somewhat unlikely scenario... In those cases you should of course not use the attached property.

I applied the attached property on the `ScrollViewer` in the `MaterialDesignTabControlBase` style and also added a sample to the demo application to showcase that it still works with scrollable content inside of the tab control.

The issue mentions that there is also a problem with the expander, but I believe that was just an issue in the demo application which had a not needed `ScrollViewer` added to `Expander.xaml`.


MaterialDesignThemes.Wpf/ScrollViewerAssist.cs
```diff
@@ -1,169 +1,254 @@
...
 
-        private static readonly DependencyProperty HorizontalScrollHookProperty = DependencyProperty.RegisterAttached(
-            "HorizontalScrollHook", typeof(HwndSourceHook), typeof(ScrollViewerAssist), new PropertyMetadata(null));
+    public static readonly DependencyProperty SupportHorizontalScrollProperty = DependencyProperty.RegisterAttached(
+        "SupportHorizontalScroll", typeof(bool), typeof(ScrollViewerAssist), new PropertyMetadata(false, OnSupportHorizontalScrollChanged));
 
-        public static readonly DependencyProperty SupportHorizontalScrollProperty = DependencyProperty.RegisterAttached(
-            "SupportHorizontalScroll", typeof(bool), typeof(ScrollViewerAssist), new PropertyMetadata(false, OnSupportHorizontalScrollChanged));
+    public static readonly DependencyProperty IgnoreVerticalScrollProperty = DependencyProperty.RegisterAttached(
+        "IgnoreVerticalScroll", typeof(bool), typeof(ScrollViewerAssist), new PropertyMetadata(false, OnIgnoreVerticalScrollChanged));
```
Keboo, line 78: I would move this down so that it is next to `OnIgnoreVerticalScrollChanged`

Keboo (approved):

MaterialDesignThemes.Wpf/ScrollViewerAssist.cs
nicolaihenriksen: Fair point; will move it.

Also, I am actually not satisfied with the name of it. Because in reality it does not ignore the vertical scroll, it simply re-raises the event allowing it to bubble further up the visual tree. Suggestions for a good name are welcome :-)

I have now renamed it to `RaiseVerticalScrollUpstream` but was also considering `BubbleVerticalScrollUpwards`. Thoughts?

Keboo: What about `BubbleVerticalScroll`? Given that routed events using bubbling to indicate that an event flow upward through the tree, I think that the direction is clear.

Keboo (approved):

MaterialDesignThemes.Wpf/ScrollViewerAssist.cs
nicolaihenriksen: Yeah I guess that makes sense.
My only concern is that it does not actually bubble the **original** event (because that is already handled), instead it creates a copy; this was the reasoning for going with the "Raise" prefix rather than the bubble.

Keboo: That is fair, though I think the "Raise" prefix might have people assuming they can register for the event on the ScrollViewer itself which would then miss the event since it starts at the ScrollViewer's parent rather than on the ScrollViewer itself.

I think I will merge this for now, but happy to revisit alternate names.

PR #2932, opened 2022-11-04
Refactor PasswordBoxAssist.Password to use behaviors (issue 2930)

### LINKED ISSUE #2930: DialogHost PasswordBox PasswordBoxAssist.Password binding is not working,
### Bug explanation

I have a PasswordBox,created by DialogHost.Then PasswordBoxAssist.Password binding is not working.
like this:

    `  var view = new SampleDialog
            {
                DataContext = new SampleDialogViewModel()
            };

            //show the dialog
            var result = await DialogHost.Show(view, "RootDialog", null, ClosingEventHandler, ClosedEventHandler);`
SampleDialog.xmal
`  <PasswordBox Grid.Row="1" wpf:HintAssist.HelperText="Name"
                       wpf:HintAssist.Hint="Name"
                       wpf:PasswordBoxAssist.Password="{Binding Path=Name, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged, ValidatesOnExceptions=True}"
                       Style="{StaticResource MaterialDesignFloatingHintPasswordBox}" />`

SampleDialogViewModel.cs
` 
public class SampleDialogViewModel : ViewModelBase
    {

        private string? _name;

        public string? Name
        {
            get => _name;
            set => SetProperty(ref _name, value);
        }

        public string? number;
        public string? Number
        {
            get => number;
            set => SetProperty(ref number, value);
        }
    }`

### Version

4.6.1

nicolaihenriksen:
Fixes #2930 

**UPDATE:** The modified UI tests run fine locally but seem to fail in the GitHub pipeline. I have no idea why.

As mentioned in the bug, there are certain scenarios where the current implementation falls short. This PR refactors the `PasswordBoxAssist` to use `Microsoft.Xaml.Behaviors` in order to handle the binding of the password. The code is **heavily** inspired by the [MahApps.Metro](https://github.com/MahApps/MahApps.Metro/blob/develop/src/MahApps.Metro/Behaviors/PasswordBoxBindingBehavior.cs) implementation of such a behavior but adjusted to fit the needs of MDIX. One difference being that the revealed password is actually editable in the MDIX implementation whereas in MahApps it is only temporarily displayed (while mouse is down), and as such is read-only.

**Important:** This introduces a dependency, and as such, the `MaterialDesignThemes.nuspec` may need to be updated accordingly. I am no nuget-master, so I am unsure what the correct modifications would be.

**Important:** Since a lot of code was copied (and slightly adjusted) from the MapApps repository, there may be some copyright/disclaimer stuff that needs to be added somewhere?

**Possible issue:** `Microsoft.Xaml.Behaviors` has a different set of compatible target frameworks than what MDIX has. This was giving me some issues, but I managed to get it running locally. Not sure if there is an issue or not.

**Improvements**:
- Binding of the password now works in all cases (fingers crossed)
- Focus and text selection of the "reveal" `TextBox` in the reveal styles has been greatly improved, see GIFs below. Since behaviors were introduced, I was able to refactor this and have better control of what is happening.

Focus and text selection before:
OriginalFocusBehavior
Note that the `I-beam` is not visible immediately when revealing the password (although typing still appends to the text). Also note that text selection is "forgotten" when toggling between revealed/hidden.

Focus and text selection now:
NewFocusBehavior


MaterialDesignThemes.Wpf/StylizedBehaviors.cs
```diff
@@ -0,0 +1,122 @@
+﻿using Microsoft.Xaml.Behaviors;
```
nicolaihenriksen, line 1: This whole file was more or less copied from MahApps

MaterialDesignThemes.Wpf/StylizedBehaviorCollection.cs
```diff
@@ -0,0 +1,8 @@
+﻿using Microsoft.Xaml.Behaviors;
```
nicolaihenriksen, line 1: This whole file was more or less copied from MahApps

MaterialDesignThemes.Wpf/PasswordBoxAssist.cs
```diff
@@ -24,45 +26,67 @@ public static class PasswordBoxAssist
...
+    private static readonly DependencyProperty RevealedPasswordTextBoxProperty = DependencyProperty.RegisterAttached(
+        "RevealedPasswordTextBox", typeof(TextBox), typeof(PasswordBoxAssist), new UIPropertyMetadata(default(TextBox)));
+    private static void SetRevealedPasswordTextBox(DependencyObject obj, TextBox? value) => obj.SetValue(RevealedPasswordTextBoxProperty, value);
+    private static TextBox? GetRevealedPasswordTextBox(DependencyObject obj) => (TextBox?)obj.GetValue(RevealedPasswordTextBoxProperty);
 
-    private static readonly DependencyProperty SettingPasswordProperty = DependencyProperty.RegisterAttached(
-        "SettingPassword", typeof(bool), typeof(PasswordBoxAssist), new PropertyMetadata(false));
 
-    private static void HandlePasswordChanged(DependencyObject d, DependencyPropertyChangedEventArgs e)
+    /// <summary>
```
nicolaihenriksen, line 45: The code from this line and to the end is more or less copy/pasted from MahApps

MaterialDesignThemes.UITests/WPF/PasswordBoxes/PasswordBoxTests.cs
```diff
@@ -83,42 +84,36 @@ public async Task OnPasswordBox_WithClearButton_ClearsPassword()
         }
 
         [Fact]
-        [Description("PR 2828")]
+        [Description("PR 2828 and Issue 2930")]
```
nicolaihenriksen, line 87: The original test did not catch the issues, because it updated the VM as the first thing. Doing that kick-started the bindings and then changes in the UI works afterwards.

So I have changed the test a bit, and I now have 2 tests: one for the non-reveal style and another for the reveal style.

MaterialDesignThemes.UITests/Samples/PasswordBox/BoundPasswordBox.xaml
```diff
@@ -0,0 +1,19 @@
+﻿<UserControl x:Class="MaterialDesignThemes.UITests.Samples.PasswordBox.BoundPasswordBox"
```
nicolaihenriksen, line 1: This control and the window and view model residing in the folder are simple here to be able to reproduce the non-working scenarios in UI tests.

MaterialDesignThemes.Wpf/MaterialDesignThemes.Wpf.csproj
```diff
@@ -24,6 +24,9 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Xaml.Behaviors.Wpf" />
```
Keboo, line 28: This change will require an update to `MaterialDesignThemes.nuspec` to add this NuGet package to the list of dependencies for each of the target frameworks. 

Historically we have resisted taking any outside dependencies in the core package, however, I think this package should be an exception as it really is core to WPF.

MaterialDesignThemes.Wpf/PasswordBoxAssist.cs
```diff
@@ -1,8 +1,10 @@
-﻿using System.Windows.Data;
+using System.Windows.Data;
+using System.Windows.Documents;
+using Microsoft.Xaml.Behaviors;
 
 namespace MaterialDesignThemes.Wpf;
 
-public static class PasswordBoxAssist
+public class PasswordBoxAssist : Behavior<PasswordBox>
```
Keboo, line 7: I think it might be better to move the behavior to its own class so there is no confusion between what is a behavior and what is an "Assist" (attached property) class. Perhaps named something like "PasswordBoxBehavior"?

MaterialDesignThemes.Wpf/StylizedBehaviors.cs
```diff
@@ -0,0 +1,122 @@
...
+
+namespace MaterialDesignThemes.Wpf;
+
+internal class StylizedBehaviors
```
Keboo, line 5: I am wondering if we should consider making this `public` and renaming it to `BehaviorsAssist` to match our other classes. If we are going to take a dependency on the behaviors NuGet it might be a useful feature to expose. Thoughts?

MaterialDesignThemes.Wpf/StylizedBehaviorCollection.cs
```diff
@@ -0,0 +1,8 @@
...
+
+namespace MaterialDesignThemes.Wpf;
+
+public class StylizedBehaviorCollection : FreezableCollection<Behavior>
```
Keboo, line 5: Thoughts on renaming this to simply "BehaviorCollection"?

MaterialDesignThemes.Wpf/Behaviors/PasswordBoxRevealTextBoxBehavior.cs
```diff
@@ -0,0 +1,92 @@
...
+        var typeTextRange = selection?.GetType().GetInterfaces().FirstOrDefault(i => i.Name == "ITextRange");
+        object? start = typeTextRange?.GetProperty("Start")?.GetGetMethod()?.Invoke(selection, null);
+        object? end = typeTextRange?.GetProperty("End")?.GetGetMethod()?.Invoke(selection, null);
+        int? startValue = start?.GetType().GetProperty("Offset", BindingFlags.NonPublic | BindingFlags.Instance)?.GetValue(start, null) as int?;
+        int? endValue = end?.GetType().GetProperty("Offset", BindingFlags.NonPublic | BindingFlags.Instance)?.GetValue(end, null) as int?;
+        int selectionStart = startValue.GetValueOrDefault(0);
+        int selectionLength = 0;
+        if (endValue.HasValue)
+        {
+            selectionLength = endValue.Value - startValue.GetValueOrDefault(0);
```
Keboo, line 74:
```suggestion
            selectionLength = endValue.Value - selectionStart;
```

MaterialDesignThemes.Wpf/Behaviors/PasswordBoxRevealTextBoxBehavior.cs
```diff
@@ -0,0 +1,92 @@
...
+
+    private PasswordBoxSelection GetPasswordBoxSelection()
+    {
+        var selection = GetSelection(AssociatedObject);
+        var typeTextRange = selection?.GetType().GetInterfaces().FirstOrDefault(i => i.Name == "ITextRange");
+        object? start = typeTextRange?.GetProperty("Start")?.GetGetMethod()?.Invoke(selection, null);
+        object? end = typeTextRange?.GetProperty("End")?.GetGetMethod()?.Invoke(selection, null);
+        int? startValue = start?.GetType().GetProperty("Offset", BindingFlags.NonPublic | BindingFlags.Instance)?.GetValue(start, null) as int?;
+        int? endValue = end?.GetType().GetProperty("Offset", BindingFlags.NonPublic | BindingFlags.Instance)?.GetValue(end, null) as int?;
+        int selectionStart = startValue.GetValueOrDefault(0);
```
Keboo, line 70: I personally like the null-coalescing operator for these, but not a big deal either way.
```suggestion
        int selectionStart = startValue ?? 0;
```

MaterialDesignThemes.Wpf/Behaviors/PasswordBoxRevealTextBoxBehavior.cs
```diff
@@ -0,0 +1,92 @@
...
+        set => SetValue(PasswordBoxProperty, value);
+    }
+
+    protected override void OnAttached()
+    {
+        base.OnAttached();
+        AssociatedObject.IsVisibleChanged += AssociatedObjectOnIsVisibleChanged;
+        if (PasswordBox != null)
+        {
+            var info = typeof(PasswordBox).GetProperty("Selection", BindingFlags.NonPublic | BindingFlags.Instance);
```
Keboo, line 29: I would pull this out into a private static property to avoid duplicate reflection calls.
Something like this: 
```C#
private static PropertyInfo SelectionProperty { get; } = typeof(PasswordBox).GetProperty("Selection", BindingFlags.NonPublic | BindingFlags.Instance) ?? throw new InvalidOprationException("Did not find 'Selection' property on PasswordBox");```

File: MaterialDesignThemes.Wpf/Behaviors/PasswordBoxRevealTextBoxBehavior.cs [hunk 13]
```diff
@@ -0,0 +1,92 @@
...
+        {
+            SetPasswordBoxSelection(AssociatedObject.SelectionStart, AssociatedObject.SelectionLength);
+            Keyboard.Focus(PasswordBox);
+        }
+    }
+
+    private PasswordBoxSelection GetPasswordBoxSelection()
+    {
+        var selection = GetSelection(AssociatedObject);
+        var typeTextRange = selection?.GetType().GetInterfaces().FirstOrDefault(i => i.Name == "ITextRange");
```
Keboo, line 65: I would also pull the reflection look-ups here out into statics to avoid duplicate reflection calls across multiple instances (may be helpful to add a static constructor to allow for some more logic). 

For example, because this one is targeting a specific interface I would extract the interface by name rather than by going off of the `selection.GetType()`.
Something like this:
```C#
static PasswordBoxRevealTextBoxBehavior()
{
    Type iTextRange = typeof(PasswordBox).Assembly.GetType("System.Windows.Documents.ITextRange")
        ?? throw new InvalidOperationException("Failed to find ITextRange");
    MethodInfo getStart = iTextRange.GetProperty("Start")?.GetGetMethod()
        ?? throw new InvalidOperationException($"Failed to find Start property on {iTextRange.FullName}");
        ...
}
```

MaterialDesignThemes.Wpf/Behaviors/PasswordBoxRevealTextBoxBehavior.cs
```diff
@@ -0,0 +1,92 @@
...
+        int selectionStart = startValue.GetValueOrDefault(0);
+        int selectionLength = 0;
+        if (endValue.HasValue)
+        {
+            selectionLength = endValue.Value - startValue.GetValueOrDefault(0);
+        }
+        return new PasswordBoxSelection(selectionStart, selectionLength);
+    }
+
+    private void SetPasswordBoxSelection(int selectionStart, int selectionLength) => typeof(PasswordBox).GetMethod("Select", BindingFlags.Instance | BindingFlags.NonPublic)?.Invoke(PasswordBox, new object[] { selectionStart, selectionLength });
```
Keboo, line 79: Similar thing here. I would cache the MethodInfo for the SelectMethod in a static member.

Keboo (changes requested):

MaterialDesignThemes.Wpf/MaterialDesignThemes.Wpf.csproj
nicolaihenriksen: This is also what I mention in the PR description. I am not entirely sure of the syntax in that nuspec file, so help would be greatly appreciated.

MaterialDesignThemes.Wpf/PasswordBoxAssist.cs
nicolaihenriksen: I agree. It wasn't easy to split out with the first behavior code I wrote, but with what I ended up with it should be easy to extract to a different class.

MaterialDesignThemes.Wpf/StylizedBehaviors.cs
nicolaihenriksen: Could be useful. There is also [this open issue](https://github.com/dotnet/wpf/issues/148) on dotnet/wpf which is requesting to have native support for it.

MaterialDesignThemes.Wpf/StylizedBehaviorCollection.cs
nicolaihenriksen: I am good with that. As mentioned in the PR description, much of this is "stolen" from MahApps, so perhaps credit should be given where credit is due. Renaming may blur that somewhat, but I will let you decide on that.

MaterialDesignThemes.Wpf/Behaviors/PasswordBoxRevealTextBoxBehavior.cs
nicolaihenriksen: Sounds reasonable. Will change.

MaterialDesignThemes.Wpf/Behaviors/PasswordBoxRevealTextBoxBehavior.cs
nicolaihenriksen: Sounds reasonable. Will change.

MaterialDesignThemes.Wpf/Behaviors/PasswordBoxRevealTextBoxBehavior.cs
nicolaihenriksen: Agreed.

MaterialDesignThemes.Wpf/MaterialDesignThemes.Wpf.csproj
nicolaihenriksen: @Keboo I have now pushed some updates to the nuspec file. I opted to downgrade to an earlier version of `Microsoft.Xaml.Behaviors.Wpf` in an attempt to reach a broader audience; not sure if that is needed.

Also, this is probably something that should be excluded from `dependabot` automatic updates since I assume we would like this dependency to stay fixed?

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.PasswordBox.xaml
```diff
@@ -406,6 +408,13 @@
     <Setter Property="wpf:HintAssist.Foreground" Value="{DynamicResource PrimaryHueMidBrush}" />
     <Setter Property="wpf:TextFieldAssist.TextBoxViewMargin" Value="{x:Static wpf:Constants.DefaultTextBoxViewMargin}" />
     <Setter Property="wpf:TextFieldAssist.UnderlineBrush" Value="{DynamicResource PrimaryHueMidBrush}" />
+    <Setter Property="wpf:StylizedBehaviors.Behaviors">
```
nicolaihenriksen, line 411: @Keboo This addition here has me a little concerned. Basically we now always apply the behavior on a `PasswordBox` in order to be able to support bindings via the `PasswordBoxAssist.Password` attached property. However, in doing so the consequence is that if you're using MDIX and just want a `PasswordBox` **without** using a binding, then the clear text password is still exposed in the value of that attached property (because the behavior wires it up). So there is really no way to opt-out of it.

Another approach would be not to add it to the "non-reveal" `Style` (NB. it is required in the reveal `Style` for the reveal `TextBox` to sync), and then let the user opt-in for it. Either by setting yet another attached property, or by directly adding the behavior; the latter would then require the behavior to be public.

nicolaihenriksen, line 411: @Keboo This addition here has me a little concerned. Basically we now always apply the behavior on a `PasswordBox` in order to be able to support bindings via the `PasswordBoxAssist.Password` attached property. However, in doing so the consequence is that if you're using MDIX and just want a `PasswordBox` **without** using a binding, then the clear text password is still exposed in the value of that attached property (because the behavior wires it up). So there is really no way to opt-out of it.

Another approach would be not to add it to the "non-reveal" `Style` (NB. it is required in the reveal `Style` for the reveal `TextBox` to sync), and then let the user opt-in for it. Either by setting yet another attached property, or by directly adding the behavior; the latter would then require the behavior to be public.

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.PasswordBox.xaml
```diff
@@ -406,6 +408,13 @@
...
+      <Setter.Value>
+        <wpf:StylizedBehaviorCollection>
+          <internalbehaviors:PasswordBoxBehavior />
+        </wpf:StylizedBehaviorCollection>
+      </Setter.Value>
+    </Setter>
```
nicolaihenriksen, line 417: @Keboo This addition here has me a little concerned. Basically we now always apply the behavior on a `PasswordBox` in order to be able to support bindings via the `PasswordBoxAssist.Password` attached property. However, in doing so the consequence is that if you're using MDIX and just want a `PasswordBox` **without** using a binding, then the clear text password is still exposed in the value of that attached property (because the behavior wires it up). So there is really no way to opt-out of it.

Another approach would be not to add it to the "non-reveal" `Style` (NB. it is required in the reveal `Style` for the reveal `TextBox` to sync), and then let the user opt-in for it. Either by setting yet another attached property, or by directly adding the behavior; the latter would then require the behavior to be public.

MaterialDesignThemes.Wpf/Behaviors/PasswordBoxBehavior.cs
```diff
@@ -0,0 +1,24 @@
+﻿using Microsoft.Xaml.Behaviors;
+
+namespace MaterialDesignThemes.Wpf.Behaviors
```
Keboo, line 3: I would use a file-scoped namespace here
```suggestion
namespace MaterialDesignThemes.Wpf.Behaviors;
```

MaterialDesignThemes.nuspec
```diff
@@ -17,12 +17,15 @@
     <dependencies>
       <group targetFramework="net462">
         <dependency id="MaterialDesignColors" version="[1.2.1, 2.0)" />
+        <dependency id="Microsoft.Xaml.Behaviors.Wpf" version="1.1.31" />
```
Keboo, line 20: Sorry I missed your earlier comment about wanting help with the syntax here. I would set it up to allow for a range of value. Assuming the library is following semantic versioning (which it appears they are) we should be good to tell consumers anything from this version up to (but excluding) a 2.0 release. Like this:
```suggestion
        <dependency id="Microsoft.Xaml.Behaviors.Wpf" version="[1.1.31, 2.0)" />
```

MaterialDesignThemes.Wpf/StylizedBehaviorCollection.cs
Keboo: Yea, let's go ahead and rename the class. You raise a good point, and I think we should leave a comment at the top indicating that the source of this code was based on their work and linking to their source file.

MaterialDesignThemes.Wpf/MaterialDesignThemes.Wpf.csproj
Keboo: I think for now, I will still let dependabot update it. Given that the library does not update much, if they ever release a breaking change (likely with a 2.0 release) then that will be the point where we need to make code changes to support it, and the nuspec file will be updated then.

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.PasswordBox.xaml
Keboo: This is a good point. I think I like the idea of making the user opt-in to using it on the non-reveal style. I have a slight preference to going the attached property route only because I have seen users struggle with nested elements in XAML.

MaterialDesignThemes.Wpf/StylizedBehaviors.cs
Keboo: O nice, I had not seen that open issue. I think for now let's go ahead and make it public. I know I have built this same type of class several times on projects and it would be a nice feature to have included.

nicolaihenriksen: Made it public and added a small code comment hinting at its origin. Please feel free to add additional stuff in there if you see fit.

MaterialDesignThemes.Wpf/StylizedBehaviorCollection.cs
nicolaihenriksen: Renamed. Did not actually include a link-back to MahApps in this file because it is so simple, and with the rename there is not much left from the MapApps.Metro repo. Instead I added the link-back in the `BehaviorsAssist` class.

MaterialDesignThemes.Wpf/Behaviors/PasswordBoxBehavior.cs
nicolaihenriksen: Changed to file-scoped namespace.

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.PasswordBox.xaml
nicolaihenriksen: I went with a slightly different approach which I think/hope works better. Now the default is to **not** wire up the event handler for PasswordChanged **unless** one of two things:

- The user of the control has set the `PasswordBoxAssist.Password` attached property using a binding, or
- The internal suppression of the above guard is applied. This is used for the "reveal" style where the listener must be wired up for the reveal `TextBox` to sync up correctly.

This allows us to keep the original syntax and not require additional properties from an end users perspective.

MaterialDesignThemes.nuspec
nicolaihenriksen: I was not really clear about what I meant here...

I think your suggestion to limit between a minimum version and a max version is reasonable. I think my original value meant that it should be "equal to or greater than" (ie. no upper limit). If you want a specific version, I think you put a single version in the hard brackets if I remember correctly.

Anyways, my concern was/is this:

We probably don't have a hard requirement on the latest version, we could compile against an older version in order to reach a broader audience (if they are locked into an older version of the same dependency for some reason). As such, I thought we could just put the smallest version in there which we know works.
**However**, my concern is regarding what the nuspec file says vs. what we actually compile against. If those two don't match (which dependabot would make sure was the case if we opted for an early version in the nuspec), I am under the impression that consumers of the library potentially would need a "binding redirect" even if they only have MDIX as a nuget package. I don't know this for sure, but I base the assumption on the fact that installing the nuget package, I would assume the nuget package manager simply pulls in the dependency in the (minimum) version that is spec'ed in the nuspec. This means the user ends up with an older version on disk, but MDIX compiled against (and thus expecting) a newer version; I believe this needs to be mitigated with a binding redirect.

You probably know more than I do about this, so please enlighten me :-)

nicolaihenriksen:
@Keboo I think the code in the PR is in a pretty good state now; pending any further review comments from you of course. However, the UI tests still fail on the PR, but the whole suite runs fine locally; I have no clue what is wrong...

MaterialDesignThemes.Wpf/MaterialDesignThemes.Wpf.csproj
nicolaihenriksen: @Keboo Do you know why the versions (lower- and upper-bound) in the nuspec do not match up with the ones shown on nuget.org (see screenshot in this comment thread)?

Keboo: Yea, in an effort to stop fighting with the NuGet versions, the [script that builds the NuGets](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/Scripts/BuildNugets.ps1) overrides the version in the NuSpec file

MaterialDesignThemes.nuspec
Keboo: This is a great question. After pulling the docs for it I realized I was wrong to request an upper bound on the version as the [recommendation from MS says to avoid that](https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/dependencies#nuget-dependency-version-ranges). I think we can also likely set our NuGet dependency to use a lower bound of `[1.1.3](https://www.nuget.org/packages/Microsoft.Xaml.Behaviors.Wpf/1.1.3)` and no upper bound (this was a mistake on my part with my earlier suggestion).

However on to your other question about how referenced assemblies work. First it is worth mentioning that the way [.NET Framework runtime](https://learn.microsoft.com/dotnet/framework/deployment/how-the-runtime-locates-assemblies) and the [.NET Core Runtime](https://learn.microsoft.com/dotnet/standard/assembly/resolve-loads) resolve dependencies is a little different (largely around the use of the GAC; Global Assembly Cache). It is a common misconception that the version of the assembly that you compile against will be the same version of the assembly that your run against. When loading NuGet dependencies it uses the lower common dependency version. For projects that wish to have a newer version of the library that can explicitly reference it with the version that they would like, as long as this version doesn't cause an error with other packages (or the error is suppressed) the project will then load that NuGet dependency. This then means that NuGet's assemblies (anything under the `lib` directory in the NuGet that matches the target framework) will be copies to the projects output directory (ie. `bin\Debug` or similar). When the project is launched, the runtime then takes over and attempts to resolve the assembly (details in those two links above). However, for most NuGet packages, this ends up simply resolving to the assembly that is alongside the running application (the same one that was copied from the NuGet package). As long as there are no errors resolving the public API members being accessed, it is able to use whatever version of the library it finds.

Keboo:
I will take a look at the failing tests to see what is going wrong

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.PasswordBox.xaml
Keboo: I think this is a very clean solution. Nice work!

nicolaihenriksen:
> I will take a look at the failing tests to see what is going wrong

Seems I forgot to "save all" before committing the rename changes. Thanks for pushing the update. My UI test(s) still fail though...

nicolaihenriksen:
@Keboo Looking at the produced screenshots of the failing test (Artifact "Screenshots-622"), it seems like the test window is minimized?!

Keboo:
Yea.... it appears to either be offscreen or minimized. Will have to look more into it.

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.PasswordBox.xaml
nicolaihenriksen: Thanks

MaterialDesignThemes.nuspec
nicolaihenriksen: @Keboo Thanks for the explanation. I am seemingly not very good at communicating my concern here :-) Reading the docs, I think they just confirm what I am trying to say (but failing to communicate). I am aware that you don't necessarily run against the same version of the dependent assembly as you compiled against; but it will attempt to bind with the exact version you compiled against unless "told otherwise" (eg. `<bindingRedirect>` in `App.Config` or another configuration file).

Say we have a project, P, which has 2 dependencies: A and B. Both of these assemblies have a dependency on C, but with different minimum versions (A requires min C.v1, and B requires min C.v2). Since A and B are compiled against different versions of C, project P needs to make a decision on which version of C to use. Typically, this ends up being handled by a `<bindingRedirect>`. In this example, the redirect would probably say that "for any requested version of C between v1 and v2, use v2". This would mean that dependency A, when resolving its dependency C will get C.v2 although it was compiled against and therefore is requesting C.v1. This is at least my understanding, and also something I often times need to deal with in my job.

Back to my concern; which is a slight variation of the above.
Say we have project P, which references MDIX. The MDIX nuget package indicates that the min required version of `Microsoft.Xaml.Behaviors.Wpf` is v1.1.3, however MDIX is compiled against v1.1.37 (latest). My concern is here that the nuget package manager will put v1.1.3 in the lib folder and copy that to the bin folder when compiling. To my understanding, the application P then needs a `<bindingRedirect>` in a configuration file in order to resolve `Microsoft.Xaml.Behaviors.Wpf` to v.1.1.3 when requesting v1.1.37 (which it was compiled against); this seems odd when simply pulling in a single nuget package.
This is why I believe we should also compile against the min version we state in the nuspec, and therefore avoid dependabot updates. Aagain, I am not 100% sure which version the nuget package manager pulls down and puts in the lib/bin folders, my assumption is that it is the one stated in the nuspec, but I could be wrong.

nicolaihenriksen:
@Keboo I managed to get the UI test to pass in the pipeline by adding this little loaded event handler to the window:

```csharp
private void BoundPasswordBoxWindow_OnLoaded(object sender, RoutedEventArgs e)
{
    Activate();
    Topmost = true;
    Topmost = false;
    Focus();
}
```

I really have no idea why that is necessary though.

Keboo (approved):

Keboo:
The temp windows that are created by XamlTest are being set to [Topmost="True"](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MaterialDesignThemes.UITests/XamlTestMixins.cs#L117) so i am not surprised that change was required.

nicolaihenriksen:
> The temp windows that are created by XamlTest are being set to [Topmost="True"](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MaterialDesignThemes.UITests/XamlTestMixins.cs#L117) so i am not surprised that change was required.

Could it be because this test actually uses "its own window"? There were other tests also doing that, so I didn't think twice about adding it. It could possibly be re-written to not require its own window, I just struggled getting a good handle to the `ViewModel` in order to test the binding updates when changing the value in the VM.

Keboo:
Yea the custom window is exactly right. I should probably look at putting something in place so this automatically gets set on the window by default.

Issue #2935, opened 2022-11-06
CharacterCounterVisibility on PasswordBox


Maxhy:
The addition of materialDesign:TextFieldAssist.CharacterCounterVisibility on TextBox was perfect, thanks a lot for that.
The original thread (#1607) talking about the implementation was thinking that such feature wasn't necessary on a PasswordBox, whereas it may even be more necessary as we cannot directly bind the password value from the control.

**Describe the solution you'd like**
Support materialDesign:TextFieldAssist.CharacterCounterVisibility on PasswordBox control to display a character count information at the bottom of the password input, just like for TextBox control.

**Describe alternatives you've considered**
Do it myself out of MaterialDesignInXaml, but it will not benefit from automation already existing with materialDesign:TextFieldAssist.CharacterCounterVisibility.

Thanks!

nicolaihenriksen:
@Maxhy You **can** actually bind the password from the `PasswordBox` using `PasswordBoxAssist.Password="{Binding ...}"`. However, I still think this is a reasonable request, so I will look into it.

nicolaihenriksen:
@Maxhy The "character counter" feature is available for `PasswordBox` in the latest nightly release. Give it a try if you have time 😃

Maxhy:
This looks perfect. Much better with official support than with my ugly workaround.
Thanks a lot.

Maxhy:
Out of subject but I've added your project to our internal list of open source libraries dependencies to support. I handle a company doing open source software for physical security and this project is used on one specific product. Because open source libraries usually have less visibilities than complete products, whereas the business model is similar even if not necessary developed for profits, it also has less revenues. We will try to redistribute a pourcentage of our revenues to such libraries, once a year. Just wanted to share, keep the good work!

PR #2937, opened 2022-11-07
Enable use of TextFieldAssist.CharacterCounterVisibility on PasswordBox

nicolaihenriksen:
Fix for #2935 

This copies over the "character counter" feature from `TextBox` the the `PasswordBox` styles. In order to monitor the change in password length, a new internal attached property was introduced (`TextFieldAssist.PasswordBoxCharacterCount`). The value is updated whenever the password changes; the event handler listening for the changes is wired up when the `TextFieldAssist.CharacterCounterVisibility` changes.

**Note:** I changed the default value of `TextFieldAssist.CharacterCounterVisibility` to `Hidden` in order to detect the change to `Visible`. You can consider this somewhat of a behavioral breaking change, but the base `TextBox` style explicitly sets it to `Visible` anyway (I find that a bit odd, I would assume this was opt-in functionality!).

### DIFF
--- a/MainDemo.Wpf/Fields.xaml
+++ b/MainDemo.Wpf/Fields.xaml
@@ -340,7 +340,7 @@
                         VerticalAlignment="Bottom"
                         UniqueKey="passwordFilled1">
         <StackPanel>
-          <CheckBox x:Name="MaterialDesignFilledPasswordBoxTextCountComboBox" Content="View Pwd Input Count" />
+          <CheckBox x:Name="MaterialDesignFilledPasswordBoxTextCountComboBox" Content="View Password Input Count" />
 
           <PasswordBox materialDesign:HintAssist.HelperText="Helper text"
                        materialDesign:HintAssist.Hint="Password"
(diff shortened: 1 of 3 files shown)


MainDemo.Wpf/Fields.xaml
```diff
@@ -339,9 +339,14 @@
                         Grid.Column="3"
                         VerticalAlignment="Bottom"
                         UniqueKey="passwordFilled1">
-        <PasswordBox materialDesign:HintAssist.HelperText="Helper text"
-                     materialDesign:HintAssist.Hint="Password"
-                     Style="{StaticResource MaterialDesignFilledPasswordBox}" />
+        <StackPanel>
+          <CheckBox x:Name="MaterialDesignFilledPasswordBoxTextCountComboBox" Content="View Pwd Input Count" />
```
Keboo, line 343: I would spell out password
```suggestion
          <CheckBox x:Name="MaterialDesignFilledPasswordBoxTextCountComboBox" Content="View Password Input Count" />
```

Keboo (approved):

MainDemo.Wpf/Fields.xaml
nicolaihenriksen: Did that initially, but then it clips in the demo tool with the current width of the ShowMeTheXAML elements (fixed width). So I took the shortcut and renamed 😀

PR #2979, opened 2022-11-27
Add global SystemParameters.FocusVisualStyleKey focus style

### LINKED ISSUE #2976: Use global SystemParameters.FocusVisualStyleKey insead of per control FocusVisual style
**Is your feature request related to a problem? Please describe.**
A [bunch](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/search?q=FocusVisual) of control styles define a `FocusVisual` style that is used as a setter value for `FocusVisualStyle` property and there is no way to customize them other than creating custom styles for each control and setting `FocusVisualStyle` to your own style.

Most of them are identical:
https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/2a38b2c70a5e851bc5b926e0641c1848e719d86b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Button.xaml#L17-L30

Only `OptionMarkFocusVisual` seems to have different Margin:
https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/2a38b2c70a5e851bc5b926e0641c1848e719d86b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.RadioButton.xaml#L13-L25

I think the `FocusVisualStyle` should be globally defined with `SystemParameters.FocusVisualStyleKey` in App.xaml:
```xaml
<Style x:Key="{x:Static SystemParameters.FocusVisualStyleKey}" TargetType="Control">
    <Setter Property="Control.Template"> 
        ...
    </Setter>
</Style>
```

If Margin is all that needs to be changed for some controls then maybe a `FocusVisualAssist.Margin` could be added to allow those controls to have their own `FocusVisualStyle`:
```xaml
<Style x:Key="OptionMarkFocusVisual" BasedOn="{x:Static SystemParameters.FocusVisualStyleKey}">
    <Setter Property="materialDesign:FocusVisualAssist.Margin" Value="0,0,0,0"/>
</Style>
```

Yoooi0:
Fixes #2976

It turned out to not be as easy as just removing mdix `FocusVisual` styles.

There is a lot of controls in wpf that have their own custom `FocusVisualStyle` (just different margin), and those styles are inherited when mdix creates their own style:
https://github.com/dotnet/wpf/search?q=FocusVisualStyle+Value

Seems like wpf team added `SystemParameters.FocusVisualStyleKey` as a global focus style but many controls override it anyway, so it is a little bit pointless.

I had to specify `FocusVisualStyle` setter for each style that inherits from:

- ToggleButton
- CheckBox
- ButtonBase
- Button
- RadioButton
- RepeatButton
- ComboBoxItem
- ListBoxItem
- TabItem

The other solution would be to use `OverridesDefaultStyle` but it would require checking that each style does not break because many properties would not be inherited from base wpf themes.

Breaking change examples:
before_radio
After:
after_radio
Before:
before_checkbox
After:
after_checkbox

And there would be a bunch more since some of the base wpf focus visual styles also have a different margin.
I checked each control in the demo and didnt find anything obviously broken.

### DIFF
--- a/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Defaults.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Defaults.xaml
@@ -14,6 +14,7 @@
     <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.DatePicker.xaml" />
     <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.DialogHost.xaml" />
     <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Expander.xaml" />
+    <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.FocusVisual.xaml" />
     <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Font.xaml" />
     <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.GridSplitter.xaml" />
     <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.GroupBox.xaml" />
--- a/MaterialDesignThemes.Wpf/Themes/MaterialDesign2.Defaults.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/MaterialDesign2.Defaults.xaml
@@ -15,6 +15,7 @@
     <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.DatePicker.xaml" />
     <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.DialogHost.xaml" />
     <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Expander.xaml" />
+    <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.FocusVisual.xaml" />
     <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Font.xaml" />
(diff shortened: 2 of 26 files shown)


Yoooi0:
For different margins we could add `FocusVisualAssist.Margin` and add custom focus style for each control so they look nicer.

I think something like this should work:
```xaml
  <Style x:Key="{x:Static SystemParameters.FocusVisualStyleKey}">
    <Setter Property="Control.Template">
      <Setter.Value>
        <ControlTemplate>
          <Rectangle Margin="{TemplateBinding materialDesign:FocusVisualAssist.Margin}"
                     SnapsToDevicePixels="true"
                     Stroke="{DynamicResource {x:Static SystemColors.ControlTextBrushKey}}"
                     StrokeDashArray="1 2"
                     StrokeThickness="1" />
        </ControlTemplate>
      </Setter.Value>
    </Setter>
  </Style>
```
```xaml
<Style x:Key="OptionMarkFocusVisual" BasedOn="{x:Static SystemParameters.FocusVisualStyleKey}">
    <Setter Property="materialDesign:FocusVisualAssist.Margin" Value="0"/>
</Style>
```

Custom styles would just specify the margin, template would be global.

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Defaults.xaml
```diff
@@ -14,6 +14,7 @@
     <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.DatePicker.xaml" />
     <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.DialogHost.xaml" />
     <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Expander.xaml" />
+    <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.FocusVisual.xaml" />
```
nicolaihenriksen, line 17: Shouldn't this addition also be added to the `MaterialDesign3.Defaults.xaml` for it to have any effect on the material design 3 styles (there is a separate demo app where you can test this)?

Probably also to `MaterialDesign2.Defaults.xaml` which I think is supposed to eventually replace `MaterialDesignTheme.Dafaults.xaml`. @Keboo correct me if I am wrong...

Yoooi0: Added

Yoooi0:
Another issue is MahApps...
It would also need to use `SystemParameters.FocusVisualStyleKey` otherwise the custom focus style will not work on MahApps styles.

I'm starting to not like the PR.
Why is it so hard to have global consistent focus visual, or so hard to disable it globally 😭

PR #3033, opened 2023-01-08
DataGrid style cleanup

### LINKED ISSUE #3009: Is it possible to embed ElementStyle & EditingElementStyle in Style?
DG1

**MD DataGrid, we have to set the column ElementStyle and EditingElementStyle in application view.
Is it possible to put everything in DataGrid Style below？**

DG2

nicolaihenriksen:
Fixes #3009

This PR introduces a new attached property, `DataGridAssist.ApplyMaterialDesignColumnStyles`, which works for both auto-generated and explicitly-defined columns. It still allows overriding of the style in XAML where needed. It works by inspecting the `ElementStyle`/`EditingElementStyle` of the columns in the `DataGrid` and applies the MDIX styles if the values have not been explicitly overridden.

This could possibly be interpreted as a "visual breaking change" as some columns that previously were not automatically styled, now are automatically styled (ComboBox).

It also introduces new attached properties, `DataGridAssist.AutoGeneratedComboBoxStyle` and `DataGridAssist.AutoGeneratedEditingComboBoxStyle`, similar to the existing ones for "text" and "checkbox" columns. This gives the user an ability to override the style for auto-generated `ComboBox` columns. These properties (along with the existing ones) are no longer needed if the intention is to simply override with the default MDIX styles, but could be useful if you have a completely custom style you would like to apply.


nicolaihenriksen:
@Keboo Could you have a look at this PR when you have time? I would like to hear if you see any pitfalls with this approach.

I have tried to keep it relatively simple and flexible without introducing breaking changes. I think I was successful in doing so, but I would like you to sanity check it 😄

nicolaihenriksen:
@Keboo And why are PRs now uploading artifacts and whatnot? Was that part of your pipeline changes? Thought it was only the "Release" PR that needed this.

image

nicolaihenriksen:
> @Keboo And why are PRs now uploading artifacts and whatnot? Was that part of your pipeline changes? Thought it was only the "Release" PR that needed this.

Nevermind, I confused myself. These are of course only uploaded to the particular build as artifacts.

MaterialDesignThemes.Wpf/DataGridAssist.cs
```diff
@@ -120,6 +120,148 @@ private static void SetGeneratedTextColumnEditingStyle(object? sender, DataGridA
...
+        bool applyElementStyle;
+        bool applyEditingElementStyle;
+        switch (column)
+        {
+            case DataGridCheckBoxColumn checkBoxColumn:
+                applyElementStyle = Equals(checkBoxColumn.ElementStyle, defaultElementStyle);
+                applyEditingElementStyle = Equals(checkBoxColumn.EditingElementStyle, defaultEditingElementStyle);
+                if (applyElementStyle)
+                {
+                    checkBoxColumn.ElementStyle =  (Style)Application.Current.TryFindResource("MaterialDesignDataGridCheckBoxColumnStyle");
```
Keboo, line 227: I think the general structure of this is ok. However, I think I would suggest leaving the setters for the attached properties on the DataGrid, and then in here, rather than using `TryFindResource` simply grab the attached property off of the DataGrid. This will be more performant as it is a single lookup, and should give reasonably discoverable options for consumers who want to replace the generated styles.

nicolaihenriksen: I am not opposed to that idea, however I have some concerns with the approach:

1. The attached properties, in their current state, are named **AutoGenerated**CheckboxStyle which very much indicates that they relate to the `DataGrid.AutoGenerateColumns` property. This is intended to also work for explicitly-defined columns without requiring the user to explicitly set the ElementStyle/EditingElementStyle on them. This is not a big issue though, and can probably be explained/documented.

2. The _ApplyMaterialDesignColumnStyleForColumn()_ method is called from the _Columns_CollectionChanged()_ callback. In here, we do not have a direct reference to the ("parent") `DataGrid` (i.e. the _sender_ is the `ObservableCollection` itself). This is the reason I opted for the _Application.Current.TryFindResource()_ approach. Thoughts on how to get a hold of the `DataGrid` in this scenario?

nicolaihenriksen: @Keboo So I think I solved the problem from bullet 2 above by introducing a `ColumnUpdater` type which simply holds on to a reference to the `DataGrid` and then the _Columns_CollectionChanged()_ callback is a method on this type. This did introduce a need for an internal attached property in order to do proper housekeeping/cleanup of that event handler to avoid memory leaks.

Still, something tells me this won't suffice, but I'm interested in hearing your input on the following:
When a `DataGrid` is created, we now default to setting an attached property ("ColumnUpdater") for that `DataGrid` instance. My concern is in the scenario where the `DataGrid` is unloaded and orphaned allowing the GC to collect it. Is the attached property a weak reference such that it can still be garbage collected, or will we need to handle Loaded/Unloaded events as well in order to clear the attached property (to allow garbage collection to cleanup the `DataGrid`)?

Have a look at the changes and let me know what your thoughts are. Thanks.

nicolaihenriksen: @Keboo I set this as ready for review, please note the comment above about a possible memory leak...

MaterialDesignThemes.Wpf/DataGridAssist.cs
```diff
@@ -120,6 +121,172 @@ private static void SetGeneratedTextColumnEditingStyle(object? sender, DataGridA
...
+        = DependencyProperty.RegisterAttached("ColumnUpdater", typeof(ColumnUpdater), typeof(DataGridAssist),
+            new PropertyMetadata(default(ColumnUpdater)));
+    private static void SetColumnUpdater(DependencyObject element, ColumnUpdater? value) => element.SetValue(ColumnUpdaterProperty, value);
+    private static ColumnUpdater? GetColumnUpdater(DependencyObject element) => (ColumnUpdater?) element.GetValue(ColumnUpdaterProperty);
+
+    private static void ApplyMaterialDesignColumnStylesPropertyChangedCallback(DependencyObject d, DependencyPropertyChangedEventArgs e)
+    {
+        var dataGrid = (DataGrid)d;
+        ColumnUpdater? columnUpdater = GetColumnUpdater(dataGrid);
+        if (columnUpdater is not { })
```
Keboo, line 202:
```suggestion
        if (columnUpdater is not null)
```

MaterialDesignThemes.Wpf/DataGridAssist.cs
Keboo: So I don't believe there is a memory leak here.
So just walking through this:
1. DataGrid is created
2. Style sets DataGridAssist.ApplyMaterialDesignColumnStyles to be true
3. The property change creates a new ``ColumnUpdater` instances
4. The DataGrid.Columns gets  reference to the `ColumnUpdater` via the event registration since the target method (`Columns_CollectionChanged`) is an instance method.
5. The `ColumnUpdater` is stored in an attached property on the DataGrid.

Since the DataGrid is the only thing holding a reference to the `ColumnUpdater` once the `DataGrid` is available for garbage collection, the `ColumnUpdater` should be cleaned up as well. 


With that said, is there any benefit to having the `ColumnUpdater` class? It appears that its purpose is to hold a reference to the `DataGrid` and then call the `ApplyMaterialDesignColumnStyleForColumn`. If that is true, then the `ApplyMaterialDesignColumnStylesPropertyChangedCallback` can be simplified as follows (which then eliminates the need for the `ColumnUpdater`):
```c#
private static void ApplyMaterialDesignColumnStylesPropertyChangedCallback(DependencyObject d, DependencyPropertyChangedEventArgs e)
{
    var dataGrid = (DataGrid)d;
    dataGrid.Columns.CollectionChanged -= Columns_CollectionChanged;
    if (Equals(true, e.NewValue))
    {
        dataGrid.Columns.CollectionChanged += Columns_CollectionChanged;    // Auto-generated columns are added later in the chain, thus we subscribe to changes.
        foreach (var column in dataGrid.Columns)
        {
            ColumnUpdater.ApplyMaterialDesignColumnStyleForColumn(dataGrid, column);
        }
    }

    void Columns_CollectionChanged(object? sender, NotifyCollectionChangedEventArgs e)
    {
        foreach (DataGridColumn column in e.NewItems?.OfType<DataGridColumn>() ?? Enumerable.Empty<DataGridColumn>())
        {
            ColumnUpdater.ApplyMaterialDesignColumnStyleForColumn(dataGrid, column);
        }
    }
}
```
After that, we could simply lift the `ApplyMaterialDesignColumnStyleForColumn` up into `DataGridAssit` as a private static method.

Overall I really like the work here. This is a great addition!

nicolaihenriksen:
@Keboo Didn't even think of using a local function to capture the dataGrid variable. MUCH nicer! Cleaned it up, rebased and force pushed.

MaterialDesignThemes.Wpf/DataGridAssist.cs
nicolaihenriksen: Thanks for pointing me in the right direction!

Keboo (approved):

PR #3074, opened 2023-02-04
TextBox: Respect externally set TextBox.Padding in triggers

### LINKED ISSUE #3058: TextBox-padding defined in ResourceDictionary not worked
### Bug explanation

If you set padding for a Textbox inline in style, its worked; but if you move the padding to a setter in ResourceDictionary, it's not working.

For example:
<Style x:Key="LInputTextField" TargetType="{x:Type TextBox}" BasedOn="{StaticResource MaterialDesignOutlinedTextBox}">
   <Setter Property="Padding" Value="5 7" /> Not work

but:
<TextBox  Style="{StaticResource MaterialDesignOutlinedTextBox}" Padding="5 7"> Worked

### Version

4.7.1.0

### LINKED ISSUE #3076: Material Design 3 TextBox Height
### Bug explanation

Hi, 
Why is not possible to put smaller height of textBox. I got smaller textBox but I can't write anything inside and also I don' see bottom line when the size is reduced. 

    <Style x:Key="NVDefaultTextBox" TargetType="TextBox"
           BasedOn="{StaticResource MaterialDesignOutlinedTextBox}">
        <Setter Property="FontSize" Value="16"/>
        <Setter Property="FontFamily" Value="Roboto"/>
        <Setter Property="FontWeight" Value="Medium"/>
        <Setter Property="Height" Value="32"/>
        <Setter Property="MinWidth" Value="48"/>
        <Setter Property="materialDesign:TextFieldAssist.DecorationVisibility" Value="Hidden"/>
        <Setter Property="materialDesign:TextFieldAssist.TextFieldCornerRadius" Value="4"/>
        <Setter Property="materialDesign:TextFieldAssist.UnderlineCornerRadius" Value="1" />
    </Style>

image

### Version

Material Design 3

nicolaihenriksen:
Fixes #3058 
Fixes #3076

This PR extracts the `TextBox.Padding` setters out of the trigger collection and into the overriden `Style` instead. This allows the user to override them.

There is a consequence to this however: you now **have** to set the style, you cannot only set the attached properties. This is a visual breaking change. You can see how I needed to modify the `MainWindow.xaml` to get the correct behavior for the `DemoItemsSearchBox`.

There is also another (internal) consequence, which is that any trigger or similar which previously relied on the "hardcoded" padding for a particular style, now needs to respect any margin because the user can override it.

There is also the fact that this is not limited to the `TextBox` styles. If we decide to go this route, it should be fixed for all filled/outlined styles in a similar manner.

In the example below, I have set `TextBox.Padding="5"` on the `SmartHint` demo page (via a style setter) to illustrate that it is now possible to override in a style.
image

### DIFF
--- a/MainDemo.Wpf/Domain/SmartHintViewModel.cs
+++ b/MainDemo.Wpf/Domain/SmartHintViewModel.cs
@@ -4,6 +4,7 @@ namespace MaterialDesignDemo.Domain;
 
 internal class SmartHintViewModel : ViewModelBase
 {
+    private bool _floatHint = true;
     private FloatingHintHorizontalAlignment _selectedAlignment = FloatingHintHorizontalAlignment.Inherit;
     private double _selectedFloatingScale = 0.75;
     private bool _showClearButton = true;
@@ -17,6 +18,16 @@ internal class SmartHintViewModel : ViewModelBase
     public IEnumerable<Point> FloatingOffsetOptions { get; } = new[] { new Point(0, -16), new Point(0, 16), new Point(16, 16), new Point(-16, -16) };
 
     public IEnumerable<string> ComboBoxOptions { get; } = new[] {"Option 1", "Option 2", "Option 3"};
+
+    public bool FloatHint
+    {
+        get => _floatHint;
+        set
+        {
+            _floatHint = value;
+            OnPropertyChanged();
+        }
+    }
 
     public FloatingHintHorizontalAlignment SelectedAlignment
     {
--- a/MainDemo.Wpf/MainWindow.xaml
+++ b/MainDemo.Wpf/MainWindow.xaml
@@ -60,13 +60,11 @@
                    Width="200"
                    Margin="16,4"
                    materialDesign:HintAssist.Hint="Search"
-                   materialDesign:HintAssist.IsFloating="True"
                    materialDesign:TextFieldAssist.DecorationVisibility="Collapsed"
                    materialDesign:TextFieldAssist.HasClearButton="True"
-                   materialDesign:TextFieldAssist.HasOutlinedTextField="True"
-                   materialDesign:TextFieldAssist.TextFieldCornerRadius="4"
                    DockPanel.Dock="Top"
-                   Text="{Binding SearchKeyword, UpdateSourceTrigger=PropertyChanged}" />
+                   Text="{Binding SearchKeyword, UpdateSourceTrigger=PropertyChanged}"
+                   Style="{StaticResource MaterialDesignOutlinedTextBox}"/>
 
           <ListBox x:Name="DemoItemsListBox"
                    Margin="0,16,0,16"
--- a/MainDemo.Wpf/SmartHint.xaml
+++ b/MainDemo.Wpf/SmartHint.xaml
@@ -40,7 +40,8 @@
 
     <GroupBox Grid.Row="0" Margin="8,0,16,16" Header="Hint Settings" Style="{StaticResource MaterialDesignCardGroupBox}">
       <StackPanel Orientation="Horizontal" Margin="10">
-        <TextBlock Text="FloatingHintHorizontalAlignment:" VerticalAlignment="Center" />
+        <CheckBox Content="IsFloating" IsChecked="{Binding FloatHint}" />
+        <TextBlock Text="Alignment:" VerticalAlignment="Center" Margin="20,0,0,0" />
         <ComboBox VerticalAlignment="Center" Margin="5,0,0,0" ItemsSource="{Binding HorizontalAlignmentOptions}" SelectedItem="{Binding SelectedAlignment, Mode=TwoWay}" />
         <TextBlock Text="FloatingScale:" VerticalAlignment="Center" Margin="20,0,0,0" />
         <ComboBox VerticalAlignment="Center" Margin="5,0,0,0" ItemsSource="{Binding FloatingScaleOptions}" SelectedItem="{Binding SelectedFloatingScale, Mode=TwoWay}" ItemStringFormat="F2" />
@@ -84,6 +85,7 @@
               <Setter Property="materialDesign:TextFieldAssist.HasClearButton" Value="{Binding ShowClearButton}" />
               <Setter Property="materialDesign:TextFieldAssist.HasLeadingIcon" Value="{Binding ShowLeadingIcon}" />
               <Setter Property="materialDesign:TextFieldAssist.LeadingIcon" Value="{StaticResource LeadingIcon}" />
+              <Setter Property="materialDesign:HintAssist.IsFloating" Value="{Binding FloatHint}" />
             </Style>
           </Grid.Resources>
           <Grid.ColumnDefinitions>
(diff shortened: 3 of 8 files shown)


nicolaihenriksen:
@Keboo I converted it back to a draft because, although it solves the problem, I am not really happy with the solution.

Reason being that it violates the "principle of least astonishment". `TextBox.Padding` really should be an exact value representing the padding; not a partial one as it is in this PR.

The problem is that the triggers override the padding for the filled/outlined styles effectively ignoring any value set by the user. We could add attached properties for these paddings allowing the user to control the padding for these styles, but then it is still sort of strange that `TextBox.Padding` does nothing and the user would need to use the AP to set this for some styles and not for others.

Do you have a good idea on how to make the `TextBox.Padding` the only source of truth and still allowing the user to control it?

Keboo:
That is a good question. I do agree that combining two Thickness values together is not expected behavior. I am wondering with the 5.0.0 release if it might be worth getting rid of the attached properties and instead split the templates. Though I realize this is significantly more work.

nicolaihenriksen:
@Keboo See my comment above. Do you see this as a reasonable approach? If not, I won't bother implementing it for the remaining templates.

Keboo:
Hi @nicolaihenriksen sorry, I lost track of the notification on this one. YES! I completely overlooked that style of solution. You are absolutely correct, that if we use control template triggers they will make it quite difficult for people to override the values. I think this is an excellent solution.

nicolaihenriksen:
> YES! I completely overlooked that style of solution .... I think this is an excellent solution.

@Keboo I applied the same changes for the `PasswordBox` (also reveal style) and the `ComboBox`.

The `DatePicker` and `TimePicker` also contain a `TextBox` where I assumed a similar fix was needed, but the padding applied to the contained `TextBox` is done using a converter there, and I have checked that overriding the padding in an overridden style for the `DatePicker`/`TimePicker` works as expected.

Keboo (approved):

nicolaihenriksen:
> and I have checked that overriding the padding in an overridden style for the `DatePicker`/`TimePicker` works as expected

@Keboo 
Turns out that was a lie. Apparently I only checked it with an explicit padding applied on the element directly.

Also, there are some alignment issues with respect to the helper text. I think I may open up a "cleanup" PR to fix the padding issue for `DatePicker`/`TimePicker` and attack some of those helper text issues.

Keboo:
> > and I have checked that overriding the padding in an overridden style for the `DatePicker`/`TimePicker` works as expected
> 
> @Keboo Turns out that was a lie. Apparently I only checked it with an explicit padding applied on the element directly.
> 
> Also, there are some alignment issues with respect to the helper text. I think I may open up a "cleanup" PR to fix the padding issue for `DatePicker`/`TimePicker` and attack some of those helper text issues.

That would be awesome. Thank you!

nicolaihenriksen:
> That would be awesome. Thank you!

@Keboo I think I may be missing some MDIX historical knowledge here, perhaps you can help me out. It is not entirely clear to me what the purpose of `TextFieldAssist.TextBoxViewMargin` is or was. The naming suggests something about being able to add a custom margin to the content part (e.g. `PART_ContentHost`) of a `TextBox`, possibly in order to ensure correct alignment of the hint text. However, it does not seem to be consistently applied to "content" across the different templates. Do you know its purpose in life? Is it really needed? Since I have refactored how the `SmartHint` is placed in some of the recent PRs, I may actually have introduced some regression in this area?!

Also, it kind of conflicts with `TextBox.Padding` which similarly is used to inject some space around the "content" of the `TextBox`. The AP is public, so not something that can simply be removed, but we could obsolete it and "make it do nothing". There may be some subtleties that I am overlooking where certain templates need minor tweaks to align hint and helper text correctly with the content, but in any case that should still respect the `TextBox.Padding`. So having the attached property as well just makes for one more variable that needs to be considered when placing the hint and helper text elements. Perhaps it would be easier to simply cut it out of the loop and only consider `TextBox.Padding`? We would probably need a converter for this (since it is most often only `Margin.Left` that is set to match the `Padding.Left`; not the whole `Thickness`). That converter could also take parameters to account for the special cases where a minor adjustment may be needed. Thoughts?

The [values currently being used for the attached property](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/231ce197b30bb98af5c69d99cf1207b4098bcf62/MaterialDesignThemes.Wpf/Constants.cs#L8-L9) are contributing to some of the slight misalignments of the hint and helper text that we're currently seeing. It is not a whole lot so I could do the PR without touching this AP, but I just worry that everything falls apart if a user sets a custom value for the AP because it is not respected/applied consistently.

Keboo:
I am going to spend some time on my stream tonight going through this. I agree this is quite complicated. I am not sure of the right answer, and I think I need to spend some time playing with the changes in your pull request.

nicolaihenriksen:
> I think I need to spend some time playing with the changes in your pull request. 

The changes currently in there only fix the overall scheme: hint and helper text should also follow padding (at least it seems so when looking at material.io).

I believe I know how to fix the remaining "minor misalignments" but don't have time to finish it before your stream.

Looking forward to see your stream.

Keboo:
Awesome. I will probably still spend some time playing with it just to make sure I understand everything. Really great work as always!

nicolaihenriksen:
> Awesome. I will probably still spend some time playing with it just to make sure I understand everything. Really great work as always!

If you could look into whether you agree with ny assesment that TextFieldAssist.TextBoxViewMargin should only be used to manipulate PART_ContentHost (and similar) and NOT hints, helpers or validation, that would be awesome. Thats is at least what I think is the right approach. The remaining part is then about ensuring left edge of all those elements align nicely.

Another topic is the placement of helper and validation when HorizontalContentAlignment is Center or Right...

Keboo:
> If you could look into whether you agree with ny assesment that TextFieldAssist.TextBoxViewMargin should only be used to manipulate PART_ContentHost (and similar) and NOT hints, helpers or validation, that would be awesome. Thats is at least what I think is the right approach. The remaining part is then about ensuring left edge of all those elements align nicely.

Yes, I believe only using `TextFieldAssist.TextBoxViewMargin` for the content host is correct.

> Another topic is the placement of helper and validation when HorizontalContentAlignment is Center or Right...

I think those can be separate PRs, but I think they should follow the HorizontalContentAlignment. We should also consider FlowDirection.

nicolaihenriksen:
> Yes, I believe only using `TextFieldAssist.TextBoxViewMargin` for the content host is correct.

Perfect, I will remove the "wrong usages" of it then.

> I think those can be separate PRs, but I think they should follow the HorizontalContentAlignment. We should also consider FlowDirection.

Agree. I will create separate PRs for that and will also consider FlowDirection. Hopefully these templates (or at least the areas regarding hint/helper) are simple enough that "it just works", but I will be sure to test it.

On your stream you were debating whether an AP should be created to provide extra control over the placement of the hint/helper/validation. I agree this is a good idea; in fact I already added that for the hint in a previous PR - although that AP only affects the positioning when the hint floats. The AP is basically an `enum` with the values `Inherit`, `Left`, `Center`, `Right`, `Stretch` where `Inherit` is the default and this basically means it will follow the `HorizontalContentAlignment`.

Another minor thing. The demo app has both the `Fields Line Up` and the `Smart Hint` demo pages which are to some extent similar. I didn't actually change the `Fields Line Up` page, I just used it to demo the issues with the `TextFieldAssist.TextBoxViewMargin` AP. Anyway I think my other PR is ready for review now, so check it out and give the `Smart Hint` demo page a run.

PR #3092, opened 2023-02-17
Fixes DialogHost visibility on VM disconnect (#3069)

### LINKED ISSUE #3069: DialogHost | Dialog no longer visible if disconnected-then-reconnect to VM running application
### Bug explanation

# Steps to reproduce
Used a hyper-v VM running Win 10 21H2 build 19044.1466 + the MainDemo.Wpf app latest commit [607ee6a](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/commit/607ee6a0c917d0bdb5f26c5575d119fd6c6498e0):

1. Transfer the MainDemo.Wpf app to the VM
2. Open the MainDemo.Wpf app and navigate to Dialogs
3. Click "Run Code" under Sample 3
4. While the dialog is visible, close the window which disconnects from the VM but leave the VM running
5. Reconect to the VM and login.
6. Notice that the dialog is no longer visible, the obscured background of the app is still greyed out and present, and nothing in the app being overlaid by the DialogHost responds to user input.

**Screenshot for step 3 above:**
image

**Screenshot for step 6 above:**
image

The short story of them problem seems to be: If the user gets disconnected from a remote session AND a dialog is visible when the disconnect occurs, then the user on reconnect will see a greyed out screen and no visible dialog on reconnect.  Anyone ran into this and is there a known workaround?  

### Version

v4.7.1

tblong:
Fixes #3069 

On reconnect to a VM where a `DialogHost`'s popup was previously visible, the `DialogHost.IsOpen` property would remain `true` in the `DialogHost.Loaded` event handler however the popup's `IsOpen` property would be `false`.  This fix ensures the two properties match and the position of the popup remains consistent on reconnect to the VM where the application is running.

### DIFF
--- a/MaterialDesignThemes.UITests/Samples/DialogHost/LoadAndUnloadControl.xaml
+++ b/MaterialDesignThemes.UITests/Samples/DialogHost/LoadAndUnloadControl.xaml
@@ -1,0 +1,29 @@
+﻿<UserControl x:Class="MaterialDesignThemes.UITests.Samples.DialogHost.LoadAndUnloadControl"
+             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
+             xmlns:d="http://schemas.microsoft.com/expression/blend/2008" 
+             xmlns:local="clr-namespace:MaterialDesignThemes.UITests.Samples.DialogHost"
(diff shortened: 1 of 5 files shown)


tblong:
Just asking to see whether it might be considered to get this included in a v4.7.2 patch if you think v4.8 might be a while.

Keboo (approved):

Keboo:
Hi @tblong thank you very much for this. I threw in a UI test to cover this to ensure we don't break this behavior in the future.
As for a release, I am planning on get 4.8.0 out fairly soon (possibly in the next week). If you need it sooner, this fix should go out in tonight's preview NuGet package for 4.8.0

PR #3096, opened 2023-02-21
Fix PasswordBox keyboard focus traversal issue

### LINKED ISSUE #3095: PasswordBox (revealed password) does not show IBeam/caret when tabbing into PasswordBox
### Bug explanation

When tabbing (Tab key) into a `PasswordBox` which is set to reveal the password, the "clear text" password field does not receive keyboard focus. The `PasswordBox` changes it state to "Focused", but the IBeam/caret does not show up in the `TextBox`.

PasswordBoxFocusIssue

In comparison, it does get the IBeam/caret when the password **is not** revealed:

PasswordBoxFocusIssue2

There shouldn't be a (visual) difference in this, seen from a users perspective.

### Version

4.7.1

nicolaihenriksen:
Fixes #3095 

This PR fixes the keyboard focus traversal associated with the "reveal style" password box. See the associated issue for details.

I updated the Smart Hint demo page to easily test this out.

I also added a UI test to verify the fix as well, however I was only able to get the "forward tabbing" to work. Either I don't know the correct syntax, or XAMLTest does not behave nicely with the special codes for SHIFT/CTRL/ALT modifiers which are mentioned approximately halfway down on this page:

https://learn.microsoft.com/dotnet/api/system.windows.forms.sendkeys.send
(I know this is Windows Forms link, but I assume XAMLTest is just using the same underlying P/Invoke calls?)

I left the "backwards tabbing" in the test, but commented out the assertions with FIXMEs attached. Perhaps you know the correct syntax to get this working?

### DIFF
--- a/Directory.packages.props
+++ b/Directory.packages.props
@@ -22,7 +22,7 @@
     <PackageVersion Include="ShowMeTheXAML.AvalonEdit" Version="2.0.0" />
     <PackageVersion Include="ShowMeTheXAML.MSBuild" Version="2.0.0" />
     <PackageVersion Include="VirtualizingWrapPanel" Version="1.5.7" />
-    <PackageVersion Include="XAMLTest" Version="1.0.0-ci390" />
+    <PackageVersion Include="XAMLTest" Version="1.0.0-ci400" />
     <PackageVersion Include="xunit" Version="2.4.2" />
     <PackageVersion Include="xunit.runner.visualstudio" Version="2.4.5" />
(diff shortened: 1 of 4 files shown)


Keboo (approved):

nicolaihenriksen:
@Keboo I changed the UI test code to match the syntax I use in the PR I have on XAMLTest. The test now fails, but it should highlight what I would like to do, and it should pass if the XAMLTest PR is merged and new version pulled in here.

Keboo (approved):

PR #3102, opened 2023-02-24
Possible fix for issue 3094 regarding TabControl.SelectedIndex via DialogHost

### LINKED ISSUE #3094: TabControl.SelectedIndex via DialogHost (repo)
### Bug Description
When opening a dialog `await DialogHost.Show(view);` via **MenuItem** (System.Windows.Controls) and setting the `SelectedIndex` of a TabControl by e.g. clicking a button, the selected tab is still the _old_ one after the dialog has been closed.
The property that holds the `SelectedIndex` receives the new value but instantly switches back to the previous index, which can result in a _flicker_.

### Reproduction
In order to reproduce the bug, please select the **second tab** (**Info**). Then click on **Menu** in the top right corner. Select **Settings** and click on the button in the dialog.

### Debugging
I have found out that it does make a difference whether the dialog is opened via said **MenuItem**, a **PopupBox** (Material Design) or a button on any **view**:
It does work es expected in both of the last two cases, just the MenuItem-way fails.
In order to make it work, insert `await Task.Delay(500);` (at least 200-300) before setting `SelectedIndex` to its new value. I'm not a fan of using `Task.Delay()` but currently out of ideas how to make it work without it.

Using DI does not make a difference in any of the above mentioned behaviors.

### Installed NuGet packages:
- CommunityToolkit.Mvvm (8.1.0)
- MaterialDesignThemes (4.7.1)
- Microsoft.Extensions.DependencyInjection (7.0.0)
- Microsoft.Extensions.Hosting (7.0.1)

### Link
[GitHub Repository](https://github.com/VR462002/MDIX-DialogHost-TabControl)

### Version

4.7.1 (latest stable)

nicolaihenriksen:
(Potentially) Fixes #3094 

This PR contains 2 commits. The first commit is simply adding in the demo project from the issue which can be used to reproduce the bug. The second commit is the actual fix **and the only commit that should be merged if the fix is deemed acceptable**. It is also the demo project from the first commit that makes it fail in the GitHub pipeline.

The issue is that the `DialogHost`, when trying to restore focus back to the "origin" from where it was opened, is doing so by pushing a message into the message pump with a priority of `DispatcherPriority.Input` which - probably - is too high. Lowering the priority down to `DispatcherPriority.Normal` fixes the issue.

The original [commit](d238ea05c23a699d5ae476014986003562faa913) that introduced this restore functionality was using `DispatcherPriority.Input` and has not been touched since. The commit does not have a link to an issue or feature or anything, so it is difficult to know whether the priority was intentionally chosen this way. I doubt that a change like this has any consequences, but I cannot be 100% sure.


nicolaihenriksen:
Abandoning this for a moment, I had added a small change to the demo app which is what was "helping it" work appropriately. Will keep working on it and post a proper fix if I find one.

PR #3103, opened 2023-02-24
Add DialogHost.DialogContentUniformCornerRadius

nicolaihenriksen:
Addresses #3088 

Allows users of `DialogHost` to specify a custom (uniform) corner radius for the nested `Card`.

An alternative approach could be to introduce a `CardAssist.UniformCornerRadius` attached property; although it would clash a little with the already defined `Card.UniformCornerRadius` normal dependency property.

image

### DIFF
--- a/MaterialDesignThemes.Wpf/DialogHost.cs
+++ b/MaterialDesignThemes.Wpf/DialogHost.cs
@@ -432,6 +432,15 @@ namespace MaterialDesignThemes.Wpf
             set => SetValue(DialogContentProperty, value);
         }
 
+        public static readonly DependencyProperty DialogContentUniformCornerRadiusProperty = DependencyProperty.Register(
+            nameof(DialogContentUniformCornerRadius), typeof(double), typeof(DialogHost), new PropertyMetadata(4d));
+
+        public double DialogContentUniformCornerRadius
+        {
+            get { return (double)GetValue(DialogContentUniformCornerRadiusProperty); }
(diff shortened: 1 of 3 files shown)


MaterialDesignThemes.Wpf/DialogHost.cs
```diff
@@ -432,6 +432,15 @@ public CornerRadius CornerRadius
             set => SetValue(DialogContentProperty, value);
         }
 
+        public static readonly DependencyProperty DialogContentUniformCornerRadiusProperty = DependencyProperty.Register(
+            nameof(DialogContentUniformCornerRadius), typeof(double), typeof(DialogHost), new PropertyMetadata(4d));
+
+        public double DialogContentUniformCornerRadius
+        {
+            get { return (double)GetValue(DialogContentUniformCornerRadiusProperty); }
+            set { SetValue(DialogContentUniformCornerRadiusProperty, value); }
```
Keboo, line 441: nit: I have been favoring using labda-bodied members for these.
```suggestion
            get => (double)GetValue(DialogContentUniformCornerRadiusProperty);
            set => SetValue(DialogContentUniformCornerRadiusProperty, value);
```

Keboo (approved):

MaterialDesignThemes.Wpf/DialogHost.cs
nicolaihenriksen: And so have I, but I guess laziness got the best of me. This was my IDE generating the dependency property boiler-plate code for me 🤣

PR #3104, opened 2023-02-25
Properly restore focus when TabControl.SelectedIndex is set via DialogHost

### LINKED ISSUE #3094: TabControl.SelectedIndex via DialogHost (repo)

nicolaihenriksen:
Fixes #3094

As per the discussion thread below, this PR adds `DialogHost.RestoreFocusElement` and `DialogHost.IsRestoreFocusDisabled` attached/dependency properties.

The former is currently used for the `TabItem` style (also the one for the "rail item") to ensure focus is returned to the parent `TabControl` rather than the `TabItem` itself.

The latter is used to completely disable the restore focus feature in case that is relevant.

Added 4 UI tests to verify the behavior.


Keboo:
After giving it some thought here would be my proposal.

1. Create a new `DialogHost.RestoreFocusElement` (perhaps with a better name)
2. When setting `DialogHost._restoreFocusDialogClose` first check that element to see if the above attached property has a value, if it does use it for the `_restoreFocusDialogClose` value.
3. In the style for the TabItem (`MaterialDesignTabItem`) we do a relative binding looking for a parent tab control (something like: `{Binding RelativeSource={RelativeSource FindAncestor, AncestorType=TabControl}}`). We may need to look at other styles as well, such as `MaterialDesignNavigationRailTabItem`.
4. Optionally, we could consider a mechanism for turning off the restore focus behavior (I say optional because rarely has anyone ever requested this). But we could create a static property on DialogHost (something like `DialogHost.DisableRestoreFocus` or a better name). Then in step 2 above if that is the value of the attached property, we would set `DialogHost._restoreFocusDialogClose` to null to effectively disable the behavior.

Thoughts?

nicolaihenriksen:
> After giving it some thought here would be my proposal.
> ...
> Thoughts?

I think this sound like a very good approach. I very much like the idea that the responsibility for picking the correct element to focus is removed from the `DialogHost` and put into the respective controls themselves. This aligns nicely with the Tell Don't Ask principle. Also, it is probably a rather limited number of controls which face this type of (possible) issue.

Having the additional option to completely disable the restore behavior also seems like a good idea to me. We'll just default the property to `false` such that it is something a caller would have to explicitly request for the `DialogHost`.

Writing UI tests to ensure this behavior works may or may not prove to be a non-trivial task. I'll spend a little time to see if I can write a (failing) UI test for the scenario included in the sample repro project included in this branch.

Keboo:
This is great. I think there is a bit of a disjoint on the names of the properties. (FocusRestore vs RestoreFocus). I think `IsFocusRestoreDisabled` should probably be the one changed to `IsRestoreFocusDisabled`. Thoughts?

nicolaihenriksen:
> This is great. I think there is a bit of a disjoint on the names of the properties. (FocusRestore vs RestoreFocus). I think `IsFocusRestoreDisabled` should probably be the one changed to `IsRestoreFocusDisabled`. Thoughts?

I agree, I'll change it.

VR462002:
Hi Nicolai and Kevin,

thank you for working on this, you guys are awesome!

@nicolaihenriksen Thanks for testing and explaining the focus problem!
@Keboo Always appreciate your input and enjoy your live videos/recordings

Keboo (approved):

PR #3128, opened 2023-03-08
Added SliderAssist.HideActiveTrack

JorisCleVR:
Added SliderAssist.ShowActiveTrack to support the ability to hide the active track of a slider.

In the screenshot of the demo the last slider of each GroupBox shows the result.

image

### DIFF
--- a/MainDemo.Wpf/Sliders.xaml
+++ b/MainDemo.Wpf/Sliders.xaml
@@ -72,7 +72,7 @@
           <Slider Maximum="50"
                   Minimum="0"
                   Value="35"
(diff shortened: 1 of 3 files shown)


JorisCleVR:
Note that I was also considering to name the property HideActiveTrack instead of ShowActiveTrack.
When preferred I could rename it to HideActiveTrack instead.

Keboo:
This looks great. I do think changing the name to `HideActiveTrack` would be a little cleaner.

JorisCleVR:
The property is now renamed to HideActiveTrack.

Keboo (approved):

PR #3133, opened 2023-03-10
Unregister hooks in ScrollViewerAssist when the ScrollViewer is unloaded

### LINKED ISSUE #3130: Potential memory leak in ScrollViewerAssist
### Bug explanation

Views containing a `TabControl` using any style based on `MaterialDesignTabControlBase` are never garbage collected due to a problem in `ScrollViewerAssist` when the attached property `SupportHorizontalScroll` is enabled.

`ScrollViewerAssist` registers a `HwndSourceHook` to handle Win32 messages but never unregisters it.

This shows the memory snapshot from my application (class names obfuscated for reasons of paranoia and over-zealous corporate IT security).
image

PR to follow!

### Version

4.8.0

chris-crowther:
Fixes #3130.

This covers the necessary clean-up of registered Win32 hooks and event handlers for the `SupportHorizontalScroll` and `BubbleVerticalScroll` attached properties.

To unregister the `HwndSourceHook` we have to capture the `HwndSource` in the Loaded event. I'm happy to consider an alternative approach.

I was tempted to promote the duplicated `OnLoaded` and `OnUnloaded` local methods to instance level methods as they are not property-specific.

### DIFF
--- a/MaterialDesignThemes.Wpf/ScrollViewerAssist.cs
+++ b/MaterialDesignThemes.Wpf/ScrollViewerAssist.cs
@@ -71,6 +71,9 @@ public static class ScrollViewerAssist
     private static readonly DependencyProperty HorizontalScrollHookProperty = DependencyProperty.RegisterAttached(
         "HorizontalScrollHook", typeof(HwndSourceHook), typeof(ScrollViewerAssist), new PropertyMetadata(null));
 
+    private static readonly DependencyProperty HorizontalScrollSourceProperty = DependencyProperty.RegisterAttached(
+        "HorizontalScrollSource", typeof(HwndSource), typeof(ScrollViewerAssist), new PropertyMetadata(null));
+
     public static readonly DependencyProperty SupportHorizontalScrollProperty = DependencyProperty.RegisterAttached(
         "SupportHorizontalScroll", typeof(bool), typeof(ScrollViewerAssist), new PropertyMetadata(false, OnSupportHorizontalScrollChanged));
 
@@ -88,6 +91,8 @@ public static class ScrollViewerAssist
                         RegisterHook(sv);
                     }
                 });
+
+                OnUnloaded(scrollViewer, RemoveHook);
             }
             else
             {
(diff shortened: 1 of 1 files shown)


MaterialDesignThemes.Wpf/ScrollViewerAssist.cs
```diff
@@ -119,10 +124,24 @@ static void OnLoaded(ScrollViewer scrollViewer, Action<ScrollViewer> doOnLoaded)
             }
         }
 
+        static void OnUnloaded(ScrollViewer scrollViewer, Action<ScrollViewer> doOnUnloaded)
+        {
+            if (!scrollViewer.IsLoaded)
```
nicolaihenriksen, line 129: Sorry for barging in on this code review; I was curious. 😃 

This guard is a bit suspicious to me. Why do we only want to listen for the Unloaded event if the `ScrollViewer` is not already loaded? I think the situation where it IS loaded is equally relevant. This would be a scenario where the AP is not statically set in XAML, but applied at runtime after the `ScrollViewer` has loaded. I acknowledge this may not be the most common use case, but it is still possible. In short, I think we should handle both the `ScrollViewer.IsLoaded=true/false` cases.

Another thing to consider is using weak event handlers. Both the `Loaded` and `Unloaded` event handlers are currently registered as strong event handlers which unregister themselves when the corresponding event fires. However, I think there are scenarios where this will not work - not particularly related to your changes though:

If the `ScrollViewer` resides in a `TabÌtem`, it will be loaded and unloaded multiple times when switching between the tabs. The current implementation only seems to wire up the hooks when the AP changes which means it will register for the `Loaded` event at startup. However, after navigating away from the `TabItem` and back to it again, the `Unloaded` event will fire, removing the hook, and then there is nothing listening for the "second" `Loaded` event to re-register the hook. Perhaps this could be mitigated by simply using weak event handlers and never unregistering them?

This comment also applies for the changes to the other AP.

Thoughts?

chris-crowther: The guard is somewhat pointless. As the handler is only fired after the `ScrollViewer` has been unloaded `scrollViewer.IsLoaded` will never be true. I'm happy to remove that guard to simplify the code.

As to your other points, weak event handlers are certainly worth considering.
As the code stands it handles the case when the tab is loaded and unloaded multiple times, as the `PropertyChangedCallback` methods are fired every time the control is loaded (and the properties are re-attached).
I have tested this in the demo app.

Keboo (approved):

Keboo:
Awesome. Thank you for putting this together

PR #3136, opened 2023-03-11
Refactor PopupBox placement to respect configurable elevation

### LINKED ISSUE #3129: PopupBox adding shadow
I want to create a custom control where it shows a `PopupBox` with `Calendar` on click. It's working fine except I want to add more shadow to emphasize the PopupBox. Unfortunately, using `md:ElevationAssist.Elevation="Dp8"` doesn't work on PopupBox.

Is there a way I can add more shadow to it?

### Version

4.5.0

nicolaihenriksen:
Fixes #3129

I set out to simply add `ElevationAssist.Elevation` support for the `PopupBox`, but I ended up opening "a can of worms".

**UPDATE**: After discussing this, we agreed not to use `ElevationAssist.Elevation`, but rather create a dedicated `PopupBox.Elevation`.

When applying a drop shadow to the `Card` contained in the `PopupBox`, the `Popup` that is displayed, needs to reserve space for said drop shadow; it cannot draw outside its bounds. To solve this problem, a `Margin` matching the drop shadow size is applied to the `Card` and this is then compensated for in the placement of the `Popup`.

The `PopupBox` also supports custom offsets of the `Popup` set by the calling code. Previously these were directly applied to the `Popup.HorizontalOffset`/`Popup.VerticalOffset` properties, and were then needed to be taken into consideration when calculating the custom popup placement. This made the calculations a bit funky; in fact I don't believe they were correct in all cases. It definitely did not support DPI scaling correctly, and I also believe it had issues with some alignments in `FlowDirection=RightToLeft`.

This PR refactors this such that the `Popup.HorizontalOffset`/`Popup.VerticalOffset` are **only** used to compensate for the drop shadow size. The custom offsets that can be applied (`PopupBox.PopupHorizontalOffset` and `PopupBox.PopupVerticalOffset`) are now included in the calculations in the `PopupBox.GetPopupPlacement()` callback instead. This callback is also refactored to adjust for DPI scaling and respect `FlowDirection` for all alignment cases.

I added a dedicated page in the demo app for the `PopupBox` to allow for manual testing of the various scenarios.

PopupBoxDemo

### DIFF
--- a/MainDemo.Wpf/PopupBox.xaml
+++ b/MainDemo.Wpf/PopupBox.xaml
@@ -1,0 +1,198 @@
+﻿<UserControl x:Class="MaterialDesignDemo.PopupBox"
+             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
+             xmlns:d="http://schemas.microsoft.com/expression/blend/2008" 
+             xmlns:local="clr-namespace:MaterialDesignDemo"
+             xmlns:materialDesign="http://materialdesigninxaml.net/winfx/xaml/themes"
+             xmlns:system="clr-namespace:System;assembly=System.Runtime"
+             mc:Ignorable="d" 
+             d:DesignHeight="450" d:DesignWidth="800">
+  <UserControl.Resources>
+    <ResourceDictionary>
+      <!-- This is needed to avoid runtime exceptions?! Seems like this might be a bug? -->
+      <ResourceDictionary.MergedDictionaries>
+        <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.PopupBox.xaml" />
+      </ResourceDictionary.MergedDictionaries>
+
+      <DataTemplate x:Key="ContentTemplateGrid">
+        <Grid Width="200" Height="100">
+          <TextBlock Text="Popup content in grid" HorizontalAlignment="Center" VerticalAlignment="Center">
+            <!-- This explicit style should not be needed; possibly because the MaterialDesignMultiFloatingActionPopupBox style enforces special brushes? Or perhaps an issue with correctly applying theming? -->
+            <TextBlock.Style>
+              <Style TargetType="TextBlock" BasedOn="{StaticResource MaterialDesignTextBlock}">
+                <Setter Property="TextElement.Foreground" Value="{StaticResource MaterialDesignLightForeground}" />
+              </Style>
+            </TextBlock.Style>
+          </TextBlock>
+        </Grid>
+      </DataTemplate>
+
+      <DataTemplate x:Key="ContentTemplateGridWithBackground">
+        <Grid Width="200" Height="100" Background="Fuchsia">
+          <TextBlock Text="Popup content in colored grid" HorizontalAlignment="Center" VerticalAlignment="Center" />
+        </Grid>
+      </DataTemplate>
+
+      <DataTemplate x:Key="ContentTemplateButtonStack">
+        <!-- Margin of 10 here is to "make room" for the elevation drop shadows of the buttons. Must be compensated for in some (left/right) alignment scenarios using PopupBox.PopupHorizontalOffset -->
+        <StackPanel Margin="10">
+          <Button Content="1"
+                  Opacity="0.5"
+                  ToolTip="One with custom opacity" />
+          <Button Content="2" ToolTip="Two" />
+          <Button Content="3" ToolTip="Three" />
+        </StackPanel>
+      </DataTemplate>
+
+      <local:ComboBoxItemToDataTemplateConverter x:Key="ComboBoxItemToDataTemplateConverter" />
+      <local:ComboBoxItemToStyleConverter x:Key="ComboBoxItemToStyleConverter" />
+      <local:ComboBoxItemToHelperTextConverter x:Key="ComboBoxItemToHelperTextConverter" />
+
+      <Thickness x:Key="Spacer">0,30,0,0</Thickness>
+    </ResourceDictionary>
+  </UserControl.Resources>
+  <Grid>
+    <Grid.ColumnDefinitions>
+      <ColumnDefinition Width="*" />
+      <ColumnDefinition Width="250" />
+    </Grid.ColumnDefinitions>
+
+    <Grid Grid.Column="0">
+      <Grid.RowDefinitions>
+        <RowDefinition Height="Auto" />
+        <RowDefinition Height="*" />
+      </Grid.RowDefinitions>
+
+      <TextBlock Grid.Row="0" HorizontalAlignment="Center" Style="{StaticResource MaterialDesignSubtitle2TextBlock}"
+                 Text="{Binding ElementName=ContentComboBox, Path=SelectedItem, Converter={StaticResource ComboBoxItemToHelperTextConverter}}" />
+
+      <materialDesign:PopupBox Grid.Row="1" VerticalAlignment="Center" HorizontalAlignment="Center" SnapsToDevicePixels="True" Padding="0"
+                               Style="{Binding ElementName=StyleComboBox, Path=SelectedItem, Converter={StaticResource ComboBoxItemToStyleConverter}}"
+                               materialDesign:ElevationAssist.Elevation="{Binding ElementName=ElevationComboBox, Path=SelectedItem}"
+                               PopupUniformCornerRadius="{Binding ElementName=UniformCornerRadiusComboBox, Path=SelectedItem}"
+                               PopupHorizontalOffset="{Binding ElementName=HorizontalOffsetComboBox, Path=SelectedItem}"
+                               PopupVerticalOffset="{Binding ElementName=VerticalOffsetComboBox, Path=SelectedItem}"
+                               PlacementMode="{Binding ElementName=PopupBoxPlacementModeComboBox, Path=SelectedItem}">
+        <ContentControl ContentTemplate="{Binding ElementName=ContentComboBox, Path=SelectedItem, Converter={StaticResource ComboBoxItemToDataTemplateConverter}}" />
+      </materialDesign:PopupBox>
+    </Grid>
+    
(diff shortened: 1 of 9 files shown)


MainDemo.Wpf/PopupBox.xaml
```diff
@@ -0,0 +1,198 @@
...
+             xmlns:materialDesign="http://materialdesigninxaml.net/winfx/xaml/themes"
+             xmlns:system="clr-namespace:System;assembly=System.Runtime"
+             mc:Ignorable="d" 
+             d:DesignHeight="450" d:DesignWidth="800">
+  <UserControl.Resources>
+    <ResourceDictionary>
+      <!-- This is needed to avoid runtime exceptions?! Seems like this might be a bug? -->
+      <ResourceDictionary.MergedDictionaries>
+        <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.PopupBox.xaml" />
+      </ResourceDictionary.MergedDictionaries>
```
nicolaihenriksen, line 16: I was quite surprised that I had to explicitly include the `MaterialDesignTheme.PopupBox.xml` for this to work. If left out, it fails to find the (default) Style for the `PopupBox`. I would assume this was already loaded in `App.Xaml` and therefore should not be needed here?

MainDemo.Wpf/PopupBox.xaml
```diff
@@ -0,0 +1,198 @@
...
+
+      <DataTemplate x:Key="ContentTemplateGrid">
+        <Grid Width="200" Height="100">
+          <TextBlock Text="Popup content in grid" HorizontalAlignment="Center" VerticalAlignment="Center">
+            <!-- This explicit style should not be needed; possibly because the MaterialDesignMultiFloatingActionPopupBox style enforces special brushes? Or perhaps an issue with correctly applying theming? -->
```
nicolaihenriksen, line 21: This was also a bit of a surprise to me. When the `PopupBox.Style` is set to `MaterialDesignMultiFloatingActionPopupBox`, it seems like something strange is going on with the content of the `Popup` that it shows. If this style override is left out, the `TextBlock` will have a white foreground, and thus is "seemingly" invisible (in the light theme of the demo app).

MainDemo.Wpf/PopupBox.xaml
```diff
@@ -0,0 +1,198 @@
...
+
+      <DataTemplate x:Key="ContentTemplateButtonStack">
+        <!-- Margin of 10 here is to "make room" for the elevation drop shadows of the buttons. Must be compensated for in some (left/right) alignment scenarios using PopupBox.PopupHorizontalOffset -->
+        <StackPanel Margin="10">
+          <Button Content="1"
+                  Opacity="0.5"
+                  ToolTip="One with custom opacity" />
+          <Button Content="2" ToolTip="Two" />
+          <Button Content="3" ToolTip="Three" />
+        </StackPanel>
```
nicolaihenriksen, line 45: The `Margin` applied to this `StackPanel` is there to ensure there is enough room for the drop shadows of the buttons in the `StackPanel`.

For the default style of the `PopupBox`, this is likely not an issue as the `Popup` contains a `Card` which contains the provided content. However, for the `MaterialDesignMultiFloatingActionPopupBox` derived styles, there isn't a concept of a container like the `Card` and it is simply the provided content which is shown. So in cases like these, if the user puts in stuff with drop shadows that would normally extend the bounds of the container, he/she would need to add a `Margin` to the root container (like done here) which is large enough to encompass the child elements AND their drop shadows. This will naturally affect the positioning of the `Popup`, and in those cases the user would need to compensate for that by setting "an equal and opposite" offset on the `PopupBox` using `PopupBox.PopupHorizontalOffset` and `PopupBox.PopupVerticalOffset`.

nicolaihenriksen:
@Keboo I have left a few comments above. 2 of them surprised me a bit and may be something that we need to look into.

The last one is simply an elaboration on a mitigation of a drop shadow clipping issue to explain why it is like it is. This scenario was also present before my changes, but I struggle to see how we can "fix" it such that the user does not need to worry about it. One option is to always "reserve space" using the largest possible drop-shadow radius, but in case the users content does not have drop shadows at all, we will be covering a lot of UI real-estate with a (transparent background) `Popup` which is hit test visible and thus will "block" interaction with elements in the parent window which are located "nearby". So I chose not to do anything about it.

Have a look at it, and give the branch a spin when you have some time. Looking forward to hear what your thoughts are about it.

MainDemo.Wpf/PopupBox.xaml
Keboo: Yea with the `MaterialDesignMultiFloatingActionPopupBox` this is expected. It is sort of a hard case because we apply the same Foreground/Background to both the button and the popup.

Keboo (commented):
One thing that seems odd, is the Elevation property is applied inconsistently to either the toggle button or the popup content depending on the style selected. 

A couple options.
1. We have two properties, one for the toggle button, and one for the popup.
2. We adjust the samples so that they are at least consistent.

nicolaihenriksen:
> One thing that seems odd, is the Elevation property is applied inconsistently to either the toggle button or the popup content depending on the style selected.

I agree that this is a bit odd. However, it was like that even before my changes. The `ElevationAssist.Elevation` **only** affected the `ToggleButton` in the `MaterialDesignMultiFloatingActionPopupBox` derived styles, and had no effect in the default style. With my changes, it now **also** has an effect in the default style, albeit a "different" effect.

We can of course "fix" the samples so they are at least consistent. However, I don't think that fixes the oddity that the two styles behave differently. I mean both styles contain a "clickable" element which opens a `Popup`, but only one of the clickable elements is affected by the `ElevationAssist.Elevation`.

In short, I am not really sure what the correct approach is. `ElevationAssist.Elevation` is a "general purpose" attached property which perhaps should only affect the "clickable" element and not the `Popup`? But currently it does nothing in the default style. And if we go this route, we probably need a dedicated dependency property on `PopupBox` which can control the elevation of the `Card` in the default style; this is also a bit weird since the DP only applies to certain styles of the control. ALternatively we introduce a `CardAssist.Elevation` (inherits) AP which can be applied to the `PopupBox` (and then inherited by the `Card` in the `Popup`; if that even works, since it is not really a child in the visual tree, but an independent "window").

Thoughts?

MainDemo.Wpf/PopupBox.xaml
nicolaihenriksen: Yeah perhaps it should then. It feels a little weird to have to include that explicitly.

MainDemo.Wpf/PopupBox.xaml
nicolaihenriksen: I did consider (and try) `ClipToBounds=False`, but that does not seem to work for a `Popup` for something like this. It is like there is a hard cut-off at the edge of the window which represents the `Popup`. This is the reason I decided to go with a solution that "reserves space" for the drop shadow.

Keboo:
> I agree that this is a bit odd. However, it was like that even before my changes. The ElevationAssist.Elevation only affected the ToggleButton in the MaterialDesignMultiFloatingActionPopupBox derived styles, and had no effect in the default style. With my changes, it now also has an effect in the default style, albeit a "different" effect.

That is a good point. Since it is not a regression we can ignore it for now.

> We can of course "fix" the samples so they are at least consistent. However, I don't think that fixes the oddity that the two styles behave differently. I mean both styles contain a "clickable" element which opens a Popup, but only one of the clickable elements is affected by the ElevationAssist.Elevation.

Given that it is not a regression this is probably not worth it right now. I took a look at the open issues that relate to the popup box and it doesn't appear that this is an issue that is really affecting people.

> In short, I am not really sure what the correct approach is. ElevationAssist.Elevation is a "general purpose" attached property which perhaps should only affect the "clickable" element and not the Popup? But currently it does nothing in the default style. And if we go this route, we probably need a dedicated dependency property on PopupBox which can control the elevation of the Card in the default style; this is also a bit weird since the DP only applies to certain styles of the control. Alternatively we could introduce a CardAssist.Elevation (inherits) AP which can be applied to the PopupBox (and then inherited by the Card in the Popup; if that even works, since it is not really a child in the visual tree, but an independent "window").

I am wondering if perhaps splitting the control would make more sense. I do agree that a dedicated DP would make the most sense, but that is awkward if we only have a single control. Splitting the control would then allow for adding the DP only to the one where it makes sense. If we decide to go this route it might be nice to do it with the 5.0.0 release since it would be a breaking change for people (albeit a simple one to address).

Though again, the more I think on it, I think this PR is probably good enough that we can merge it and consider future improvements down the road.

nicolaihenriksen:
> Though again, the more I think on it, I think this PR is probably good enough that we can merge it and consider future improvements down the road.

I agree it probably makes sense to split into 2 different controls going forward.

I just took another quick look at the `Popup` template. We already have a dedicated DP (`PopupUniformCornerRadius`) which is only used in the card-based `Popup`. So perhaps we should just stick with that asymmetric approach and create a dedicated `PopupElevation` DP which is also only used in the card-based style?

The `ElevationAssist.Elevation` is after all not created as an `Inherits` attached property, so it is a bit misleading that you need to use that to affect a "child" element; the fact that it currently does not affect the button that opens the `Popup` in the default style is a different issue that could be handled separately.

If you agree with the `Popup.PopupElevation` DP approach, I can make those changes, rebase and clean up the branch to make it ready for review?

What are your thoughts on the dedicated PopupBox demo page? It is different from all the others, and more along the lines of what you've talked about in your streams; being able to manipulate the properties of the "displayed" control. We could leave it out, but then use it for inspiration for a future more generic approach; the current one is really tied to the `Popup` control.

Keboo:
> What are your thoughts on the dedicated PopupBox demo page? It is different from all the others, and more along the lines of what you've talked about in your streams; being able to manipulate the properties of the "displayed" control. We could leave it out, but then use it for inspiration for a future more generic approach; the current one is really tied to the Popup control.

Yea I sort of like it. I say we keep it.

Go ahead with the `Popup.PopupElevation` DP approach. I am good with adding it.

MainDemo.Wpf/PopupBox.xaml
Keboo: Upon further inspection I think it is working as expected.
All controls have their default style loaded by the App.xaml only referencing the Defaults resource dictionary, with users needing to reference the other resource dictionaries, so for now I think the behavior is correct.

Keboo (approved):

PR #3153, opened 2023-03-30
Update PaletteHelper.cs

JLdgu:
Correct variable name


JLdgu:
Also https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/colorThoughts/mdresgen/Brushes.cs
says msregen in header instead of mdresgen

Keboo (approved):
Thanks!

PR #3162, opened 2023-04-08
TextBox: Removed not needed VerticalAlignment and it now respects VerticalContentAlignment

### LINKED ISSUE #3161: TextBox with Height puts HintAssist.Hint right in the middle instead of the top.
### Bug explanation

We updated `MaterialDesignThemes` from `4.1` to `4.8` and noticed all the `TextBox` we had with a configured height will result in the `materialDesign:HintAssist.Hint` being put right in the middle of the TextBox instead of at the top (like it was before).  For a test, I pulled latest on the MaterinalDesign repo, adjusted the height in the sample app to confirm it happens there also.

Example:
image
image

In our app with v4.8:
image

With the previous nuget package  v4.1: 
image

I am not sure if it's related to 4.8 directly - not sure how many changes between 4.1 and 4.8, but I suspect this is not how you want it rendering even for larger size TextBoxes.  

If there is a style override, I can put that in, but wanted to report it as the Demo even shows this issue. Let me know if this is the desired behavior though so I can adjust accordingly. Thanks!

### Version

4.8.0

nicolaihenriksen:
Fixes #3161 

Apparently I had missed this when doing the refactoring, and I must admit I cannot recall exactly why that `VerticalAlignment` was introduced in the `TextBox` style. My guess is one of two things happened:
* Either it is a left over change I forgot to revert when trying various approaches to get the `SmartHint` to align correctly, or
* An attempt to align across the `PasswordBox` and `TextBox` styles; because `The PasswordBox` had this alignment already in its 2 styles.

~~Anyways, this PR removes it as it does not seem needed. I think there may still be an issue though (should probably be created as a separate issue): I would assume that the hint as well as the "text box content" would respect the `TextBox.VerticalContentAlignment` which I don't think it currently does.~~

~~Another thing is, as mentioned in #3161, we should put some UI tests in place to catch regressions in these horizontal/vertical alignment scenarios.~~

I managed to find a little free time after all. I updated the PR to consider `TextBox.VerticalContentAlignment` as well, and added UI tests for all `VerticalAlignment` enum values. For the "non-top-aligned" scenarios, the user will need to manually adjust the `HintAssist.FloatingOffset` to get correct floating placement of the hint (I have done this for the GIFs below).

@Keboo The UI test is probably not the best test in the world, as it simply asserts the `VerticalAlignment` value is correctly propagated down to the hint. I guess ideally it should also assert that the absolute position of the hint is "correct" (i.e. top/center/bottom of the `TextBox`).

`VerticalAlignment.Top` or `VerticalAlignment.Stretch`
AlignmentTopOrStretch

`VerticalAlignment.Center`
AlignmentMiddle

`VerticalAlignment.Bottom`
AlignmentBottom

### DIFF
--- a/MaterialDesignThemes.UITests/WPF/TextBoxes/TextBoxTests.cs
+++ b/MaterialDesignThemes.UITests/WPF/TextBoxes/TextBoxTests.cs
@@ -562,6 +562,34 @@ public class TextBoxTests : TestBase
 
         recorder.Success();
     }
+
+    [Theory]
+    [InlineData(VerticalAlignment.Stretch, VerticalAlignment.Top)]
+    [InlineData(VerticalAlignment.Top, VerticalAlignment.Top)]
+    [InlineData(VerticalAlignment.Bottom, VerticalAlignment.Bottom)]
+    [InlineData(VerticalAlignment.Center, VerticalAlignment.Center)]
+    [Description("Issue 3161")]
+    public async Task TextBox_MultiLineAndFixedHeight_RespectsVerticalContentAlignment(VerticalAlignment alignment, VerticalAlignment expectedFloatingHintAlignment)
+    {
+        await using var recorder = new TestRecorder(App);
+
+        var stackPanel = await LoadXaml<StackPanel>($$"""
+            <StackPanel>
+              <TextBox Style="{StaticResource MaterialDesignFilledTextBox}"
+                materialDesign:HintAssist.Hint="Hint text"
+                VerticalContentAlignment="{{alignment}}"
+                AcceptsReturn="True"
+                Height="200" />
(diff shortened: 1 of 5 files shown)


TWhidden:
I had not considered the Password Style as having multi-line would be odd, but for sure can confirm that is the change I made on my local source for the TextBox style and did resolve it.  I suppose adding some multi-line textboxes to the UI test app would be good. For the meantime however, this PR looks good to me.

nicolaihenriksen:
I agree, PasswordBox style does not need to consider multiline, but nor was the VerticalAlignment needed; so I removed it for consistency.

Keboo (approved):

Issue #2592, opened 2022-03-03
Outlined NumericUpDown not displayed correctly when disabled


Int32Overflow:
If the NumericUpDown with the style "MaterialDesignOutlinedNumericUpDown" was disabled (IsEnabled=false), then a wrong style is displayed.

**Versions:**
- MahApps.Metro: 2.4.9
- MaterialDesignThemes: 4.3.0
- MaterialDesignThemes.MahApps: 0.2.0

**Used Style:** `MaterialDesignOutlinedNumericUpDown`

**IsEnabled=true**
![grafik](https://user-images.githubusercontent.com/20139038/156523117-5b811faa-03f6-461e-a388-5545492412f6.png)

**IsEnabled=false**
![grafik](https://user-images.githubusercontent.com/20139038/156523181-aa224c85-48f9-4027-93cb-fbe4aa5be1ee.png)

MichelMichels:
It seems to me that the NumericUpDown control is not supported by this repository. I don't know if this is still under active development.

MichelMichels:
Closing due to staleness and out of scope.

Issue #2557, opened 2022-01-10
NumericUpDown: Outline has a line at the bottom


Int32Overflow:
**Version:** 4.3.0

**Description:** The NumericUpDown with the outlined style has a thicker line on the bottom when it gets the focus.

![NumericUpDown-LineBottom](https://user-images.githubusercontent.com/20139038/148739363-56be271b-9513-44b2-8c7b-eda16f36e5b4.gif)

**Expected:** Should look like the "Textbox" in outlined style.

Keboo:
@Int32Overflow can you provide which version of MahApps you are using?

Int32Overflow:
@Keboo I'm using: 

- MahApps.Metro 2.4.9
- MaterialDesignThemes.MahApps 0.2.0
- MaterialDesignThemes 4.3.0

Keboo:
Thanks

Issue #2448, opened 2021-09-30
NumericUpDown Outlined


Int32Overflow:
For MahApp's NumericUpDown there is no style "Outline" as it is available for other text fields

PR #3175, opened 2023-04-25
Pull Request: Adding New Control NumericUpDown Control to Material Design XAML Toolkit

Asivaprasadam:
# Pull Request: Adding NumericUpDown control to Material Design
## Preface

Hello everyone,

This is my first contribution to a GitHub project, and I am excited to be a part of the MaterialDesignXamlToolkit community. I have been working on this NumericUpDown control for Material Design, and I believe it will be a valuable addition to the project.

I must admit that my lack of experience contributing to open-source projects has delayed me in committing to this code. However, I have decided that I don't want to delay it any further by waiting to complete all the planned improvements before submitting my contribution when there is a strong community that can help me out with my 1st contribution.

Since this is my first GitHub contribution, I would appreciate any feedback and advice you can provide. I am eager to learn from more experienced developers and improve my skills, so please do not hesitate to give me constructive criticism and suggestions for improvements. I am committed to contributing to this project and look forward to continuing to learn and improve.

Thank you for taking the time to review my contribution, and I look forward to your feedback.

## Description

This pull request adds a new NumericUpDown control to the Material Design project. The control provides a user-friendly way to input numeric values and is designed to fit seamlessly into the Material Design UI.

## Changes Made

 - [x] Created a new NumericUpDown control that allows users to input numeric values within a specific range.
 - [x] Implemented the first base style for the control, nothing fancy and definitely needs improvement.
 - [x] Integrated the NumericUpDown with the Material Design MainDemo and added a page to display styles.

##  Pending Potential Changes

 - [ ] Provided additional documentation to help users understand the control's functionality and usage.
 - [ ] Improve the support for common input methods, such as keyboard and mouse input.
 - [ ] Create logarithmic click/ scroll support to enhance the UX/ user experience.
 - [ ] Tune the base style to adhere more closely to Material Design principles.
 - [ ] Create additional styling templates following Material Design guidelines to provide more options for users
 - [ ] Add all the new styles to the Main Demo App NumericUpDown control page to showcase the control.

## Screenshots

image

## Related Issues

 1. Issue #2592
 2. Issue #2557
 3. Issue #2448 

## Additional Notes
None


MaterialDesignThemes.Wpf/NumericUpDown.cs
```diff
@@ -0,0 +1,213 @@
...
+        {
+            OnDecrease();
+            e.Handled = true;
+        }
+        base.OnPreviewKeyDown(e);
+    }
+
+    protected override void OnPreviewMouseWheel(MouseWheelEventArgs e)
+    {
+        if (IsKeyboardFocusWithin)
```
Keboo, line 198: It might be nice to create another dependency property to disable this behavior for those who don't want it.

MaterialDesignThemes.Wpf/NumericUpDown.cs
```diff
@@ -0,0 +1,213 @@
...
+        if (IncreaseCommand?.CanExecute(null) ?? false)
+            IncreaseCommand.Execute(null);
+    }
+
+    private void OnDecrease()
+    {
+        SetCurrentValue(ValueProperty, Value - 1);
+
+        if (DecreaseCommand?.CanExecute(this) ?? false)
+            DecreaseCommand.Execute(this);
```
Keboo, line 178: For commands, we should create new dependency properties for the CommandParameters. This should be similar to how the `Button` works.

MainDemo.Wpf/NumericUpDown.xaml
```diff
@@ -0,0 +1,37 @@
...
+               Style="{StaticResource MaterialDesignHeadline5TextBlock}"
+               Text="NumericUpDown Control" />
+
+    <StackPanel Margin="0,16,0,0"
+                DockPanel.Dock="Top"
+                Orientation="Horizontal">
+      <smtx:XamlDisplay Margin="10,5"
+                        VerticalAlignment="Top"
+                        UniqueKey="numericUpDown_default">
+        <materialDesign:NumericUpDown />
```
Keboo, line 31: Would be nice to put in a few more examples showing the commands, Min and Max, and any other styles.

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.NumericUpDown.xaml
```diff
@@ -0,0 +1,58 @@
...
+      as part of Material Design in XAML Toolkit. Other themes, for existing controls
+      must be selected manually by the user, so that they override default styles.
+      The easiest way to do this is include:
+      
+      MaterialDesignTheme.Defaults.xaml
+      in your App.xaml
+    -->
+  </ResourceDictionary.MergedDictionaries>
+  <!-- set up default styles for our custom Material Design in XAML Toolkit controls -->
+  <Style TargetType="{x:Type wpf:NumericUpDown}">
```
Keboo, line 21: One idea for a cleaner look could be to match the look of the [one from MatBlazor](https://www.matblazor.com/NumericUpDownField)

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.NumericUpDown.xaml
```diff
@@ -0,0 +1,58 @@
...
+              <RepeatButton x:Name="PART_DecreaseButton"
+                            Content="{local:PackIcon Kind=Minus}"
+                            DockPanel.Dock="Left"
+                            Padding="8,4"
+                            Command="{TemplateBinding DecreaseCommand}"/>
+              
+              <TextBox x:Name="PART_TextBoxField"
+                       Margin="4 0"
+                       VerticalAlignment="Center"
+                       HorizontalContentAlignment="Center" />
```
Keboo, line 49: We probably want to consider forwarding the HorizontalContentAlignment from the parent here, and default the value to centered in the Style setters.
```suggestion
                       HorizontalContentAlignment="{TemplateBinding HorizontalContentAlignment}" />
```

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.NumericUpDown.xaml
```diff
@@ -0,0 +1,58 @@
...
+                            DockPanel.Dock="Right"
+                            Padding="8,4"
+                            Command="{TemplateBinding IncreaseCommand}"/>
+              <RepeatButton x:Name="PART_DecreaseButton"
+                            Content="{local:PackIcon Kind=Minus}"
+                            DockPanel.Dock="Left"
+                            Padding="8,4"
+                            Command="{TemplateBinding DecreaseCommand}"/>
+              
+              <TextBox x:Name="PART_TextBoxField"
```
Keboo, line 46: We probably also want to forward the TextBox hint properties. These are attached properties that should be forwarded into the TextBox so that consumers can change them.
https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MainDemo.Wpf/Fields.xaml#L107-L118

Keboo (commented):
One other thing that would be good to do is write some UI tests for this control. If you need any help setting that up, let me know.

Xaalek:
Plus and Minus seem inverted according to your screenshot. Here are some examples of what we usually see in UI:
image
image

tim-gromeyer:
Any update on this?  @Asivaprasadam?  Would be cool to have this in my project, this is exactly what I need!

arjun-sivaprasadam:
@tim-gromeyer I have changed my work abit and switched from UI development to more of a backend development. But I'll check with the progress and do an update possibly asap, but lets say optimally by end of Jan worst case by march.

MaterialDesignThemes.UITests/TestBase.cs
```diff
@@ -16,7 +17,7 @@ namespace MaterialDesignThemes.UITests;
 
 public abstract class TestBase(ITestOutputHelper output) : IAsyncLifetime
 {
-    protected bool AttachedDebuggerToRemoteProcess { get; set; } = true;
+    protected bool AttachedDebuggerToRemoteProcess { get; set; } = false;
```
nicolaihenriksen, line 20: @Keboo please revert this before merging

MaterialDesignThemes.Wpf/NumericUpDown.cs
```diff
@@ -0,0 +1,257 @@
...
+    public object? DecreaseCommandParameter
+    {
+        get => (object?)GetValue(DecreaseCommandParameterProperty);
+        set => SetValue(DecreaseCommandParameterProperty, value);
+    }
+
+    public static readonly DependencyProperty DecreaseCommandParameterProperty =
+        DependencyProperty.Register(nameof(DecreaseCommandParameter), typeof(object), typeof(NumericUpDown), new PropertyMetadata(null));
+
+    #endregion DependencyProperty : DecreaseCommandProperty
```
nicolaihenriksen, line 133: Let me know if I am completely off the rails here, but here are my thoughts.

I think these commands are simply left-overs from when the XAML was using them. And I agree that is not common practice for a custom control; you should rely on the template parts and deal with this stuff in the code-behind.

Also, in their current state, I really don't think they should be part of the public API for a `NumericUpDown`. If you look at a [Slider](https://source.dot.net/#PresentationFramework/System/Windows/Controls/Slider.cs,96) for example, it does expose some static `RoutedCommands` to modify the value, but they don't have public setters. This way the calling code is not involved in determining whether `Command.CanExecute` returns true/false which would be a really weird thing; it is the control itself that has the knowledge to determine this, not the caller. The commands are still exposed so they can be bound to and thus executed in custom UI/flows.

MaterialDesignThemes.Wpf/NumericUpDown.cs
```diff
@@ -0,0 +1,257 @@
...
+    private void IncreaseButtonOnClick(object sender, RoutedEventArgs e) => OnIncrease();
+
+    private void DecreaseButtonOnClick(object sender, RoutedEventArgs e) => OnDecrease();
+
+    private void OnIncrease()
+    {
+        SetCurrentValue(ValueProperty, Value + 1);
+
+        if (IncreaseCommand?.CanExecute(IncreaseCommandParameter) ?? false)
+            IncreaseCommand.Execute(IncreaseCommandParameter);
```
nicolaihenriksen, line 214: Related to my comment above, this feels weird to me. The command should be used to actually change the value, not be executed once the value has changed. For that, there is a `ValueChanged` `RoutedEvent`.

MaterialDesignThemes.Wpf/NumericUpDown.cs
```diff
@@ -0,0 +1,257 @@
...
+            if (e.Delta > 0)
+            {
+                OnIncrease();
+                e.Handled = true;
+            }
+            else if (e.Delta < 0)
+            {
+                OnDecrease();
+                e.Handled = true;
+            }
```
nicolaihenriksen, line 253: Not sure if it is an actual issue, but this will not mark the event as handled if `e.Delta == 0` which seems a bit strange.

Imagine if the control is located in a `ScrollViewer`, it has keyboard focus you the user user the scroll wheel to change the values. Would there be a scenario where you think you are scrolling the values, but hit the edge case and suddenly the `ScrollViewer` scrolls instead because the event bubbles up? That would be annoying!

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.NumericUpDown.xaml
nicolaihenriksen: I think, similar to some of the other controls that "host" a `TextBox`, we probably want to strip that particular `TextBox` of all the fancy MDIX stuff and implement it directly in the template for the `NumericUpDown`.

In other words, I assume this should eventually include all the prefix-/suffix-texts, leading-/trailing-icons, `SmartHint`, etc. If you look at some of the other controls (`ComboBox` for example), you will see that those things are handled in the "host" template, and not leveraging a "styled" `TextBox`.

To elaborate on the above. Today we cannot easily leverage a "styled" `Textbox` with the way our triggers are working today I believe, and therefore we end up having to duplicate a lot of the stuff (you'll see that in the other control styles). It may be a future refactoring task to see if we can somehow leverage a "styled" `TextBox` in the controls that nests one of those rather than having to handle all the fancy stuff in the "host" template.

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.NumericUpDown.xaml
```diff
@@ -0,0 +1,56 @@
...
+                <ColumnDefinition Width="Auto" />
+              </Grid.ColumnDefinitions>
+
+              <TextBox x:Name="PART_TextBoxField"
+                       Margin="4 0"
+                       VerticalAlignment="{TemplateBinding VerticalContentAlignment}"
+                       HorizontalContentAlignment="{TemplateBinding HorizontalContentAlignment}" />
+
+              <RepeatButton x:Name="PART_DecreaseButton"
+                            Content="{local:PackIcon Kind=Minus}"
```
nicolaihenriksen, line 39: I think the `PackIcons` used for the plus/minus buttons should be exposed as dependency properties defaulting to the values you have set here. That gives a caller better control of what to show in the button.

Possibly it shouldn't even be a `PackIcon` but simply an `object` in case we want callers to be able to add arbitrary content to these buttons.

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.NumericUpDown.xaml
```diff
@@ -0,0 +1,56 @@
...
+                            Style="{StaticResource MaterialDesignFlatButton}"
```
nicolaihenriksen, line 40: If/when we introduce more styles for this control (e.g. "Floating", "Filled", and "Outlined"), the `MaterialDesignFlatButton` style hardcoded here should probably be a parameter passed into the base style (internal DP/AP?).

Alternatively a dedicated `RepeatButtonStyle` can be passed in from the style-override.

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.NumericUpDown.xaml
```diff
@@ -0,0 +1,56 @@
...
+                            Padding="8,4"
+                            Grid.Column="2" />
+            </Grid>
+          </Border>
+        </ControlTemplate>
+      </Setter.Value>
+    </Setter>
+  </Style>
+
+  <Style TargetType="wpf:NumericUpDown" BasedOn="{StaticResource MaterialDesignNumericUpDown}" />
```
nicolaihenriksen, line 55: These default implicit styles for the custom controls are typically set up in `Generic.xaml`; see [here](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/7f27b729df9ed950e339a17d7aa0bce1bc46f3ce/MaterialDesignThemes.Wpf/Themes/Generic.xaml#L37-L44).

nicolaihenriksen (commented):
Left a few comments to consider.

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.NumericUpDown.xaml
Keboo: Perfect, I will see about getting that change done.

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.NumericUpDown.xaml
Keboo: That is a good point, just expose the button's content and let consumers do what they like with it.

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.NumericUpDown.xaml
Keboo: That is a good point, going to punt on those for now and see if they get requested.

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.NumericUpDown.xaml
Keboo: Deciding to punt on doing this for now and not having a hint on this control to start.

PR #3181, opened 2023-05-01
Additional SmartHint refactoring

### LINKED ISSUE #3174: Filled Text Box hint position
### Bug explanation

The hint of the filled text box doesn't seem to be in the right position:

image
image

For reference, here is an outlined text box with the same code, which has the right alignment:

image

My code:
```xaml
<TextBox Style="{StaticResource MaterialDesignFilledTextBox}"
                             materialDesign:HintAssist.Hint="Path"
                             materialDesign:HintAssist.IsFloating="True"
                             materialDesign:TextFieldAssist.HasLeadingIcon="True"
                             materialDesign:TextFieldAssist.LeadingIcon="DatabaseOutline"
                             Margin="15 0 0 0"
                             FontSize="15"
                             Padding="12 8"
                             Height="56"
                             Foreground="{DynamicResource OnBackground}" />
```

I tried using TextBoxViewMargin but without much success.

### Version

4.8.0

nicolaihenriksen:
Fixes #3174

A number of issues (including the one above) have been reported with the alignment of the `SmartHint`, and there are also several different approaches to "solving the same problem" for the different controls.

This PR aims to align this across the control styles for:
* `TextBox`
* `RichTextBox`
* `PasswordBox`
* `ComboBox`
* `DatePicker`
* `TimePicker`

There are a number of variables to consider when figuring out the hint placement, and I don't believe any of the current controls handle all of them. As mentioned, this PR aims to align the approach across the above templates taking the following into account for the placement of the hint (and leading/trailing icons, buttons, prefix texts, and suffix texts):

* `Control.Padding`
* `Control.VerticalContentAlignment`
* `Control.HorizontalContentAlignment`
* `Control.Height`
* `Control.FontSize` - 🔴 Currently only partially supported; similar to before this PR
* `HintAssist.FloatingScale`
* `HintAssist.FloatingOffset`
* `HintAssist.FloatingHintHorizontalAlignment`
* `HintAssist.FontFamily`
* Styles ("Floating", "Filled", "Outlined")

The `Smart Hint` demo app page is extended to include manual testing options for all of this. Please check out the branch and take "this for a spin" playing around with the various options. As noted below, you can configure some combinations which are a bit strange, but I believe you could do that before this PR as well. Additional clean up could be done.

image

In the current state of the PR, the `FontSize` is only supported when no explicit `Height` is applied, and even then it is only partially supported if a non-default `FontSize` is applied. It was like that before this PR as well, although it there may be some differences in certain scenarios; it is a bit hard to say. I am still working on my local branch trying to make this work for all font sizes, but that might end up being a different PR in the end as I think it may require some significant changes to the many converters that are in play for adjusting `Padding`/`Margin` in the `Top`/`Center`/`Bottom`/`Stretch` cases of `VerticalContentAlignment`.

The `TextBox` styles support all of the properties that can be modified in these panels, but the other control styles only support a sub-set of them. Each control style seemingly supporting a different sub-set 🤔. This PR does not seek to align that. If people need it, I hope/assume they will create an issue in the repo.

It is quite possible for consumers of the controls to configure a combination of the features which end up not looking as nice as you would hope. I believe these are separate issues though, and could be handled as such.

As we've discussed, I believe it might be a good idea to split some of the templates apart going forward. Having a single control template trying to juggle 3 different styles (via attached properties) and placing/moving content accordingly makes the templates quite complex. The changes - in particular added use of converters - in this PR does not make it any less complex!

### DIFF
--- a/MaterialDesignThemes.Wpf/Converters/PickerInnerPaddingConverter.cs
+++ b/MaterialDesignThemes.Wpf/Converters/PickerInnerPaddingConverter.cs
@@ -3,6 +3,7 @@ using System.Windows.Data;
 
 namespace MaterialDesignThemes.Wpf.Converters
 {
+    [Obsolete("No longer used internally by MDIX, but since it is public I have chosen to keep it here. Will be removed in the 5.0 release")]
     public class PickerInnerPaddingConverter : IValueConverter, IMultiValueConverter
     {
         private readonly FloatingHintOffsetCalculationConverter _offsetCalculation = new FloatingHintOffsetCalculationConverter();
--- a/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.TextBox.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.TextBox.xaml
@@ -32,6 +32,7 @@
   <converters:ThicknessCloneConverter x:Key="OutlinedStyleTrailingIconMarginConverterCenter" CloneEdges="Top" AdditionalOffsetTop="0" />
   <converters:ThicknessCloneConverter x:Key="OutlinedStyleTrailingIconMarginConverterBottom" CloneEdges="Top" FixedBottom="-2" AdditionalOffsetTop="0" />
   <wpf:VerticalAlignmentConverter x:Key="VerticalAlignmentConverter" />
+  <converters:NullableToVisibilityConverter x:Key="NullableToVisibilityConverter" />
 
   <Style x:Key="MaterialDesignCharacterCounterTextBlock"
          TargetType="TextBlock"
@@ -205,7 +206,8 @@
                                  VerticalAlignment="{TemplateBinding VerticalContentAlignment}"
                                  FontSize="{TemplateBinding FontSize}"
                                  Opacity="{TemplateBinding wpf:HintAssist.HintOpacity}"
-                                 Text="{TemplateBinding wpf:TextFieldAssist.SuffixText}" />
+                                 Text="{TemplateBinding wpf:TextFieldAssist.SuffixText}"
+                                 Visibility="{TemplateBinding wpf:TextFieldAssist.SuffixText, Converter={StaticResource NullableToVisibilityConverter}}" />
                     </Grid>
 
                     <wpf:SmartHint x:Name="Hint"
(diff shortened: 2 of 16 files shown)


MaterialDesignThemes.Wpf/Converters/PickerInnerPaddingConverter.cs
```diff
@@ -3,6 +3,7 @@
 
 namespace MaterialDesignThemes.Wpf.Converters
 {
+    [Obsolete("No longer used internally by MDIX, but since it is public I have chosen to keep it here. Will be removed in the 5.0 release")]
```
Keboo, line 6: I added removing this to the list of breaking 5.0.0 changes

Keboo (approved):
This is awesome work! thank you so much!

MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.TextBox.xaml
```diff
@@ -19,6 +19,20 @@
   <converters:CursorConverter x:Key="IBeamCursorConverter" FallbackCursor="IBeam" />
   <converters:StringLengthValueConverter x:Key="StringLengthValueConverter" />
   <converters:ThicknessCloneConverter x:Key="HelperTextMarginConverter" CloneEdges="Left" />
+  <converters:ThicknessCloneConverter x:Key="DefaultOrFilledStyleLeadingIconMarginConverterTop" CloneEdges="Top" FixedRight="6" AdditionalOffsetTop="-2" />
+  <converters:ThicknessCloneConverter x:Key="DefaultOrFilledStyleLeadingIconMarginConverterCenter" CloneEdges="Top" FixedRight="6" AdditionalOffsetTop="-12" />
+  <converters:ThicknessCloneConverter x:Key="DefaultOrFilledStyleLeadingIconMarginConverterBottom" CloneEdges="Top" FixedRight="6" FixedBottom="-2" />
+  <converters:ThicknessCloneConverter x:Key="DefaultOrFilledStyleTrailingIconMarginConverterTop" CloneEdges="Top" AdditionalOffsetTop="-2" />
+  <converters:ThicknessCloneConverter x:Key="DefaultOrFilledStyleTrailingIconMarginConverterCenter" CloneEdges="Top" AdditionalOffsetTop="-12" />
```
nicolaihenriksen, line 26: @Keboo 
It's the constants used here (and other similar places) in the FixedXxx and AdditionalOffsetXxx properties that are problematic when a non-default `FontSize` is applied.

So the correct approach is probably to calculate the values based on the `FontSize` (if we can manage to come up with the correct equation for it). That would however also require that the converter becomes an `IMultiValueConverter` which will make the templates even more complex. So I am trying to think of other ways to solve it, but I have yet to come up with a better solution...

PR #3222, opened 2023-05-26
Add IsSynchronizedWithCurrentItem to MaterialDesignDataGridComboBox

### LINKED ISSUE #1197: MaterialDataGridComboBoxColumn Not Working With ListCollectionView
Very Simple Issue:

When using a MaterialDataGridComboBoxColumn and binding to a ListCollectionView instead of a standard list<T> or an observablecollection<T> all of the instances of the MaterialDataGridComboBoxColumn in that datagrid are not displaying the correct value and instead are all showing the same value. If you select a new value in that combobox for one row then all other rows are duplicating that value. I have attached the most barebones sample I could make. Its just one window, with a datagrid and a single MaterialDataGridComboBoxColumn. The MaterialDataGridComboBoxColumn should have the values "1", "2", "3" in the example three rows generated but instead shows just "3" for all three rows.

[MaterialDesignTest.zip](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/files/2902537/MaterialDesignTest.zip)

nicolaihenriksen:
On behalf of @rzgthb I am adding this PR which fixes #1197. 

See comment below for details.

### DIFF
--- a/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.DataGrid.ComboBox.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.DataGrid.ComboBox.xaml
@@ -432,6 +432,7 @@
     <Setter Property="ScrollViewer.VerticalScrollBarVisibility" Value="Auto" />
     <Setter Property="Stylus.IsFlicksEnabled" Value="False" />
     <Setter Property="Template" Value="{StaticResource MaterialDesignDataGridComboBoxTemplate}" />
+    <Setter Property="IsSynchronizedWithCurrentItem" Value="False"/>
     <Style.Triggers>
       <Trigger Property="IsEditable" Value="true">
         <Setter Property="IsTabStop" Value="false" />


rzgthb:
### Issue

The default behavior for controls which are derived from Selector and use a CollectionView is to synchronize their selected item to the current item of the CollectionView: [IsSynchronizedWithCurrentItem](https://learn.microsoft.com/en-us/dotnet/api/system.windows.controls.primitives.selector.issynchronizedwithcurrentitem?view=windowsdesktop-7.0#system-windows-controls-primitives-selector-issynchronizedwithcurrentitem). Since DataGridComboBoxColumn is typically bound to a CollectionViewSource, this behavior would affect all DataGridComboBoxColumn bound to the same CollectionView, resulting in all contained ComboBox showing the same item on changes to any of them. The default implementation of DataGridComboBoxColumn therefore sets this property to false for the DefaultElementStyle of the contained combo box, disabling this behavior. The DefaultEditingElementStyle returns DefaultElementStyle.


MDIX' DefaultElementStyle is based on MaterialDataGridComboBoxColumnStyle, its DefaultEditingElementStyle is based on MaterialDataGridComboBoxColumnEditingStyle, which both are in turn based on the MDIX style MaterialDesignDataGridComboBox, which is not based on any other styles. MaterialDesignDataGridComboBox does not set the IsSynchronizedWithCurrentItem property to false, therefore enabling item synchronization when bound to a CollectionView.

### Purpose of this PR

This request aims to introduce code changes setting IsSynchronizedWithCurrentItem to false for MaterialDesignDataGridComboBox, therefore disabling item synchronization.

### Changes

Add `<Setter Property="IsSynchronizedWithCurrentItem" Value="False"/>` to MaterialDesignTheme.DataGrid.ComboBox.xaml.

Keboo (approved):

PR #3279, opened 2023-07-19
Ensure clear button does not overlap editable text in ComboBox

### LINKED ISSUE #3276: MaterialDesignFloatingHintComboBox style and TextFieldAssist.HasClearButton="True"
### Bug explanation

Setting the TextFieldAssist.HasClearButton="True" attached property on the MaterialDesignFloatingHintComboBox style causes the "clear text" button to overlap the text inside the combo box.

I checked the other combobox styles, and it appears that they all have the same issue, except for MaterialDesignDataGridComboBox which (at first glance anyway) doesn't look like it supports the TextFieldAssist.HasClearButton attached property.

### Version

4.9

nicolaihenriksen:
Fixes #3276 

This PR ensures the `PART_ClearButton` does not overlap the editable text in `PART_EditableTextBox` of the `ComboBox` template.

The reason for the somewhat weird initial offset of -8 is because of some internal margin/padding between the "selected content (text)" and the "chevron arrow" which is already in place and thus should not be included in the margin added to the right side of the `PART_EditableTextBox`. If you have a better approach of dynamically getting this value, feel free to update the code 😛 

image

### DIFF
--- a/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.ComboBox.xaml
+++ b/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.ComboBox.xaml
@@ -26,6 +26,7 @@
   <converters:FloatingHintOffsetCalculationConverter x:Key="FloatingHintOffsetCalculationConverter" />
   <converters:ComboBoxClearButtonMarginConverter x:Key="ComboBoxClearButtonMarginConverter" />
   <converters:DoubleToThicknessConverter x:Key="DoubleToThicknessConverter" />
+  <converters:DoubleToThicknessConverter x:Key="DoubleToThicknessConverterWithNegativeOffset" InitialOffset="-8" />
   <converters:ThicknessCloneConverter x:Key="HelperTextMarginConverter" CloneEdges="Left" />
   <converters:ThicknessCloneConverter x:Key="DefaultOrFilledStyleTrailingIconPaddingConverterTop"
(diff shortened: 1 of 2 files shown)


MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.ComboBox.xaml
```diff
@@ -26,6 +26,7 @@
   <converters:FloatingHintOffsetCalculationConverter x:Key="FloatingHintOffsetCalculationConverter" />
   <converters:ComboBoxClearButtonMarginConverter x:Key="ComboBoxClearButtonMarginConverter" />
   <converters:DoubleToThicknessConverter x:Key="DoubleToThicknessConverter" />
+  <converters:DoubleToThicknessConverter x:Key="DoubleToThicknessConverterWithNegativeOffset" InitialOffset="-8" />
```
Keboo, line 29: Can we add a comment on this converter saying where this -8 value comes from.

nicolaihenriksen: Sure thing

nicolaihenriksen: Added comment

Keboo (approved):

PR #3280, opened 2023-07-19
Added ElementStringFormatProperty

jbooth88:
Added ElementStringFormatProperty to show an alternate string format when a cell is not in edit mode.

### DIFF
--- a/MaterialDesignThemes.Wpf/BindingEx.cs
+++ b/MaterialDesignThemes.Wpf/BindingEx.cs
@@ -1,0 +1,53 @@
+using System;
+using System.Windows.Data;
+
+namespace MaterialDesignThemes.Wpf
+{
+    internal static class BindingEx
+    {
+        /// <summary>Creates a shallow copy of a System.Windows.Data.Binding object.</summary>
+        /// <remarks>Bindings have an isSealed internal field that prevents it from being modified after it has been used.</remarks>
+        public static Binding Clone(this Binding source)
+        {
+            var result = new Binding()
+            {
+                AsyncState = source.AsyncState,
+                BindingGroupName = source.BindingGroupName,
+                BindsDirectlyToSource = source.BindsDirectlyToSource,
+                Converter = source.Converter,
+                ConverterCulture = source.ConverterCulture,
+                ConverterParameter = source.ConverterParameter,
+                FallbackValue = source.FallbackValue,
+                IsAsync = source.IsAsync,
+                Mode = source.Mode,
+                NotifyOnSourceUpdated = source.NotifyOnSourceUpdated,
+                NotifyOnTargetUpdated = source.NotifyOnTargetUpdated,
+                NotifyOnValidationError = source.NotifyOnValidationError,
+                Path = source.Path,
+                StringFormat = source.StringFormat,
+                TargetNullValue = source.TargetNullValue,
+                UpdateSourceExceptionFilter = source.UpdateSourceExceptionFilter,
+                UpdateSourceTrigger = source.UpdateSourceTrigger,
+                ValidatesOnDataErrors = source.ValidatesOnDataErrors,
+                ValidatesOnExceptions = source.ValidatesOnExceptions,
+                XPath = source.XPath,
+            };
+
+            //Note: setting more than one of [Source, ElementName, RelativeSource] will throw an exception.
+            if (source.Source is not null)
+            {
+                result.Source = source.Source;
+            }
+            else if (source.ElementName is not null)
+            {
+                result.ElementName = source.ElementName;
+            }
+            else if (source.RelativeSource is not null)
+            {
+                result.RelativeSource = source.RelativeSource;
+            }
+
+            return result;
+        }
+    }
+}
--- a/MaterialDesignThemes.Wpf/DataGridAssist.cs
+++ b/MaterialDesignThemes.Wpf/DataGridAssist.cs
@@ -339,12 +339,6 @@ public static class DataGridAssist
         => element.SetValue(CornerRadiusProperty, value);
     #endregion
 
-    #region AttachedProperty : ElementStringFormatProperty
-    public static readonly DependencyProperty ElementStringFormatProperty = DependencyProperty.RegisterAttached("ElementStringFormat", typeof(string), typeof(DataGridAssist), new PropertyMetadata(null));
-    public static string? GetElementStringFormat(MaterialDesignThemes.Wpf.DataGridTextColumn element) => element.GetValue(ElementStringFormatProperty) is string s ? s : null;
-    public static void SetElementStringFormat(MaterialDesignThemes.Wpf.DataGridTextColumn element, string? value) => element.SetValue(ElementStringFormatProperty, value);
-    #endregion
-
     private static void EditOnSpacebarPress(object sender, KeyEventArgs e)
     {
         var dataGrid = (DataGrid)sender;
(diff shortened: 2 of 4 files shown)


MaterialDesignThemes.Wpf/BindingEx.cs
```diff
@@ -0,0 +1,53 @@
+using System;
+using System.Windows.Data;
+
+namespace MaterialDesignThemes.Wpf
+{
+    internal static class BindingEx
```
nicolaihenriksen, line 6: This library typically names these classes/files with a suffix of "Extensions", so I think this should be `BindingExtensions`

MaterialDesignThemes.Wpf/DataGridAssist.cs
```diff
@@ -339,6 +339,12 @@ public static void SetCornerRadius(DataGrid element, CornerRadius value)
         => element.SetValue(CornerRadiusProperty, value);
     #endregion
 
+    #region AttachedProperty : ElementStringFormatProperty
+    public static readonly DependencyProperty ElementStringFormatProperty = DependencyProperty.RegisterAttached("ElementStringFormat", typeof(string), typeof(DataGridAssist), new PropertyMetadata(null));
+    public static string? GetElementStringFormat(System.Windows.Controls.DataGridTextColumn element) => element.GetValue(ElementStringFormatProperty) is string s ? s : null;
+    public static void SetElementStringFormat(System.Windows.Controls.DataGridTextColumn element, string? value) => element.SetValue(ElementStringFormatProperty, value);
```
nicolaihenriksen, line 345: This setter/getter pair seem to indicate that this works for normal `System.Windows.Controls.DataGridTextColumn` columns, however I don't believe that is the case because there you don't have control over the `GenerateElement()` method which you override in the MDIX variant of the `DataGridTextColumn` (file below).

MaterialDesignThemes.Wpf/DataGridTextColumn.cs
```diff
@@ -15,6 +15,25 @@ public class DataGridTextColumn : System.Windows.Controls.DataGridTextColumn
...
+                && t.GetBindingExpression(TextBlock.TextProperty)?.ParentBinding is Binding binding
+                )
+            {
+                var NewBinding = binding.Clone();
+                NewBinding.StringFormat = stringFormat;
+                t.SetBinding(TextBlock.TextProperty, NewBinding);
+            }
+
+            return element;
+        }
```
nicolaihenriksen, line 35: Although this works, I can imagine a similar override being needed in other column types if the attached DP (`ElementStringFormat`) should be leveraged in those.

For this reason, I think it may be cleaner to create a dedicated column type which encapsulates this type of functionality.

nicolaihenriksen (changes requested):
Thank you very much for submitting the PR!

After having thought about it, I personally still think that a dedicated column type (derived from `DataGridTemplateColumn`) which encapsulates this type of functionality is preferable. Perhaps something like `DataGridFormattableTextColumn` or something along those lines. Then you don't need an attached DP, and can just expose the desired properties directly on the column type

You could also consider exposing properties allowing a consumer to inject custom `DataTemplate` for both the "editing" and "not editing" variations to provide even more granular control to the consumer.

I will add @Keboo (repo maintainer) as a reviewer as well, he may see things different than me 😄

MaterialDesignThemes.Wpf/BindingEx.cs
jbooth88: Done in latest commit

MaterialDesignThemes.Wpf/DataGridAssist.cs
jbooth88: Good call. I originally thought this could apply to any DataGridTextColumn, but I left that thought halfway finished in the first commit.

MaterialDesignThemes.Wpf/DataGridTextColumn.cs
jbooth88: Fixed in latest commit

jbooth88: Don't worry about implementing any of this, I just wanted to show what I did to get my own project working and provide ideas for future development if a similar topic ever comes up. If you use/modify any or none of this code, no problem.

Thank you very much for your input!

MaterialDesignThemes.Wpf/DataGridTextColumn.cs
```diff
@@ -15,6 +17,30 @@ public class DataGridTextColumn : System.Windows.Controls.DataGridTextColumn
...
+        {
+            var element = base.GenerateElement(cell, dataItem);
+
+            if (
+                (cell.Column as MaterialDesignThemes.Wpf.DataGridTextColumn)?.ElementStringFormat is string stringFormat
+                && element is TextBlock t
+                && t.GetBindingExpression(TextBlock.TextProperty)?.ParentBinding is Binding binding
+                )
+            {
+                var NewBinding = binding.Clone();
```
nicolaihenriksen, line 30: Consider using camel case rather than pascal case for local variables. e.g. "newBinding" or "bindingClone" or similar.

MaterialDesignThemes.Wpf/DataGridTextColumn.cs
```diff
@@ -15,6 +17,30 @@ public class DataGridTextColumn : System.Windows.Controls.DataGridTextColumn
             return null;
         }
 
+        protected override FrameworkElement? GenerateElement(DataGridCell cell, object dataItem)
+        {
+            var element = base.GenerateElement(cell, dataItem);
+
+            if (
+                (cell.Column as MaterialDesignThemes.Wpf.DataGridTextColumn)?.ElementStringFormat is string stringFormat
```
nicolaihenriksen, line 25: Consider using pattern matching for the first part of this condition as well; effectively splitting it into 2 conditions.

nicolaihenriksen (commented):
To me this looks good. Moving the DP into the `DataGridTextColumn` makes it more discoverable and avoids an attached property being applied in places where it isn't really supported. I like that much better.

Keboo:
This works, but I am wondering if it is the right solution. Typically, when you need this level of disjoint between your view and edit cell templates, moving to a template column is the better option. It seems a bit odd to give only the StringFormat portion of the binding special treatment, I could see similar requests coming for things like the Converter property as well. I will need to think on this more, but I think simply pointing people to using a template column may be the better approach.

jbooth88:
@Keboo the more I thought about this over the last couple of weeks, the more I agree with you and @nicolaihenriksen that a DataGridTemplateColumn is the better approach. When I first came up with this idea I had two goals in mind:

1. Remove a significant amount of boilerplate code
2. More importantly for my own usage, have something that would work at runtime

Since first posting the original change request, I stumbled upon the solution I was originally looking for: how to create a custom DataGridTemplateColumn at runtime (why I have to do this is a whole nightmare on its own, customer wants a single view to manage [TPH](https://learn.microsoft.com/en-us/ef/core/modeling/inheritance#table-per-hierarchy-and-discriminator-configuration) view mdoels). I learned how to leverage the FrameworkElementFactory and create my own DataTemplates in code behind and everything is happy now.

I am not experienced in git, so at least for me I learned a bit on how to fork and submit change requests. Thank you guys very much for all the support! Let me know if I should be the one to cancel this change request.

Keboo:
@jbooth88 I talked a bit on this PR on my [live stream last night](https://keboo.twitch.com). Thank you for your willingness to open a PR. Happy to consider any other PRs you may have in the future.

PR #3292, opened 2023-08-11
Fix issue that IME gets misplaced when text input control is on a Popup

### LINKED ISSUE #3147: IME Misplaced
### Bug explanation

IME misplaces when textboxes are on DialogHost.

If textboxes are not on a DialogHost, IME behavior is correct, like this:
image

However, if a textbox is placed on a DialogHost, the IME will be misplaced.
image

### Version

4.8.0

AndrewKeepCoding:
This applies the workaround mentioned in this [issue](https://github.com/dotnet/wpf/issues/4315) on the WPF repo.
I confirmed that works with Japanese IME.

Fixes #3147.


Keboo (approved):

jschroedl:
Note that this workaround does not work if the TextBox is shown within a ContextMenu (as in my app). We have a menu item to filter quickly to a smaller set and the IME shows in the wrong location and the workaround causes the ContextMenu to close b/c it causes it to lose mouse capture. :-(

PR #3293, opened 2023-08-11
Adding launchSettings.json to demo apps

nicolaihenriksen:
I saw you mention this in your stream, and I have also wanted it for a while. It allows you to start the demo app(s) directly into a given demo item page.

It is a very simple implementation, simply using the first CLI argument (if any) to match on the name of the `DemoItem`. If not found (or not supplied), it defaults to the first one as it did before.

I put "Home" in the launchSettings.json files just to make it easier to see where you need to make a change if/when you have the need.


Keboo (approved):

Keboo:
Nice!

PR #3294, opened 2023-08-12
Add new options to the ExpanderAssist class

AbderrahmaneAhmam:
- ExpanderPosition: to set the position of the Expander Icon
image

  --> Start: at the start of the header materialDesign:ExpanderAssist.ExpanderPosition="Start"        
  --> End: at the end of the header materialDesign:ExpanderAssist.ExpanderPosition="End"

- CollapsedIcon: to set the icon when the expander is collapsed
  --> materialDesign:ExpanderAssist.CollapcedIcon="{materialDesign:PackIcon Kind=Plus}"
image

- ExpandedIcon: to set the icon when the expander is expanded
  --> materialDesign:ExpanderAssist.ExpandedIcon="{materialDesign:PackIcon Kind=Minus}"
image


Keboo (commented):
Two things.
1. It appears that it is only implemented for Expand direction Up and Down. This should also support ExpandDirection Left and Right (this is the 3a expanders in the demo app).

2. Rather than having two attached properties, I think a single one for the ToggleButton's content will work. Perhaps called something like `ExpanderButtonContent` (or similar). I think naming it so that consumers realize that it accepts arbitrary content, rather than just icons. The caller can then toggle it doing something like this:
```xaml
<Expander.Style>
  <Style TargetType="Expander" BasedOn="{StaticResource MaterialDesignExpander}">
    <Setter Property="materialDesign:ExpanderAssist.ExpanderButtonContent" Value="{materialDesign:PackIcon Kind=Minus}" />
    <Style.Triggers>
      <Trigger Property="IsExpanded" Value="True">
        <Setter Property="materialDesign:ExpanderAssist.ExpanderButtonContent" Value="{materialDesign:PackIcon Kind=Plus}" />
      </Trigger>
    </Style.Triggers>
  </Style>
</Expander.Style>
```

AbderrahmaneAhmam:
Hello @Keboo

- The **ExpandDirection** property is solely utilized to determine the opening direction of an expander control. On the other hand, the **ExpanderPosition** property is employed to adjust the position of the expander button either at the beginning or the end. This distinction is particularly valuable since certain _UI/UX_ designs call for the expander button to be positioned at the start. This could pose challenges in development, as it would necessitate creating the template anew.

- Regarding the "ExpanderButtonContent," you are correct. We can make it like this; it's easier.

Keboo:
> * The **ExpandDirection** property is solely utilized to determine the opening direction of an expander control. On the other hand, the **ExpanderPosition** property is employed to adjust the position of the expander button either at the beginning or the end. This distinction is particularly valuable since certain _UI/UX_ designs call for the expander button to be positioned at the start. This could pose challenges in development, as it would necessitate creating the template anew.

This is true, but we also use it in the Expander style to [swap to a different header template](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/master/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Expander.xaml#L312). I would expect that we need to add similar triggers for changing the expanding icon in the vertical template as well.

Keboo:
> Hello @Keboo ,
> 
> I've made some changes following your review. I combined the two properties, "Collapse" and "Expand" icons, into a single property called ExpanderButtonContent. Additionally, I renamed the "ExpanderPosition" property to ExpanderButtonPosition.
> 
> I have a question: is it possible to utilize these properties in the TreeView as well? This would allow for altering the expander button's content and position using the same assistant.

I think we could do something similar for tree view, but I would make it a new property on the TreeViewAssist class rather than re-using this one.

Keboo (approved):

PR #3312, opened 2023-09-01
Use System.CommandLine to configure initial app values

nicolaihenriksen:
So I am currently watching your stream and was intrigued by your "request" because I have actually wanted to play with `System.CommandLine` just to get a feel for it.

So this leverages `System.CommandLine` to enable optional command line arguments for setting the: initial demo page, initial theme, and initial flow direction.

I am quite certain that I have done things completely wrong, so please use this PR to educate me 😝 For starters, I actually did not want to add a dedicated "configure" command, but was unable to get the options working without having that command in place. Ideally I just want the options and a way to extract them (if set). Currently there isn't a shared place to put code for the two demo apps, so it is duplicated code in both projects right now.
<br/>

### **ALSO:** Regarding multi select `TreeView`!!
I have actually within the last year had to roll one of these. I went through a similar research process as you're currently doing in the stream I am watching. I ended up deriving from `TreeView` and also doing a custom `TreeViewItem` (derived from `TreeViewItem`). That worked out pretty well, with the caveat that I needed the "extra" `IsMultiSelected` property to handle my own selection logic. I still let the `TreeView` use the existing `IsSelected` property which I then basically just use to track "last selected item" which is useful in the multi select (CTRL/SHIFT) scenarios.

After having fought with it though, I have had a user story sitting in our backlog for a while, regarding refactoring it into a control deriving from `ListBox`/`ListView` instead. The primary motivation for doing this is that I believe virtualization can be handled much better in such a case. With the normal `TreeView`, I believe you only get "real" virtualization on the root level nodes, because sub-level nodes are actually part of the visual tree of the parent and that makes reusing of `DataTemplates` (i.e. item containers) kind of a mess; You end up with circular references and other horrid nightmares... However, I think to get the real benefit of virtualization, I have to flatten the hierarchy to a flat list and then handle the parent-child relationships in a custom manner.

Can you confirm that the hierarchy of `TreeViewItems` is problematic with regards to virtualization? I have not found a clear answer, but I think my many trial and error runs have indicated that only root/top-level items have proper virtualization.


Keboo:
I pushed a comment with what I am thinking on the command line stuff. Hopefully this helps make a bit more sense. I tried to leave a long commit message explaining everything, but let me know if you have any questions on it.

Keboo:
On the TreeViewItem question I believe the answer is yes it is problematic. It does set the VirtualizingStackPanel on each of the TreeViewItems but it would not surprising me to find that in many cases this does not properly result in the ItemsHost not properly virtualizing the items.

nicolaihenriksen:
> I tried to leave a long commit message explaining everything, but let me know if you have any questions on it

Thanks for the explanation. I had something quite similar at one point, but could not get the RootCommand to behave without adding the additional command. I must have done something else wrong.

You pushed this to it's own branch, are you merging that (to master? or my branch?). Or don't you want it on master at all?

nicolaihenriksen:
> It does set the VirtualizingStackPanel on each of the TreeViewItems but it would not surprising me to find that in many cases this does not properly result in the ItemsHost not properly virtualizing the items.

I think what I ran into was some issues where it was trying to reuse an items container which effectively was a grand-parent (X levels up) node scrolled out of view, and then it (quite naturally) complained that it already had visual ancestors and needed to be detached and stuff like that. So I ended up not using the `VirtualizationMode=Recycle`.

Keboo:
> > I tried to leave a long commit message explaining everything, but let me know if you have any questions on it
> 
> Thanks for the explanation. I had something quite similar at one point, but could not get the RootCommand to behave without adding the additional command. I must have done something else wrong.
> 
> You pushed this to it's own branch, are you merging that (to master? or my branch?). Or don't you want it on master at all?

I pushed it to your branch. Happy to merge whenever, just wanted to give you a chance to see it first.

MainDemo.Wpf/CommandLineOptions.cs
```diff
@@ -0,0 +1,42 @@
...
+            FlowDirectionOption,
+            ThemeOption
+        };
+
+    public static (string StartPage, FlowDirection FlowDirection, BaseTheme BaseTheme) ParseCommandLine(string[] args)
+    {
+        ParseResult parseResult = RootCommand.Parse(args);
+
+        return new(
+            parseResult.GetValueForOption(PageOption)!,
```
Keboo, line 37:
```suggestion
    public static (string? StartPage, FlowDirection FlowDirection, BaseTheme BaseTheme) ParseCommandLine(string[] args)
    {
        ParseResult parseResult = RootCommand.Parse(args);

        return new(
            parseResult.GetValueForOption(PageOption),
```

Keboo (approved):

Keboo (approved):

PR #3333, opened 2023-09-28
NEW FEATURE: Multi-select TreeView

Keboo:
This control is based on a ListView NOT a TreeView. 
The interaction is expected to be similar to a TreeView with a HierarchialDataTemplate, but the rest of the UI should match ListView behavior.
At present there is no support for ListView.GridView.


HppZ:
does it work well with virtualization?

Keboo:
> does it work well with virtualization?

Yea, because it is just a list view, as long as it has virtualization enabled it works well for that.

PR #3349, opened 2023-10-27
ListBoxAssist.IsToggle - Ensure last clicked item is always selected

### LINKED ISSUE #3235: ListBox toggle can be unselected
### Bug explanation

ListBox in toggle mode:

image

should not be able to get into the state
image

Now when I click the selected item, it is unselected.

### Version

4.9.0

nicolaihenriksen:
Fixes #3235 

As the last comment in the issue states, when explicitly applying `ListBoxAssist.IsToggle=true`, the expected behavior is most likely that the items should behave similar to a radio button group. This ensures that the user cannot do anything to end up in a situation where no elements are selected. However, there are still no guards around selection being "removed" from code.

To fix the latter part, we could attach a `PropertyChangedCallback` on the `IsToggle` DP, but we would probably also need to listen for changes in the `ListBox.Items` collection which I think is probably taking it a step too far. If the developer explicitly sets `ListBoxAssist.IsToggle=true`, let's assume that he/she also knows not to do anything weird.

### DIFF
--- a/MainDemo.Wpf/Toggles.xaml
+++ b/MainDemo.Wpf/Toggles.xaml
@@ -338,11 +338,11 @@
 
       <smtx:XamlDisplay HorizontalAlignment="Left" UniqueKey="buttons_63">
         <!-- the following based on https://material.io/guidelines/components/buttons.html#buttons-toggle-buttons -->
-        <ListBox SelectedIndex="0" Style="{StaticResource MaterialDesignToolVerticalToggleListBox}">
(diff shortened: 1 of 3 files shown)


Keboo:
This works, and I get that in most cases we probably expect the radio button group style behavior, but I am wondering if a better fix. But I am thinking there are cases where "none" would be valid. I am thinking we might want to allow that case to still exist but have this be the default.

nicolaihenriksen:
> But I am thinking there are cases where \"none\" would be valid.

If you read the issue, you will see I had the same concern. So do we need an AP to allow the none case then?

Keboo (approved):

Keboo:
Yea something like that looks good to me.

PR #3382, opened 2023-11-23
SplitButton leveraging PopupBox

nicolaihenriksen:
This PR adds a SplitButton to the MDIX library by introducing a `SplitButton` custom control which is a composite control consisting of a normal `Button` and a `PopupBox` next to it. It tries to leverage the existing button styles to avoid too much maintenance work.

<br/>

**PROs**:
* Discoverability (independent control residing in the MDIX namespace)
* Hope to have fine grained control if tweaks are needed that should not affect normal buttons (may need some work though!)
* Does not impact the visual tree of normal buttons

**CONs**:
* Some level of duplication is still present, but not a whole lot

<br/>

**NOTE:** I opted to **not** use the new cool resource key stuff generated by `mdresgen` yet. It will probably require changes in the `Button` templates as well, and as such I would rather get this in and then open a new PR where I can try to use the new features of the theme generation.

<br/>

Demo apps are extended with the following (on the Buttons page) showcasing all currently added styles:
image

Example usage is something along these lines:
```xaml
<materialDesign:SplitButton Content="Split Button"
                            Style="{StaticResource MaterialDesignOutlinedSplitButton}"
                            PopupPlacementMode="TopAndAlignRightEdges">
  <materialDesign:SplitButton.PopupContent>
    ...
  </materialDesign:SplitButton.PopupContent>
</materialDesign:SplitButton>
```


tim-gromeyer:
This is exactly what I was looking for. Would be very useful to have in my project. Being a beginner, I haven't been able to create such a button yet. I tried it several times but there was always something not working. Looking forward to this being merged!

PR #3411, opened 2023-12-07
This makes the automation tree look more like a regular tree view

Keboo:
This ends up displaying nested nodes as child automation peers of their parent.


MyZQL:
Does the TreeListView support to add headers and columns like this: 
image

Keboo:
@MyZQL at present it does not, though in theory it would be possible. The `ListView` (its base class) has a `GridView` property that could be utilized to do this. There is an outstanding TODO on this control to support that but at present that work has not been done (and likely wont be done for the 5.0.0 release)

Issue #3031, opened 2023-01-06
InvalidOperationException: Ripple cause InvalidOperationException to throw when opening widow from another thread


Roydoy7:
### Bug explanation

If a window is opened from another thread other than the main UI thread. If **mouse moves into the opened widow** there is a chance that will cause InvalidOperationException to throw.

To reproduce the issue:
Content of the MainWindows.xaml, if TreeViewItem is double clicked  open a window from another thread:
```
<Window x:Class="WpfApp3.MainWindow"
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
        xmlns:local="clr-namespace:WpfApp3"
        mc:Ignorable="d"
        Title="MainWindow" Height="450" Width="800">
    <Grid>
        <TreeView>
            <TreeViewItem Header="Will cause InvalidOperationException to throw" MouseDoubleClick="TreeViewItem_MouseLeftButtonDown"/>
        </TreeView>        
    </Grid>
</Window>
```

Content of the TreeViewItem_MouseLeftButtonDown that opens a window from another thread:
```
private void TreeViewItem_MouseLeftButtonDown(object sender, System.Windows.Input.MouseButtonEventArgs e)
        {
            var thread = new Thread(() =>
            {
                var window = new ThreadWindow();
                window.ShowDialog();
            });
            thread.SetApartmentState(ApartmentState.STA);
            thread.IsBackground = true;
            thread.Start();
        }
```

Content of the ThreadWindow.xaml, is an empty window:
```
<Window x:Class="WpfApp3.ThreadWindow"
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
        xmlns:local="clr-namespace:WpfApp3"
        mc:Ignorable="d"
        Title="ThreadWindow" Height="250" Width="300">
    <Grid>        
    </Grid>
</Window>

```
Call stack when this issue happens:
```
[...]

PR #3419, opened 2023-12-11
Fix issue 3031

quicoli:
This PR fixes issue #3031.

I was able to reproduce the issue.
Using the current Dispatcher fixes the issue. I'd like to discuss the implementation done. 

For .NET 6 and above there's a class called [DispatcherExtensions](https://learn.microsoft.com/en-us/dotnet/api/system.windows.threading.dispatcherextensions) that handles these thread issues offering a static access.

For .NET Framework we can achieve the same effect using `Dispatcher.CurrentDispatcher.CheckAccess()`  + `Dispatcher.CurrentDispatcher.Invoke`.


Keboo (approved):
Great work, thank you

Issue #2505, opened 2021-11-29
TranslateTransform animation in Flipper.FrontContent doesn't work properly


Roberto2707:
Hello.
After version 4.3.0-ci82, in Flipper.FrontContent, TranslateTransform animation doesn't work properly:
```
<Window x:Class="WpfApp1.MainWindow"
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:d="http://schemas.microsoft.com/expression/blend/2008" xmlns:local="clr-namespace:WpfApp1"
        xmlns:materialDesign="http://materialdesigninxaml.net/winfx/xaml/themes" xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
        Title="MainWindow"
        Width="300" Height="300"
        WindowStartupLocation="CenterScreen"
        mc:Ignorable="d">

    <Window.Resources>
        <Storyboard x:Key="LoadWindow">
            <DoubleAnimationUsingKeyFrames Storyboard.TargetName="txtUser" Storyboard.TargetProperty="(UIElement.RenderTransform).(TransformGroup.Children)[3].(TranslateTransform.X)">
                <EasingDoubleKeyFrame KeyTime="0" Value="800" />
                <EasingDoubleKeyFrame KeyTime="0:0:0.5" Value="800" />
                <EasingDoubleKeyFrame KeyTime="0:0:0.7" Value="0" />
            </DoubleAnimationUsingKeyFrames>
            <DoubleAnimationUsingKeyFrames Storyboard.TargetName="AcceptButton" Storyboard.TargetProperty="(UIElement.RenderTransform).(TransformGroup.Children)[3].(TranslateTransform.X)">
                <EasingDoubleKeyFrame KeyTime="0" Value="800" />
                <EasingDoubleKeyFrame KeyTime="0:0:0.5" Value="800" />
                <EasingDoubleKeyFrame KeyTime="0:0:0.9" Value="0" />
            </DoubleAnimationUsingKeyFrames>
            <DoubleAnimationUsingKeyFrames Storyboard.TargetName="CancelButton" Storyboard.TargetProperty="(UIElement.RenderTransform).(TransformGroup.Children)[3].(TranslateTransform.X)">
                <EasingDoubleKeyFrame KeyTime="0" Value="800" />
[...]

Issue #694, opened 2017-04-21
[Bug|missing feature] Flipper prevents content from changing colors on switching palette


Insire:
Reproduction:
- start MaterialDesignDemo app from the Demos folder in the solution
- switch to Cards view and check color of the Flipper and its child ColorZone being violet and lime accent
- then change palette in palette view to a different color for accent and primary
- switch back to Cards view

**Expectation:**
- Flipper and its child ColorZone have the selected colors

**Result:**
- Flipper and its child ColorZone still have the old colors (violet and lime)

The issue can be verified in my own [project](https://github.com/Insire/Maple) too
- start the app
- click the globe in the window header
- view with flipper opens
- click brush in window header
- change color in the opening view for primary and accent
- click globe again
- flipper still has the same color

Insire:
Here are some more information on the problem. 

I did some looking around on google, and from my limited understanding of the matter, it seems that there is no easy universal way to update dynamic resources in an animation. The common approach seems to be to handcode the original animation and additionally handcode the animation for the color change in code behind. Obviously thats not a viable approach for a control library.

Does anyone have a better insight on the topic? Did i miss something? This is a really cool control and it would be sad if it would have such limited functionality.

Aangbaeck:
I'm missing this badly as well. As it is now, the user has to select a theme and then restart the app for flippers to follow theme.

Tide:
This bug is still present.  Has anyone found a workaround/fix?

ChrisLawford:
I would love to know if anyone has a workaround for this aswell. I was wondering if there is anyway to get the theme colour from the Palette selector but had no luck.

Keboo:
Flagging this for review with the 2.5.0 release. This just means I will spend some time digging into the cause of the issue (maybe fixing depending on what it is) prior to releasing 2.5.0.

sebaJoSt:
I think this is still open. I tried to implement the flipper in my own app, but the colors are not updated when switching from light to dark mode and vice versa. The latest MaterialDesignDemo app shows the same behaviour.

bee-eater:
Push

Xaalek:
I'm going to check that

PR #3431, opened 2024-01-04
New FlipperControl

Keboo:
This control is functionally intended to replace the old Flipper control.
It has a similar animation to the ToggleButton. For people still wanting the old Flipper animation, the old control has been renamed to FlipperClassic. The animation is clearly different but it comes without all of the overhead of Plane3d animations

Breaking Change: Consumer using the `MaterialDesignCardFlipper` style will either need to delete the style reference or update to using the implicit key `{x:Type Flipper}` instead.
Alternatively, they can go back to the old behavior by replacing `Flipper` with `FlipperClassic` everywhere.

Flippers

Fixes : #694
Fixes: #2505
Fixes: #2729


ElieTaillard:
Also fixes : #2729

Esterl-Christian:
I think this issue is not solved. I use MaterialDesignThemes 4.9.0 for WPF

nicolaihenriksen:
> I think this issue is not solved. I use MaterialDesignThemes 4.9.0 for WPF (I need the nuget package)

The fix is in the 5.0 release which is not yet officially released. You can test the nightly preview builds where it is included.

PR #3451, opened 2024-02-08
Fixing issue with ComboBoxes in dialog host

Keboo:
This revert the fix that was done for #3249.
Need to revisit fix that was done and find an alternative.

@AndrewKeepCoding if you are able, do you mind taking a look and seeing if there is a way to limit how often we call `SetFocus` as it appear to close the popup in the ComboBox.

### DIFF
--- a/MaterialDesignThemes.UITests/Samples/DialogHost/WithComboBox.xaml
+++ b/MaterialDesignThemes.UITests/Samples/DialogHost/WithComboBox.xaml
@@ -1,0 +1,38 @@
+﻿<UserControl x:Class="MaterialDesignThemes.UITests.Samples.DialogHost.WithComboBox"
+             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+             xmlns:materialDesign="http://materialdesigninxaml.net/winfx/xaml/themes"
+             xmlns:d="http://schemas.microsoft.com/expression/blend/2008" 
+             xmlns:local="clr-namespace:MaterialDesignThemes.UITests.Samples.DialogHost"
+             mc:Ignorable="d" 
+             d:DesignHeight="450" d:DesignWidth="800">
+
+  <materialDesign:DialogHost x:Name="SampleDialogHost"
+                             Loaded="DialogHost_Loaded">
+    <materialDesign:DialogHost.DialogContent>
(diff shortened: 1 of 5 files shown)


AndrewKeepCoding:
@Keboo I created a test app with minimal code (C# only) that reproduces the issue.

These 3 reproduce the issue:

```cs
private async void Button_Click(object sender, RoutedEventArgs e)
{
    ComboBox comboBox = new ComboBox();
    comboBox.Items.Add("A");
    comboBox.Items.Add("B");
    comboBox.Items.Add("C");
    await MaterialDesignThemes.Wpf.DialogHost.Show(comboBox);
}
```

or

```cs
private void DialogHost_Loaded(object sender, RoutedEventArgs e)
{
    if (sender is not MaterialDesignThemes.Wpf.DialogHost dialogHost)
    {
        return;
    }

    ComboBox comboBox = new ComboBox();
    comboBox.Items.Add("A");
    comboBox.Items.Add("B");
    comboBox.Items.Add("C");
    dialogHost.ShowDialog(comboBox);
    // The code below also reproduces the issue.
    // dialogHost.DialogContent = comboBox;
    // dialogHost.IsOpen = true;
}
```

I also confirmed that by commenting out the `SetFocust()` on **DialogHost.cs** the `ComboBox` shows its items.

AndrewKeepCoding:
Assuming that we only need to set the focus ONCE when the dialog is opened, why not just remove the event handler?

```cs
private void OnPreviewGotKeyboardFocus(object sender, KeyboardFocusChangedEventArgs e)
{
    if (_popup != null &&
        PresentationSource.FromVisual(_popup.Child) is HwndSource hwndSource)
     {
        SetFocus(hwndSource.Handle);
        PreviewGotKeyboardFocus -= OnPreviewGotKeyboardFocus;
    }
}
```

This way nobody makes the `ComboBox` hide its items.

Keboo:
> Assuming that we only need to set the focus ONCE when the dialog is opened, why not just remove the event handler?
> 
> ```cs
> private void OnPreviewGotKeyboardFocus(object sender, KeyboardFocusChangedEventArgs e)
> {
>     if (_popup != null &&
>         PresentationSource.FromVisual(_popup.Child) is HwndSource hwndSource)
>      {
>         SetFocus(hwndSource.Handle);
>         PreviewGotKeyboardFocus -= OnPreviewGotKeyboardFocus;
>     }
> }
> ```
> 
> This way nobody makes the `ComboBox` hide its items.

Perfect, yea given that it is hard for me to reproduce the original issue that was fixed by this, I wanted to confirm I wasn't going to re-break it.

PR #3459, opened 2024-02-12
Add DialogHost.DialogCardStyle DP to allow control from calling code

### LINKED ISSUE #3430: DialogHost  Style
How can i change "wpf:Card.Style" with "MaterialDesignOutlinedCard"?

nicolaihenriksen:
Fixes #3430 

### UPDATE
The outstanding issue we have with this PR is the fact that calling code can effectively set a (card) style, but several properties from that style will be ignored because they are "passed-in/hardcoded" in the `PART_PopupContentElement` element in the `DialogHost` style.

---

This PR allows the calling code to control the style of the nested `Card` in the `DialogHost`. Default `Background` property is now set directly on the card, and the trigger to fall back to the default background is moved to the `ControlTemplate.Triggers` collection.

@Keboo I started on writing some UI test to verify this gets passed down, but I hit a roadblock around getting the `Style` (or perhaps just the name of the style) out of an `IVisualElement`, any hints you can provide?

### DIFF
--- a/MaterialDesignThemes.UITests/WPF/DialogHosts/DialogHostTests.cs
+++ b/MaterialDesignThemes.UITests/WPF/DialogHosts/DialogHostTests.cs
@@ -480,4 +480,56 @@ public class DialogHostTests : TestBase
 
         recorder.Success();
     }
+
+    [Theory]
+    [InlineData("MaterialDesignElevatedCard")]
+    [InlineData("MaterialDesignOutlinedCard")]
+    [Description("Issue 3430")]
+    public async Task DialogHost_DefaultStyleWithDialogCardStyleApplied_PassesStyleToNestedCard(string dialogCardStyle)
+    {
+        await using var recorder = new TestRecorder(App);
+
+        IVisualElement grid = await LoadXaml<Grid>($$"""
+           <Grid>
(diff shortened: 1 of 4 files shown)


MaterialDesignThemes.UITests/WPF/DialogHosts/DialogHostTests.cs
```diff
@@ -480,4 +480,56 @@ public async Task DialogHost_ChangesSelectedRailItem_DoesNotPerformRailChangeWhe
...
+                 <TextBlock Text="Some Text" Margin="100" />
+               </materialDesign:DialogHost.DialogContent>
+             </materialDesign:DialogHost>
+           </Grid>
+           """);
+
+        IVisualElement<DialogHost> dialogHost = await grid.GetElement<DialogHost>("DialogHost");
+        IVisualElement<Card> nestedCard = await dialogHost.GetElement<Card>("PART_PopupContentElement");
+
+        // TODO: Assert the style on the nested card matches what is set in DialogHost.DialogCardStyle
```
nicolaihenriksen, line 505: This is where I hit a road block. Any help would be appreciated 🙏

nicolaihenriksen:
@Keboo Since we have not settled on a good approach for this, and the PR is now horribly outdated compared to `master`, I am closing the PR. We can always revisit if/when we get a better idea.

PR #3470, opened 2024-02-27
Prepare for smart hint refactor

nicolaihenriksen:
In preparation for the `SmartHint` refactoring, I would like to make these changes/additions to the demo app page (and a converter cleanup) in order to avoid cluttering the `SmartHint` refactoring changes with a bunch of changes that are not really relevant for the changes being done.

This should hopefully simplify the PR that I will eventually create when I have the `SmartHint` cleaned up and working using the new approach.

### DIFF
--- a/MainDemo.Wpf/InputElementContentControl.cs
+++ b/MainDemo.Wpf/InputElementContentControl.cs
@@ -1,0 +1,8 @@
+﻿namespace MaterialDesignDemo;
+
+/// <summary>
+/// ContentControl variation simply used to capture the input elements in the "Smart Hint" demo page and apply some common properties.
+/// </summary>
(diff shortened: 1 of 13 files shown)


Keboo (approved):

MainDemo.Wpf/InputElementContentControl.cs
nicolaihenriksen, line 1: @Keboo Yes this replaces the use of ShowMeTheXAML, because, as your previous stream also showed, that feature does not really work on the SmartHint page because most properties are dynamically added or set via an inherited style, and thus do not show up in the ShowMeTheXAML window. So I opted to use the UI real-estate for the control instead, and make this simple wrapper to allow setting some common properties in a style.

PR #3490, opened 2024-03-11
Update migrations scripts

michaelmairegger:
Now by executing this script DynamicResources will be migrated too.
Furthermore I have added a new file MigrateStyles.ps1 which will update all ObsoleteStyles to the corresponding non-obsolete one


Keboo:
This is great work! Thank you

michaelmairegger:
@Keboo I have added a new script that will migrate the obsolete styles too. This script file was not present in the current master branch

Keboo:
I am fine with separate scripts, as I suspect some users will want to run only one or the other depending on their application. Thank you so much for doing this.

PR #3515, opened 2024-04-03
Adopt new SmartHint in ComboBox style

nicolaihenriksen:
This PR builds on top of the `SmartHint` refactoring branch (PR #3486). The intention was to document what it takes to adopt the new approach, but this PR actually covers a bit more than that because it also aligns the `ComboBox` style to match the layout of the `TextBox` a bit better, making it easier to make "the same change" in both styles if/when modifications are needed.

The actual changes for the `SmartHint` are relatively simple, but doing so often renders a lot of triggers/elements no longer needed, and thus results in a bit of housekeeping work. But this illustrates that it is indeed possible to adopt the new scheme one style at a time.


Keboo (approved):

Keboo:
Only one minor comment. The positioning of the floating hint on the outline style appears to be 1 pixel lower than on the text box style. Not a big deal but did catch my eye.

nicolaihenriksen:
> Only one minor comment. The positioning of the floating hint on the outline style appears to be 1 pixel lower than on the text box style. Not a big deal but did catch my eye.

Nice catch!! I think it is a rounding error issue. I had to add 0.5 to a converter offset to make it look correct (on my 150% DPI scaling 4K monitor)

PR #3527, opened 2024-04-18
fix 3526

### LINKED ISSUE #3526: ColorReference does not contain any color
### Bug explanation

When launching DemoApp on the official website, there was no response. When I pulled the code from GitHub and ran it locally, an error occurred: color reference does not contain any color. I have not encountered this situation on other computers before. Is it because my computer is missing something?

### Version

5.0.0

Ha-rlan:
Fixes #3526 

When AppsUseLightTheme does not exist in the corresponding path of the registry, give it a default value, otherwise an error will be reported, and there will be no prompt in DemoAppexe

### DIFF
--- a/MaterialDesignThemes.Wpf/Theme.cs
+++ b/MaterialDesignThemes.Wpf/Theme.cs
@@ -20,7 +20,7 @@ public partial class Theme
 
             if (registryValue is null)
             {
-                return null;
+                return BaseTheme.Light;
             }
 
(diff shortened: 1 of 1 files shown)


Keboo:
@henon thank you for tracking down the source of the problem and providing a suggested fix. 

Given that this API has been public for a few releases, I am hesitant to change its behavior went the registry key is not present. I have provided an alternate fix #3531 that addresses the null case at the caller to prevent the error from occurring. Closing this PR for now in favor of the other one.

PR #3529, opened 2024-04-21
Adopt new SmartHint in TimePicker style

nicolaihenriksen:
This PR adopts the new `SmartHint` approach in the `TimePicker` style.

It modifies the `TextBlock` style slightly, such that it can be leveraged in the `TimePicker` style instead of having a nearly full replica of the `TextBox` style.

* Extends `TextBox` with the option to control inactive/active border thickness. This was already present for `DatePicker`/`TimePicker` and is now generalized for the `TextBox`.
  - Obsoletes the more specific AP's which serve the same purpose but for `DatePicker`/`TimePicker` only.
* Maintains back-compat of obsoletions above using a `PriorityBinding`.
* Removes "replica" `TextBox`  style from the `TimePicker` template.
* Simplifies the `TimePicker` template and "adopts" the `SmartHint` approach by leveraging a full-fledged `TextBox` (with the appropriate style being forwarded in).

### DIFF
--- a/MaterialDesignThemes.Wpf/Converters/OutlinedDateTimePickerActiveBorderThicknessConverter.cs
+++ b/MaterialDesignThemes.Wpf/Converters/OutlinedDateTimePickerActiveBorderThicknessConverter.cs
@@ -4,7 +4,7 @@ using System.Windows.Data;
 namespace MaterialDesignThemes.Wpf.Converters;
 
 /* TODO: Should be obsoleted when both TimePicker and DatePicker have adopted the new SmartHint. */
-//[Obsolete("This class is obsolete and will be removed in a future version. Please use the more generic-named OutlinedStyleActiveBorderThicknessConverter instead.")]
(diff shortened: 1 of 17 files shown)


MaterialDesignThemes.Wpf/Converters/OutlinedDateTimePickerActiveBorderThicknessConverter.cs
```diff
@@ -3,6 +3,8 @@
 
 namespace MaterialDesignThemes.Wpf.Converters;
 
+/* TODO: Should be obsoleted when both TimePicker and DatePicker have adopted the new SmartHint. */
+//[Obsolete("This class is obsolete and will be removed in a future version. Please use the more generic-named OutlinedStyleActiveBorderThicknessConverter instead.")]
```
Keboo, line 7: :+1:

Keboo (approved):

PR #3548, opened 2024-04-29
Fix hover issue on DatePicker and TimePicker

### LINKED ISSUE #3547: DatePicker / TimePicker hover effect is lost when hovering calendar/time button
### Bug explanation

For controls that leverage the existing `TextBox` style and simply "draw on top" with additional content, the `TextBox` does not behave correctly with regards to mouse hovering. The reason is that the mouse is actually **not** hovering the `TextBox` or a child thereof, but rather a sibling which just so happens to draw some content on top of the `TextBox`.

This issue is currently present in the `TimePicker` and `DatePicker` controls.

TextFieldWrapperIsMouseOverIssue

### Version

latest on master (not released)

nicolaihenriksen:
Fixes #3547 

Please double check if you feel there are any negative side-effects to this fix. It is a slight hack in my opinion, but one that I currently cannot really see how we can avoid if we want to leverage the existing `TextBox` style in other "wrapper" controls like the `DatePicker` and `TimePicker`.

The PR introduces a new attached property which the "wrapper" (i.e. `DatePicker`/`TimePicker`) can use to communicate its `IsMouseOver` state down to the nested `TextBox`. The `TextBox` in turn relies on this `TextFieldAssist.IsMouseOver` to determinne whether the mouse is over it or not. To also handle the "normal case", it first inspects its own `IsMouseOver` for a `true` value and if so, overrides the value of the AP to `true` for the subsequent triggers to take effect.

I added 2 UI tests to cover the cases; red-green approach. Currently only for the outlined style, other tests could be added for the other styles if we feel the need.

**Note:** The color of the time/calendar button when it is hovered is the primary color; whether that is what we want or not I don't know, but that is a different issue.

Before:
TextFieldWrapperIsMouseOverIssue

After:
TextFieldWrapperIsMouseOverIssuePotentialFix

### DIFF
--- a/src/MaterialDesignThemes.Wpf/TextFieldAssist.cs
+++ b/src/MaterialDesignThemes.Wpf/TextFieldAssist.cs
@@ -9,15 +9,6 @@ namespace MaterialDesignThemes.Wpf;
 public static class TextFieldAssist
 {
     /// <summary>
-    /// Used by text field "wrappers" (i.e. controls hosting a text field and decorating on top of it) to signal to the text field that the mouse is over it,
-    /// when in fact it is over a sibling (i.e. something in the wrapper) which is visually placed on top of the text field.
-    /// </summary>
-    public static readonly DependencyProperty IsMouseOverProperty = DependencyProperty.RegisterAttached(
-        "IsMouseOver", typeof(bool), typeof(TextFieldAssist), new PropertyMetadata(default(bool)));
-    public static void SetIsMouseOver(DependencyObject element, bool value) => element.SetValue(IsMouseOverProperty, value);
-    public static bool GetIsMouseOver(DependencyObject element) => (bool) element.GetValue(IsMouseOverProperty);
-
-    /// <summary>
     /// The text box view margin property
     /// </summary>
     public static readonly DependencyProperty TextBoxViewMarginProperty = DependencyProperty.RegisterAttached(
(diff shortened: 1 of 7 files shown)


src/MaterialDesignThemes.Wpf/TextFieldAssist.cs
```diff
@@ -8,6 +8,15 @@ namespace MaterialDesignThemes.Wpf;
 /// </summary>
 public static class TextFieldAssist
 {
+    /// <summary>
+    /// Used by text field "wrappers" (i.e. controls hosting a text field and decorating on top of it) to signal to the text field that the mouse is over it,
+    /// when in fact it is over a sibling (i.e. something in the wrapper) which is visually placed on top of the text field.
+    /// </summary>
+    public static readonly DependencyProperty IsMouseOverProperty = DependencyProperty.RegisterAttached(
```
Keboo, line 15: I think this approach looks great. Though I am thinking we probably want to put this into a different class that is inside of the Internal namespace (but have the class be public) so that people don't assume this is a simply customization like the other APs in here.

nicolaihenriksen: @Keboo Fair point. Although I don't see an existing class that seems fit for purpose. If I create a new one, what should the naming be? It would be confusing to have another `TextFieldAssist` simply in another namespace...

image

Keboo (approved):

PR #3557, opened 2024-05-11
Use CommunityToolkit.Mvvm with Buttons view in demo

JLdgu:
Kept changes to each Command separate in order to make review less complex


Keboo (approved):
Really nice work. Thank you!

PR #3615, opened 2024-06-24
Align NumericUpDown style with other controls

### LINKED ISSUE #3614: NumericUpDown style does not align with other controls
### Bug explanation

image

I am currently working on a PR. Stay tuned

### Version

5.1.0

michaelmairegger:
This PR aligns the NumericUpDown style to the one of other controls (closes #3614)

# Tasks

- [x] Add to fields line up
- [x] Validation Style
- [x] Floating Hint
- [x] Helper Text
- [x] Suffix
- [x] Prefix

# Breaking

Currently IncreaseContent and DecreaseContent are ignored since this two values should be the same as in TimePicker, not changable.

> Should we use IncreaseContent and DecreaseContent or remove it. I would prefer to remove it and use it predefined as in TimePicker. If we ignore both properties, then we have to remove it from code before merging

|old|new|
|---|---|
|image|image|
||image|
||image|
||image|
||image|

### DIFF
--- a/src/MainDemo.Wpf/Domain/MainWindowViewModel.cs
+++ b/src/MainDemo.Wpf/Domain/MainWindowViewModel.cs
@@ -191,7 +191,7 @@ public class MainWindowViewModel : ViewModelBase
         };
 
         yield return new DemoItem(
-            "Field line up",
+            "Fields line up",
             typeof(FieldsLineUp),
             new[]
             {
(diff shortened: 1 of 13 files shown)


nicolaihenriksen:
> Should we use IncreaseContent and DecreaseContent or remove it. I would prefer to remove it and use it predefined as in TimePicker. If we ignore both properties, then we have to remove it from code before merging

First of all, nice work!

Regarding your comment: I believe we should use them. We should default them to the `PackIcon` values they have currently, but I see no reason not to let someone manipulate them from the outside. For example if they want up/down arrows instead of +/-.

As for your comment about the `TimePicker` I think it would also make sense there to make the content configurable; thus could be considered a bug or a missing feature.

michaelmairegger:
@nicolaihenriksen Ok, I will change the code that it will respect the user-configurated values for up/down

michaelmairegger:
@nicolaihenriksen Done

PR #3618, opened 2024-06-28
Move MaterialDesignCharacterCounterTextBlock back out to global scope

### LINKED ISSUE #3613: Cannot find resource named 'MaterialDesignCharacterCounterTextBlock'.
### Bug explanation

Application is crashing when I'm trying to use a style based on MaterialDesignCharacterCounterTextBlock. All worked well in 5.0 and below so I don't really know what is changed. 

Exception: Cannot find resource named 'MaterialDesignCharacterCounterTextBlock'. Resource names are case sensitive.

I tried merging MaterialDesignTheme.TextBox.xaml but still getting the same error. Error only appeared after I updated to 5.1

This is my style based on MaterialDesignCharacterCounterTextBlock...

    <Style
        x:Key="ShowCounterOnKeyboardFocus"
        BasedOn="{StaticResource MaterialDesignCharacterCounterTextBlock}"
        TargetType="{x:Type TextBlock}">
        <Setter Property="Height" Value="0" />
        <Style.Triggers>
            <DataTrigger Binding="{Binding IsKeyboardFocusWithin, RelativeSource={RelativeSource FindAncestor, AncestorType=TextBox}}" Value="True">
                <Setter Property="Height" Value="{x:Static sys:Double.NaN}" />
            </DataTrigger>
        </Style.Triggers>
    </Style>

### Version

5.1.0

nicolaihenriksen:
Fixes #3613 

Moving this particular style under the `TextBox` style was apparently a breaking change for some. This PR moves just that style (and its required converter) back out to the global scope.

This also means the `AutoSuggestBox` no longer needs its own copy of it.

### DIFF
--- a/src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.AutoSuggestBox.xaml
+++ b/src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.AutoSuggestBox.xaml
@@ -16,7 +16,6 @@
       <converters:CursorConverter x:Key="ArrowCursorConverter" FallbackCursor="Arrow" />
       <converters:CursorConverter x:Key="IBeamCursorConverter" FallbackCursor="IBeam" />
       <converters:MathConverter x:Key="MathMultiplyConverter" Operation="Multiply" />
-      <converters:StringLengthValueConverter x:Key="StringLengthValueConverter" />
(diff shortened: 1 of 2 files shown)


nicolaihenriksen:
Temporarily closed and re-opened to kick off a new PR verification run; it seems GitHub was having trouble starting the first one...

PR #3619, opened 2024-06-28
Transitioner: Short-circuit SelectionChanged event when not originating from Transitioner

### LINKED ISSUE #3268: Transitioner seems to hook into SelectionChanged events of descendants?
### Bug explanation

TL;DR: The `Transitioner` control is inspecting its descendants for controls that offer the `SelectionChanged` event and hooking their `SelectionChanged` events into its own. This seems like incorrect behavior; I would expect it to only report when its visible immediate child control has changed and not care about what is deep within those controls. I have a GitHub repo which demonstrates the issue in a minimum viable VS project.

I use a `Transitioner` to flip between several different panels, mostly programmatically. Certain panels require a little external setup when they become active, so I hook to the `Transitioner`'s `SelectionChanged` event, to see which of the child panels has become active. This works great. But recently I've run into a problem:  If one of those child panels have a descentant that offers its own `SelectionChanged` event (like `ListView`, `TabControl`, etc.), the `Transitioner` seems to be hooking into those events and forwarding them to its own `SelectionChanged` event.

GitHub repo to demonstrate:  [josh2112/TestTransitionerSelection](https://github.com/josh2112/TestTransitionerSelection)

This is a VS2022 solution containing a .NET 6.0 WPF application. Just clone & run.

Click on the "next" and "prev" buttons to transition between the two panels. Watch the debug output for `Transitioner.SelectionChanged` being called:

```
10:54:13:344	Transition from FrameworkElement (name=panel1) to FrameworkElement (name=panel2)
10:54:14:089	Transition from FrameworkElement (name=panel2) to FrameworkElement (name=panel1)
```

Now go to panel 2 and click some list items. In the debug output you can see `Transitioner.SelectionChanged` is reporting those changes as well:
```
10:54:16:839	Transition from ListItem (text=item 3) to ListItem (text=item 4)
10:54:16:839	Transition from ListItem (text=item 4) to ListItem (text=item 3)
```

I can work around this issue by looking at `SelectionChangedEventArgs.Source`, which will report either the `Transitioner` or the `ListView` but I still feel this is strange & unexpected behavior that should be classified as a bug.

Thanks!

### Version

4.9.0

nicolaihenriksen:
Potentially fixes #3268 

Effectively short-circuits the `SelectionChanged` event from bubbling up from the `Transitioner` when the original source of the event is not the `Transitioner` itself. Please consider if you see any negative side-effects to this; I could not really see any, but I would like your input on that as well.

### DIFF
--- a/src/MaterialDesignThemes.Wpf/Transitions/Transitioner.cs
+++ b/src/MaterialDesignThemes.Wpf/Transitions/Transitioner.cs
@@ -62,6 +62,7 @@ public class Transitioner : Selector, IZIndexController
         CommandBindings.Add(new CommandBinding(MoveFirstCommand, MoveFirstHandler));
         CommandBindings.Add(new CommandBinding(MoveLastCommand, MoveLastHandler));
         AddHandler(TransitionerSlide.InTransitionFinished, new RoutedEventHandler(IsTransitionFinishedHandler));
+        AddHandler(SelectionChangedEvent, new RoutedEventHandler(SelectionChangedEventHandler));
         Loaded += (sender, args) =>
         {
             if (SelectedIndex != -1)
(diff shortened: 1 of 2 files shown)


MichelMichels (approved):
Seems good, I added a test to your work and confirmed it failed on current code base and is succeeding on this branch

PR #3638, opened 2024-07-21
Demo: Move ButtonsViewModel to shared project

JLdgu:
The FloatingActionDemoCommand does not seem to be triggered from the PopupBox, should it be removed?


Keboo:
So the changes here look just fine. I think the demo app command is correct, however, the PopupBox I believe is the one with the bug. I suspect we need to revisit how that command is actually implemented since it does not appear to be very useful.

Keboo (approved):

JLdgu:
Not sure how to take this PR off draft.
I will raise a PR to change the FloatingActionDemoCommand to MVVM toolkit and also an issue with popup box

src/MaterialDesign3.Demo.Wpf/Buttons.xaml.cs
```diff
@@ -1,5 +1,5 @@
 using System.Diagnostics;
-using MaterialDesign3Demo.Domain;
+using MaterialDesignDemo.Shared.Domain;
```
krishna-rathi51, line 3: using System.Diagnostics;
using MaterialDesign3Demo.Domain;
using MaterialDesignDemo.Shared.Domain;

namespace MaterialDesign3Demo;

JLdgu: The `ButtonsViewModel`  used by the `DataContext` has moved from `MaterialDesign3Demo.Domain` to `MaterialDesignDemo.Shared.Domain` which is why I removed line 2 above.

Is there a reason to keep `MaterialDesign3Demo.Domain` even when it is unused?

PR #3661, opened 2024-08-29
NumericUpDown control support decimals value, and the value for each increase or decrease can be set.

futuresBoy:
NumericUpDown control support int and decimals value, and the value for each increase or decrease can be set.

### DIFF
--- a/Directory.packages.props
+++ b/Directory.packages.props
@@ -26,7 +26,7 @@
     <PackageVersion Include="System.Net.Http" Version="4.3.4" />
     <PackageVersion Include="System.Text.RegularExpressions" Version="4.3.1" />
     <PackageVersion Include="VirtualizingWrapPanel" Version="1.5.8" />
-    <PackageVersion Include="XAMLTest" Version="1.2.1" />
+    <PackageVersion Include="XAMLTest" Version="1.2.2" />
     <PackageVersion Include="xunit" Version="2.6.2" />
     <PackageVersion Include="xunit.runner.visualstudio" Version="2.5.4" />
     <PackageVersion Include="Xunit.StaFact" Version="1.1.11" />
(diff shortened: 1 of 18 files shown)


Keboo:
I am hesitant to change all of the types from `int` to `decimal` as that would be a breaking change for everyone currently using it.

I would propose the following alternate solution:
1. Create a new abstract `UpDownBase<T>` class. Most of the code could move into this base class with abstract members where needed.
2. Create a new DependencyProperty for `ValueStep` to represent the amount to be incremented or decremented.
3. Make the `NumericUpDown` derive from `UpDownBase<int>`
4. Create a new `DecimalUpDown` class and derive from `UpDownBase<decimal>`
5. Create a new implicit style for `DecimalUpDown` and have it based on the NumericUpDown style.

Thoughts?

futuresBoy:
I got your proposal. I will create a new DecimalUpDown base on NumericUpDown

futuresBoy:
Follow your proposal, I made the following changes:
1. Create a new abstract `UpDownBase<T>` class. 'NumericUpDown ' code move into this base class.
2. Besides, need define an interface 'IArithmetic ' to abstract the operation of the T-value between int and decimal and other Type.
3. Create a new DependencyProperty for `ValueStep` to represent the amount to be incremented or decremented.
4. Make the `NumericUpDown` derive from `UpDownBase<int>` and 'IArithmetic<int>'
5. Create a new `DecimalUpDown` class and derive from `UpDownBase<decimal>` and 'IArithmetic<int>'
6. Change TargetType 'NumericUpDown'  to 'UpDownBase', so the 'NumericUpDown ' and 'DecimalUpDown ' share the same style.

And in this way, it does't affect everyone who are using 'NumericUpDown' currently.

Keboo:
@futuresBoy this is great work. I spent some time on my like stream doing some updates to leverage the new generic math on .NET8. 

There is an issue with my testing library (XAMLTest) that I want to address then we can get this merged.

Keboo (approved):

PR #3664, opened 2024-09-03
added "ClockButtonVisibility" property to TimePicker

### LINKED ISSUE #3663: Make TimePicker Icon-Button hidable
It would be nice to hide the Clock-Button of the TimePicker control whenever I want to. For example I would like to hide the button, when the TimePicker is disabled. I could define a binding for that or just hide the button indefinitely.

Maybe with a dependency Property "bool HidePopupButton" or "Visibility HidePopupButton". So that I could use it like this:

````
<TimePicker HidePopupButton="True" />
or
<TimePicker HidePopupButton="{Binding IsEnabled, RelativeSource={RelativeSource Self}, Converter={StaticResource InvertBooleanConverter}}" />

````

We have similar controls in our project, where the button is hidden, when the control is disabled. With the TimePicker we want to have the same behaviour.

corvinsz:
added a `ClockButtonVisibility` property to the `TimePicker` to fix #3663 

TimePickerShowClickButton

### DIFF
--- a/src/MainDemo.Wpf/Pickers.xaml
+++ b/src/MainDemo.Wpf/Pickers.xaml
@@ -213,15 +213,31 @@
         </smtx:XamlDisplay>
       </StackPanel>
 
-      <smtx:XamlDisplay Margin="0,16,0,0"
-                        HorizontalAlignment="Left"
-                        VerticalAlignment="Top"
-                        UniqueKey="pickers_7">
-        <materialDesign:TimePicker x:Name="PresetTimePicker"
-                                   Width="100"
-                                   Is24Hours="True"
-                                   SelectedTimeChanged="PresetTimePicker_SelectedTimeChanged" />
-      </smtx:XamlDisplay>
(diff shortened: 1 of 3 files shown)


corvinsz:
I know it's hard to define rules as to when and what to test, but are there any guidelines what should be tested and what not?

For this change to the `TimePicker` I asked myself if that's worth writing a test for. I ended up not writing one because it seems like such a small change.

Keboo (approved):

Keboo:
> I know it's hard to define rules as to when and what to test, but are there any guidelines what should be tested and what not?
> 
> For this change to the `TimePicker` I asked myself if that's worth writing a test for. I ended up not writing one because it seems like such a small change.

In this case I believe your assessment is correct, this probably doesn't warrant any testing. The tests in this project are more integration tests then true unit tests and consequently have a higher cost to run. So it is always good to consider if a change really needs a test.

PR #3668, opened 2024-09-12
[Nuget] Inject correct copyright parameter into *.nuspec files

nicolaihenriksen:
The *.nuspec files expect a parameter named `copyright` to be injected, but this was being passed in as `copywrite` instead. This was done as part of #1538 and the Copyright on the produced nuget packages have been missing ever since.

I verified the fix locally.

### Before:
image

### After:
image


Keboo:
Nicely done. This might be the longest one of my spelling mistakes has lived.

Keboo (approved):

PR #3680, opened 2024-09-24
Fix typo and make ListBoxItems toggleable in Demo

corvinsz:
The ListBoxItem's ToolTip says it should behave like a toggle, but it didn't, because `materialDesign:ListBoxAssist.CanUserToggleSelectedItem="True"` was not set.


bebenins:
@corvinsz Nice!  I didn't know this was already built-in.  Thank you!  
Like you said, it'd be nice to have a true/false style button instead of ListBox.  But at least I have a working solution for now.

Keboo (approved):
Thank you for this!

Issue #3675, opened 2024-09-20
Unpredictable behavior of multiline ComboBox with hint


omllmo:
### Bug explanation

![1](https://github.com/user-attachments/assets/1bba54fe-b177-406a-b7e1-f0bef701130d)

Style Combobox - 'MaterialDesignFloatingHintComboBox'. 
If ComboBoxItem takes up several lines, then ComboBox begins to behave incorrectly.

```xaml
<Grid Width="500"  HorizontalAlignment="Stretch" Margin="15">
         <Grid.RowDefinitions>
                <RowDefinition Height="Auto"/>
                <RowDefinition Height="Auto"/>
           </Grid.RowDefinitions>
            <Grid.ColumnDefinitions>
                 <ColumnDefinition />
                 <ColumnDefinition Width="Auto" />
             </Grid.ColumnDefinitions>
             <materialDesign:PackIcon Kind="CardAccountMailOutline" Height="25" Width="25"/>
              <TextBlock Text="Test Test Test" Margin="30 2 0 10" Grid.ColumnSpan="3" VerticalAlignment="Bottom" FontSize="15"/>
               <Border HorizontalAlignment="Stretch" VerticalAlignment="Bottom" Grid.ColumnSpan="3"
                            BorderBrush="{StaticResource WindowTitleColorBrush}" BorderThickness="1" Height="1"/>
                <ComboBox FontSize="14"
                              Grid.Row="1" Margin="10 30 0 0"
                              materialDesign:HintAssist.Hint="Test Test Test Test "
                              materialDesign:HintAssist.IsFloating="True"
                              materialDesign:TextFieldAssist.OutlinedBorderActiveThickness="0 0 0 0"
                              materialDesign:HintAssist.FloatingOffset="0 -25"
                              materialDesign:HintAssist.FloatingScale="1"
                              ItemTemplate="{StaticResource This.DataTemplate.Combobox.Items.UserMinInfo}"
                              SelectionChanged="CBSetupRecipientVisibleButton_SelectionChanged"/>
</Grid>
```

### Version

5.1

omllmo:
Also, unpredictable behavior is observed in a multi-line TextBox with a hint. The hint also starts to float when the window size is changed.
![1](https://github.com/user-attachments/assets/dddcf699-abe2-47e1-9888-6c66afc91c94)
[...]

PR #3681, opened 2024-09-26
[SmartHint] Improvement for multi-line TextBox hint positioning

nicolaihenriksen:
This is intended as a partial-fix for #3675. Partial, because it only concerns the `TextBox` based styles, while a similar issue is present with the `ComboBox` template; I intend to do another PR to address that issue eventually.

### Problem
When a `TextBox` has multiple lines of text, either via `AcceptsReturn=True` or `TextWrapping=Wrap/WrapWithOverflow`, the hint position (and prefix/suffix text position) is not always correctly calculated.

#### Example:
image

### Proposed solution
In order to calculate the position correctly, knowledge about the number of rows/lines of text is needed. The information is available in the BCL on the `TextBox` type (see [here](https://source.dot.net/#PresentationFramework/System/Windows/Controls/TextBox.cs,909)), but unfortunately this is not a dependency property. To make the information available via a bindable property (i.e. attached property), I have created a `TextBoxLineCountBehavior` which hooks up to `TextBox.TextChanged` and effectively updates `TextFieldAssist.LineCount` and `TextFieldAssist.IsMultiLine` accordingly.

These attached properties are then used in bindings and converters to improve the calculation of the hint position.

### Example (GIF):
MultiLineFix

### Changes
* Extends the "Smart Hint" demo page with means of reproducing the issue(s).
* Adds `TextBoxLineCountBehavior`, `TextFieldAssist.LineCount` and `TextFieldAssist.IsMultiLine` to support correct calculations.
* Modifies `FloatingHintInitialVerticalOffsetConverter` which is responsible for doing the calculation.
* Modifies relevant styles to leverage new APs.
* Changes default value for `RichTextBox.VerticalContentAlignment` to `Center` because this seems to be the best fit if/when you want a hint in a `RichTextBox`; row/line count is not really available in this control, so hint placement can be wrong depending on how you "configure" the control.

### Considerations
This effectively registers a `TextBox.TextChanged` listener on every single instance of a `TextBox`/`AutoSuggestBox` in an app using the MDIX styles.

#### Measurements
I wrote a simple WPF app simply laying out a configurable number of `TextBox` instances in a `Canvas` (to avoid the overhead of a `Grid` having to calculate row/columns sizes). I then measured the startup time and memory footprint (3 times and took an average value; not exactly empirical data, but meh!). The results are listed in the table below. I was looking for a linear increase which would indicate an issue, but the differences I see are so insignificant that I believe the changes in the PR do not introduce a perf cost. The fact that startup time was faster for the 2000-case also indicates that there is no issue.

I ran it on my laptop with the following specs:
_Intel Core Ultra 7 165H, 1400 Mhz, 16 cores, 22 logical processors, 32GB RAM_

<table>
<tr>
  <td/>
  <td colspan="3">Before</td>
  <td colspan="3">After</td>
</tr>
<tr>
  <td># TextBoxes</td>
  <td>100</td>
  <td>1000</td>
  <td>2000</td>
  <td>100</td>
  <td>1000</td>
  <td>2000</td>
</tr>
<tr>
  <td>Startup time</td>
  <td>~2 sec</td>
  <td>~8.5 sec</td>
  <td>~13.2 sec</td>
  <td>~2.2 sec</td>
  <td>~8.5 sec</td>
  <td>~12.8 sec</td>
</tr>
<tr>
  <td>Memory footprint</td>
  <td>~711 MB</td>
  <td>~1028 MB</td>
  <td>~1280 MB</td>
  <td>~711 MB</td>
  <td>~1026 MB</td>
  <td>~1285 MB</td>
</tr>

### DIFF
--- a/src/MaterialDesignThemes.Wpf/Converters/FloatingHintInitialVerticalOffsetConverter.cs
+++ b/src/MaterialDesignThemes.Wpf/Converters/FloatingHintInitialVerticalOffsetConverter.cs
@@ -32,8 +32,6 @@ public class FloatingHintInitialVerticalOffsetConverter : IMultiValueConverter
             // that are actually visible on screen) to avoid moving the hint further away than the actual viewport.
             return Math.Max(0, (contentHostHeight - hintHeight) / 2 - (offsetMultiplier * hintHeight));
         }
-
-        PasswordBox pb = new();
         return 0.0;
     }
 
--- a/src/MaterialDesignThemes.Wpf/Converters/FloatingHintInitialVerticalOffsetConverter.cs
+++ b/src/MaterialDesignThemes.Wpf/Converters/FloatingHintInitialVerticalOffsetConverter.cs
@@ -3,15 +3,38 @@ using System.Windows.Data;
 
 namespace MaterialDesignThemes.Wpf.Converters;
 
+/// <summary>
+/// This converter is used to apply an initial vertical offset (downwards) of the floating hint in the case where the
+/// <see cref="SmartHint.FloatingTarget"/> is taller than the <see cref="SmartHint"/> itself. This is typically the case
+/// if a fixed (large) height is applied to the host control (e.g. <see cref="TextBox"/> or similar). In these cases the
+/// hint should not float directly on top of the <see cref="SmartHint.FloatingTarget"/>, but rather be pushed down to sit
+/// on top of the text inside the <see cref="SmartHint.FloatingTarget"/>.
+///
+/// There is an edge case that need to be dealt with, which is when the host element allows for text to wrap (i.e. in
+/// <see cref="TextBox"/> based templates). In this case, we need to take the  number of text rows/line count into account
+/// in the calculation.
+/// </summary>
 public class FloatingHintInitialVerticalOffsetConverter : IMultiValueConverter
 {
     public object? Convert(object?[]? values, Type targetType, object? parameter, CultureInfo culture)
     {
-        if (values is [double contentHostHeight, double hintHeight])
+        if (values is [double contentHostHeight, double hintHeight, int lineCount])
         {
-            return (contentHostHeight - hintHeight) / 2;
+            double offsetMultiplier = 0;
+            if (lineCount > 1)
+            {
+                // Edge case where there are multiple rows of text so we need to calculate how far the hint should be pushed down.
+                // If there are 2 rows, we need to reduce the offset by 0.5*height, 3 rows should reduce by 1*height, 4 rows should reduce by 1.5*height, etc.
+                offsetMultiplier = lineCount / 2.0 - 0.5;
+            }
+            // Set an initial offset in order to push the hint down to where the actual text is displayed.
+            // The value is clamped to be >= 0 which is needed for TextBoxes where a vertical scrollbar is needed (i.e. more lines
+            // that are actually visible on screen) to avoid moving the hint further away than the actual viewport.
+            return Math.Max(0, (contentHostHeight - hintHeight) / 2 - (offsetMultiplier * hintHeight));
         }
-        return 0;
+
+        PasswordBox pb = new();
+        return 0.0;
     }
 
     public object?[] ConvertBack(object? value, Type[] targetTypes, object? parameter, CultureInfo culture)
--- a/src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.TextBox.xaml
+++ b/src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.TextBox.xaml
@@ -336,7 +336,14 @@
               </MultiTrigger.Conditions>
               <Setter TargetName="PrefixTextBlock" Property="VerticalAlignment" Value="Stretch" />
               <Setter TargetName="SuffixTextBlock" Property="VerticalAlignment" Value="Stretch" />
-              <Setter TargetName="Hint" Property="VerticalAlignment" Value="Top" />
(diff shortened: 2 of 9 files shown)


src/MaterialDesignThemes.Wpf/Converters/FloatingHintInitialVerticalOffsetConverter.cs
```diff
@@ -3,15 +3,38 @@
...
+                offsetMultiplier = lineCount / 2.0 - 0.5;
+            }
+            // Set an initial offset in order to push the hint down to where the actual text is displayed.
+            // The value is clamped to be >= 0 which is needed for TextBoxes where a vertical scrollbar is needed (i.e. more lines
+            // that are actually visible on screen) to avoid moving the hint further away than the actual viewport.
+            return Math.Max(0, (contentHostHeight - hintHeight) / 2 - (offsetMultiplier * hintHeight));
         }
-        return 0;
+
+        PasswordBox pb = new();
```
campersau, line 36: Looks like this isn't needed.

```suggestion
```

nicolaihenriksen: Thank you! Committing dead-debug code was never intentional 😆

src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.TextBox.xaml
```diff
@@ -591,6 +600,13 @@
          TargetType="{x:Type TextBox}"
          BasedOn="{StaticResource MaterialDesignTextBoxBase}">
     <Setter Property="Padding" Value="{x:Static wpf:Constants.TextBoxDefaultPadding}" />
+    <Setter Property="wpf:BehaviorsAssist.Behaviors">
```
Keboo, line 603: Probably doesn't need to be addressed now, but I wonder if we should be concerned about someone using `BehaviorsAssist.Behaviors` to add in their own behaviors and inadvertently clearing this one. Might make for some difficult to diagnose issues later.

Keboo (approved):

src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.TextBox.xaml
nicolaihenriksen: That is a fair point! When thinking about it, it almost feels like the new behavior type I added should be public? That way you can at least add it back in. Would also make it easier when copy-pasting one of our styles to make certain tweaks...

AhmadKelany:
Would it be a good idea to make the suf text aligned to the last line of the text?

nicolaihenriksen:
> Would it be a good idea to make the suf text aligned to the last line of the text?

I believe we discussed this very topic when refactoring the `SmartHint`. If I recall correctly, we went with the top line of text to avoid the "jumpy UI" where the prefix/suffix moves in the vertical direction when the number of lines change. @Keboo correct me if I am wrong...

omllmo:
Is the TextBox fix already in version 5.1.1-ci776?

nicolaihenriksen:
> Is the TextBox fix already in version 5.1.1-ci776?

Yes, it should be.

nicolaihenriksen:
> Installed version 5.1.1-ci776. When entering text manually, the hint began to behave correctly. But if the text is specified via Binding="{Binding Content}", then the location of the hint remained incorrect with multi-line TextBox.

Is it a text with actual "newlines" in it, or is it simply wrapping using `TexWrapping=Wrap[WithOverFlow]`? Also, what is the value of `TextBox.AcceptsReturn` in your case? I would like to repro this locally. There is also a type that I need to change from internal to public, so I will be updating it anyways. Might as well look at this at the same time.

nicolaihenriksen:
> Do you plan to make the hint behave correctly with the specified HintAssist.FloatingOffset?

If this is set explicitly, that is an opt-out of automatic positioning, and thus you need to change it yourself based on changes in height/lines/etc. So I guess the answer is no.

@omllmo Can you elaborate on my questions in my previous comment, that will make it easier for me to repro locally. Thanks.

omllmo:
Wrapped with TexWrapping=Wrap[WithOverFlow].
TextBox.AcceptsReturn is the default value.

@nicolaihenriksen As I wrote above, setting VerticalContentAlignment="Top" solves the problem with incorrect placement of the hint.

nicolaihenriksen:
> As I wrote above, setting VerticalContentAlignment="Top" solves the problem with incorrect placement of the hint.

@omllmo That is good, but I would like it not to be necessary; I'll see if I can make it work.

omllmo:
In another situation, incorrect behavior of a multi-line TextBox with a hint was noticed (in particular, the MaterialDesignOutlinedTextBox style):
- Value VerticalAligment="Stretch"
- Value VerticalContentAligment="Top."

If you do not set the value VerticalContentAligment="Top.", then the text displayed in the TextBox is located in the center, not at the top

nicolaihenriksen:
> If you do not set the value VerticalContentAligment="Top.", then the text displayed in the TextBox is located in the center, not at the top

This is expected behavior. The default value (`VerticalContentAlignment=Stretch`) has been chosen to make the most-common use case work "out of the box", whereas other scenarios (like this one), may require you to change a couple of properties. Single line without a custom height is deemed as the most-common use case.

By default, the text vertically aligns to the center such that it (vertically) aligns with all the other elements that could possible be there: leading icon, prefix text, suffix text, trailing icon, clear button.

PR #3684, opened 2024-09-30
Set Text of ComboBox (DataGridComboBoxColumn) when IsEditing true

### LINKED ISSUE #3683: Invalid DataGridComboBoxColumn.IsEditable = True
### Bug explanation

I will reproduce it with the MaterialDesignDemo.exe app(.NET 8). In the first sample of the DataGrid, for the `Food with long header` column, even though `IsEditable="True"`, when you actually enter data and move to another cell, the entered value disappears.

### Version

5.1.0

corvinsz:
I essentially just added logic to set the `ComboBox.Text` when `GenerateElement(...)` is called:
```C#
if (cell?.Content is ComboBox cb)
{
    comboBox.Text = cb.Text;
}
```

Because `GenerateElement()` and `GenerateEditingElement()` share alot of logic I extracted some of it into a new method `InternalGenerateElement(...)`.

This should fix #3683


Keboo:
Thank you for this, but I think the current behavior is correct (see my comment on the original issue)

PR #3692, opened 2024-10-09
Return to ElementName binding to avoid runtime exception

### LINKED ISSUE #3522: DialogHost: XamlObjectWriterException - Unresolved reference 'ContentCoverBorder'
### Bug explanation

When running a Window with nothing else but a `DialogHost`, following exception gets thrown:
`System.Xaml.XamlObjectWriterException in System.Xaml.dll`
Additional text: `Unresolved reference 'ContentCoverBorder`.

It has something to do with this PR: https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/commit/bfe348fed9334ea36eebe97f0897ef01c96afb15

When I revert this change, the issue goes away but the binding error shows up again.

### Version

5.0.0

nicolaihenriksen:
Fixes #3522 

This PR effectively reverts this commit bfe348fed9334ea36eebe97f0897ef01c96afb15. But by simply reverting the commit, we re-introduce a temporary binding error. To mitigate this, I leverage the `FirstNonNullCoverter` to simply return the value of the binding or `DependencyProperty.UnsetValue` in case the value is `null`. This is essentially the same as a `PriorityBinding`, but the priority binding will show a XAML binding error, the converter approach does not.

Inspecting the `PART_ContentCoverGrid.OpacityMask.Visual` with snoop reveals that the correct value is applied at runtime, and that no binding errors have been reported.

image

image

### DIFF
--- a/src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.DialogHost.xaml
+++ b/src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.DialogHost.xaml
@@ -1,6 +1,7 @@
 ﻿<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
                     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-                    xmlns:wpf="clr-namespace:MaterialDesignThemes.Wpf">
+                    xmlns:wpf="clr-namespace:MaterialDesignThemes.Wpf"
+                    xmlns:converters="clr-namespace:MaterialDesignThemes.Wpf.Converters">
   <ResourceDictionary.MergedDictionaries>
     <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Font.xaml" />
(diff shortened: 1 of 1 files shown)


MichelMichels (approved):
This seems like a very nice workaround. Any idea why this binding error occurs when doing it the other way (with 1 binding instead of the multibinding)?

nicolaihenriksen:
> Any idea why this binding error occurs when doing it the other way (with 1 binding instead of the multibinding)?

I believe the issue lies inside the binding engine. When it is a "direct binding" (i.e. without a converter), a binding error will be reported even if is temporary (a `PriorityBinding` also reports the error). In other words, because the binding initially fails, a binding error is reported. Shortly after the binding works so there aren't any issues runtime.

By introducing the converter, we short-circuit the binding engines "error reporting" such that, by returning the `DependencyProperty.UnsetValue`, the binding will simply fallback to it's default/fallback value and NOT report an error. Once the real binding returns a proper value, that is used instead (effectively making this work in runtime). You could argue that there is a short moment in time where it does not work in runtime, but I don't believe that is an issue.

Keboo (approved):

PR #3696, opened 2024-10-16
Removing demo font size change

Keboo:


nicolaihenriksen:
@Keboo Just in case you didn't notice, the default startup page was also changed away from "Home". Consider reverting that as well. Not a big issue though...

PR #3699, opened 2024-10-17
[SmartHint] Avoid timing issue in hint position calculation

### LINKED ISSUE #3695: Incorrect behavior of hint in multiline TextBox (continuation)
### Bug explanation

1)
Hint1

Сode:
```xaml
<Window x:Class="WpfApp1.MainWindow"
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
        xmlns:materialDesign="http://materialdesigninxaml.net/winfx/xaml/themes"
        xmlns:local="clr-namespace:WpfApp1"
        mc:Ignorable="d"
        Title="MainWindow" Height="450" Width="800"
        Loaded="Window_Loaded">
    <materialDesign:DialogHost Name="dialogHost" Margin="10 0 0 0"
                               Style="{StaticResource MaterialDesignEmbeddedDialogHost}">
        <materialDesign:DialogHost.DialogContent>
            <Grid>
                <Grid Name="gridInDialogHost" Width="300"
                      Margin="15"
                      DataContext="{x:Null}" >
                    <Grid.RowDefinitions>
                        <RowDefinition Height="Auto"/>
                        <RowDefinition Height="Auto"/>
                        <RowDefinition Height="Auto"/>
                    </Grid.RowDefinitions>
                    <materialDesign:PackIcon Kind="PencilBoxMultiple" Height="25" Width="25"/>
                    <TextBlock Text="Test"
                               Margin="30 2 0 10" Grid.ColumnSpan="3"
                               VerticalAlignment="Bottom"
                               FontSize="15"/>
                    <Border HorizontalAlignment="Stretch" VerticalAlignment="Bottom"
                            BorderBrush="Blue" BorderThickness="1" Height="1"/>
                    <TextBox Grid.Row="1" Margin="10 20 0 10"
                             Text="{Binding Comment, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}"
                             VerticalAlignment="Stretch"
                             HorizontalAlignment="Stretch"
                             materialDesign:HintAssist.Hint="HINT TEST"
                             materialDesign:HintAssist.FloatingScale="1"
                             materialDesign:TextFieldAssist.IconVerticalAlignment="Top"
                             AcceptsReturn="True"
                             MaxLength="300"
                             Style="{StaticResource MaterialDesignFilledTextBox}"
                             TextWrapping="Wrap"
                             FontSize="14"/>
                    <Button Content="Close"
                            Grid.Row="2" Margin="0 25 5 5" HorizontalAlignment="Right" 
                            Width="100"
                            Command="{x:Static materialDesign:DialogHost.CloseDialogCommand}"
                            IsCancel="True"/>
                </Grid>
            </Grid>
        </materialDesign:DialogHost.DialogContent>
        <DataGrid Name="dataGrid"
                  FontSize="14" 
                  MinHeight="150"
                  AutoGenerateColumns="False" 
                  IsReadOnly="True" 
                  SelectionMode="Single"
                  MouseDoubleClick="DG_MouseDoubleClick">
            <DataGrid.Columns>
                <DataGridTextColumn Header="Description" 
                                    Binding="{Binding Comment}" 
                                    Foreground="Red"/>
            </DataGrid.Columns>
        </DataGrid>
    </materialDesign:DialogHost>
</Window>
```
----------------------------------------------------------------------------------------------
```csharp
    public partial class MainWindow : Window
    {
        ObservableCollection<TestClass> list;
        TestClass testClass;

        public MainWindow()
        {
            InitializeComponent();
        }

        private void Window_Loaded(object sender, RoutedEventArgs e)
        {
            list = new ObservableCollection<TestClass>
            {
                new TestClass() { Comment = "test test test test test test test test test test test test  test test test test test test test test test test test " }
            };

            dataGrid.ItemsSource = list;
        }

        private void DG_MouseDoubleClick(object sender, MouseButtonEventArgs e)
        {
            if (e.LeftButton == MouseButtonState.Pressed)
            {
                if (dataGrid.SelectedItem is TestClass tc && DataGrid.GetIsSelectionActive(VisualTreeHelper.HitTest(dataGrid, e.GetPosition(dataGrid)).VisualHit))
                {
                    gridInDialogHost.DataContext = null;

                    testClass = new TestClass() { Comment = tc.Comment };

                    gridInDialogHost.DataContext = testClass;

                    dialogHost.IsOpen = true;
                }
            }
        }
    }
```
----------------------------------------------------------------------------------------------
```csharp
public class TestClass : INotifyPropertyChanged
    {
        public event PropertyChangedEventHandler PropertyChanged;

        private int id;
        public int ID { get => id; set { id = value; RaisePropertyChanged("ID"); } }

        private string comment;
        public string Comment { get => comment; set { comment = value; RaisePropertyChanged("Comment"); } }

       
        protected virtual void RaisePropertyChanged(PropertyChangedEventArgs e)
        {
            PropertyChanged?.Invoke(this, e);
        }
        protected void RaisePropertyChanged(string propertyName)
        {
            RaisePropertyChanged(new PropertyChangedEventArgs(propertyName));
        }
    }
```

### Version

5.1.1-ci780

nicolaihenriksen:
Fixes #3695 

I was not able to track down exactly why the timing issue occurs, but it is evident in the issue mentioned above, that the same user interaction can cause different behavior which is a good indication of a timing related issue.

I tracked down the wrong calculation to this particular line of code in the converter:
https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/eb9338c720dd93fd53aef6d6499586efa63a87ca/src/MaterialDesignThemes.Wpf/Converters/FloatingHintTranslateTransformConverter.cs#L46
The result of this calculation is essential in the placement of the hint, but not one that can easily be used as a trigger for the converter to recalculate.

The fix here, which I would like your thoughts on, is to extract said calculation into a attached property which is calculated based on a change in layout, and then use the AP in the multi-binding.

This seems to do the trick, and I can no longer reproduce the issue in the provided repo sample.

### DIFF
--- a/src/MaterialDesignThemes.Wpf/Behaviors/SmartHintBehavior.cs
+++ b/src/MaterialDesignThemes.Wpf/Behaviors/SmartHintBehavior.cs
@@ -1,0 +1,39 @@
+﻿using Microsoft.Xaml.Behaviors;
+
+namespace MaterialDesignThemes.Wpf.Behaviors;
+
+public class SmartHintBehavior : Behavior<SmartHint>
+{
+    public static readonly DependencyProperty YOffsetProperty =
+        DependencyProperty.RegisterAttached("YOffset", typeof(double), typeof(SmartHintBehavior), new PropertyMetadata(0.0));
+    public static double GetYOffset(DependencyObject obj)
+        => (double)obj.GetValue(YOffsetProperty);
+    public static void SetYOffset(DependencyObject obj, double value)
+        => obj.SetValue(YOffsetProperty, value);
+
+    private void UpdateSmartHintLocationRecalculationTrigger()
+    {
+        if (AssociatedObject?.FloatingTarget is null) return;
+
+        double yOffset = AssociatedObject.FloatingTarget.TranslatePoint(new Point(0, 0), AssociatedObject).Y;
+        AssociatedObject.SetCurrentValue(YOffsetProperty, yOffset);
+    }
+
+    private void HintHostOnLayoutUpdated(object? sender, EventArgs e)
+        => UpdateSmartHintLocationRecalculationTrigger();
+
+    protected override void OnAttached()
+    {
+        base.OnAttached();
+        AssociatedObject.LayoutUpdated += HintHostOnLayoutUpdated;
+    }
+
+    protected override void OnDetaching()
+    {
+        if (AssociatedObject != null)
+        {
+            AssociatedObject.LayoutUpdated -= HintHostOnLayoutUpdated;
+        }
+        base.OnDetaching();
+    }
+}
--- a/src/MaterialDesignThemes.Wpf/Converters/FloatingHintTranslateTransformConverter.cs
+++ b/src/MaterialDesignThemes.Wpf/Converters/FloatingHintTranslateTransformConverter.cs
@@ -1,4 +1,4 @@
-﻿using System.Globalization;
+using System.Globalization;
 using System.Windows.Data;
 using System.Windows.Media;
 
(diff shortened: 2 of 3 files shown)


Keboo (approved):
Seems like a good fix to me!

PR #3701, opened 2024-10-19
[SmartHint] Add opt-in AP to control hint padding brush

### LINKED ISSUE #3700: Weird black sidebars on SmartHints
### Bug explanation

We've been using materialDesign for wpf for a while now but since the last update those TextBox:

Hlk9BM2O

That looks like this in our XML:
```
<TextBox
Grid.Row="1"
Style="{StaticResource MaterialDesignOutlinedTextBox}"
GotFocus="SizeTextBox_GotFocus" LostFocus="TextBox_LostFocus"
PreviewKeyDown="TextBox_PreviewKeyDown"
materialDesign:HintAssist.Hint="{x:Static p:Resources.Height}"
FontSize="25" Height="60" Padding="10" Margin="20"
materialDesign:TextFieldAssist.SuffixText="{Binding Unit, Converter={StaticResource UnitAbbreviation}}"
BorderBrush="{StaticResource PrimaryColorBrush}"
materialDesign:HintAssist.Background="{StaticResource ThirdColorBrush}"
materialDesign:ValidationAssist.Background="{StaticResource ThirdColorBrush}">
    <TextBox.Text>
        <Binding Path="SelectedParameter.Height" UpdateSourceTrigger="Explicit">
            <Binding.ValidationRules>
            <utility:NumericValidationRule/>
            </Binding.ValidationRules>
        </Binding>
    </TextBox.Text>
</TextBox>
```
have those weird black sidebars.

I tried to modify each and everyone props you can access through materialDesign:HintAssist but nothing did the trick.

I also tried to use this to modify the hint :

`
<TextBox.Resources>
    <Style TargetType="{x:Type materialDesign:SmartHint}">
        <Setter Property="Background" Value="red"/>
    </Style>
</TextBox.Resources>
`

It seems I can modify some stuff, mostly props accessible through HintAssist, but setting some other properties doesn't change a thing, margin and padding do nothing for example.

Sorry if that's not a bug but me using a materialStyle i shouldn't associate with hints or something else, but it looks like a bug.

EDIT : Sorry for the badly written issue, it's my first. What you linked seems to be exactly it i'll explore it in my workday thanks !

### Version

4.5

nicolaihenriksen:
Fixes #3700 
Fixes #3262 

If applying a custom background, the `MaterialDesignOutlinedTextBox` style by default uses the MDIX window background brushes for the "padding" on the hint (used to "erase" some of the outline border).

This PR introduces an opt-in AP which allows consumers to override this to match their custom background.

In principle, a `PriorityBinding` should suffice for this in the `SmartHint` style, but yet again I was unable to make it work 😞 Thus I went for the converter approach I have ended up with before. Apparently I am not great friends with the `PriorityBinding`.

For a more detailed explanation of the issue, see my comment here: https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/discussions/3262#discussioncomment-10905148

### Before
`HintAssist.Background` left at default value (`Transparent`):
image

`HintAssist.Background` set to match background:
image

### After
`HintAssist.Background` not needed unless you want a custom background on the hint (floating and non-floating)
image

### DIFF
--- a/src/MaterialDesignThemes.Wpf/HintAssist.cs
+++ b/src/MaterialDesignThemes.Wpf/HintAssist.cs
@@ -123,10 +123,10 @@ public static class HintAssist
     public static readonly DependencyProperty HintPaddingBrushProperty =
         DependencyProperty.RegisterAttached("HintPaddingBrush", typeof(Brush), typeof(HintAssist), new PropertyMetadata(null));
 
-    public static Brush GetHintPaddingBrush(DependencyObject obj)
+    public static Brush? GetHintPaddingBrush(DependencyObject obj)
         => (Brush)obj.GetValue(HintPaddingBrushProperty);
 
-    public static void SetHintPaddingBrush(DependencyObject obj, Brush value)
+    public static void SetHintPaddingBrush(DependencyObject obj, Brush? value)
         => obj.SetValue(HintPaddingBrushProperty, value);
     #endregion
 
--- a/src/MaterialDesignThemes.Wpf/HintAssist.cs
+++ b/src/MaterialDesignThemes.Wpf/HintAssist.cs
@@ -119,6 +119,17 @@ public static class HintAssist
         => element.SetValue(BackgroundProperty, value);
     #endregion
 
+    #region AttachedProperty: HintPaddingBrush
+    public static readonly DependencyProperty HintPaddingBrushProperty =
+        DependencyProperty.RegisterAttached("HintPaddingBrush", typeof(Brush), typeof(HintAssist), new PropertyMetadata(null));
+
+    public static Brush GetHintPaddingBrush(DependencyObject obj)
+        => (Brush)obj.GetValue(HintPaddingBrushProperty);
+
+    public static void SetHintPaddingBrush(DependencyObject obj, Brush value)
+        => obj.SetValue(HintPaddingBrushProperty, value);
+    #endregion
+
     #region AttachedProperty : HelperTextProperty
     public static readonly DependencyProperty HelperTextProperty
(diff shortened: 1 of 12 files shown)


src/MaterialDesignThemes.Wpf/HintAssist.cs
```diff
@@ -119,6 +119,17 @@ public static void SetBackground(DependencyObject element, Brush value)
...
 
+    #region AttachedProperty: HintPaddingBrush
+    public static readonly DependencyProperty HintPaddingBrushProperty =
+        DependencyProperty.RegisterAttached("HintPaddingBrush", typeof(Brush), typeof(HintAssist), new PropertyMetadata(null));
+
+    public static Brush GetHintPaddingBrush(DependencyObject obj)
+        => (Brush)obj.GetValue(HintPaddingBrushProperty);
+
+    public static void SetHintPaddingBrush(DependencyObject obj, Brush value)
+        => obj.SetValue(HintPaddingBrushProperty, value);
```
Keboo, line 130:
```suggestion
    public static Brush? GetHintPaddingBrush(DependencyObject obj)
        => (Brush)obj.GetValue(HintPaddingBrushProperty);

    public static void SetHintPaddingBrush(DependencyObject obj, Brush? value)
        => obj.SetValue(HintPaddingBrushProperty, value);
```

nicolaihenriksen: Doh! I always forget this when using the built-in macros. Thanks 👍

Keboo (approved):

PR #3715, opened 2024-11-01
Using freezable objects for enhanced performance

ArshadAriff:
Added docs for Using freezable objects for enhanced performance


ArshadAriff:
@Keboo !

Keboo (approved):

PR #3732, opened 2024-11-26
Removed global converter resources

JorisCleVR:
- Removed the global converter resources
- Introduced Instance static readonly field(s) on converters whenever possible
  - Used the Instance field whenever applicable because it is better for the overall performance
- Introduced ObsoleteConverters.xaml to easily include these deprecated converter keys in consumer projects

### DIFF
--- a/src/MaterialDesignThemes.Wpf/Converters/BooleanToVisibilityConverter.cs
+++ b/src/MaterialDesignThemes.Wpf/Converters/BooleanToVisibilityConverter.cs
@@ -2,6 +2,9 @@
 
 public sealed class BooleanToVisibilityConverter : BooleanConverter<Visibility>
 {
+    public static readonly BooleanToVisibilityConverter Instance = new();
+    public static readonly BooleanToVisibilityConverter InverseInstance = new() { FalseValue = Visibility.Visible, TrueValue = Visibility.Hidden };
+
     public BooleanToVisibilityConverter() :
         base(Visibility.Visible, Visibility.Collapsed)
     { }
--- a/src/MaterialDesignThemes.Wpf/Converters/InvertBooleanConverter.cs
+++ b/src/MaterialDesignThemes.Wpf/Converters/InvertBooleanConverter.cs
@@ -1,7 +1,8 @@
 ﻿namespace MaterialDesignThemes.Wpf.Converters;
 
-internal class InvertBooleanConverter : BooleanConverter<bool>
+public class InvertBooleanConverter : BooleanConverter<bool>
 {
+    public static readonly InvertBooleanConverter Instance = new();
     public InvertBooleanConverter()
         : base(false, true)
     {
--- a/src/MaterialDesignThemes.Wpf/Converters/NullableToVisibilityConverter.cs
+++ b/src/MaterialDesignThemes.Wpf/Converters/NullableToVisibilityConverter.cs
@@ -5,6 +5,9 @@ namespace MaterialDesignThemes.Wpf.Converters;
 
 public class NullableToVisibilityConverter : IValueConverter
 {
+    public static readonly NullableToVisibilityConverter Instance = new();
+    public static readonly NullableToVisibilityConverter InverseInstance = new() { NullValue = Visibility.Visible, NotNullValue = Visibility.Hidden };
+
     public Visibility NullValue { get; set; } = Visibility.Collapsed;
     public Visibility NotNullValue { get; set; } = Visibility.Visible;
 
--- a/src/MaterialDesignThemes.Wpf/Converters/TextFieldHintVisibilityConverter.cs
+++ b/src/MaterialDesignThemes.Wpf/Converters/TextFieldHintVisibilityConverter.cs
@@ -5,6 +5,9 @@ namespace MaterialDesignThemes.Wpf.Converters;
 
 public class TextFieldHintVisibilityConverter : IValueConverter
 {
+    public static readonly TextFieldHintVisibilityConverter Instance = new();
+    public static readonly TextFieldHintVisibilityConverter StringIsEmptyInstance = new() { IsEmptyValue = Visibility.Collapsed, IsNotEmptyValue = Visibility.Visible};
+
     public Visibility IsEmptyValue { get; set; } = Visibility.Visible;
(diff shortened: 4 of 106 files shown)


src/MaterialDesignThemes.Wpf/Converters/BooleanToVisibilityConverter.cs
```diff
@@ -2,6 +2,9 @@
 
 public sealed class BooleanToVisibilityConverter : BooleanConverter<Visibility>
 {
+    public static readonly BooleanToVisibilityConverter Instance = new();
+    public static readonly BooleanToVisibilityConverter InverseInstance = new() { FalseValue = Visibility.Visible, TrueValue = Visibility.Hidden };
```
Keboo, line 6: I think we probably want a different name here since it is not actually the inverse of `Instance`. Thoughts on `NotHiddenInstance`?

src/MaterialDesignThemes.Wpf/Converters/InvertBooleanConverter.cs
```diff
@@ -1,7 +1,8 @@
 ﻿namespace MaterialDesignThemes.Wpf.Converters;
 
-internal class InvertBooleanConverter : BooleanConverter<bool>
+public class InvertBooleanConverter : BooleanConverter<bool>
```
Keboo, line 3:
```suggestion
public sealed class InvertBooleanConverter : BooleanConverter<bool>
```

src/MaterialDesignThemes.Wpf/Converters/TextFieldHintVisibilityConverter.cs
```diff
@@ -5,6 +5,9 @@ namespace MaterialDesignThemes.Wpf.Converters;
 
 public class TextFieldHintVisibilityConverter : IValueConverter
 {
+    public static readonly TextFieldHintVisibilityConverter Instance = new();
+    public static readonly TextFieldHintVisibilityConverter StringIsEmptyInstance = new() { IsEmptyValue = Visibility.Collapsed, IsNotEmptyValue = Visibility.Visible};
```
Keboo, line 9: Should the name of this be inverted?
```suggestion
    public static readonly TextFieldHintVisibilityConverter StringIsNotEmptyInstance = new() { IsEmptyValue = Visibility.Collapsed, IsNotEmptyValue = Visibility.Visible};
```

src/MaterialDesignThemes.Wpf/Converters/NullableToVisibilityConverter.cs
```diff
@@ -5,6 +5,9 @@ namespace MaterialDesignThemes.Wpf.Converters;
 
 public class NullableToVisibilityConverter : IValueConverter
 {
+    public static readonly NullableToVisibilityConverter Instance = new();
+    public static readonly NullableToVisibilityConverter InverseInstance = new() { NullValue = Visibility.Visible, NotNullValue = Visibility.Hidden };
```
Keboo, line 9: I think we probably want a different name here since it is not actually the inverse of Instance. Thoughts on `NotHiddenInstance`?

Keboo (commented):
This is great work. Just a couple small comments on naming things, then I will make sure we add information on this change to the release notes.

src/MaterialDesignThemes.Wpf/Converters/BooleanToVisibilityConverter.cs
JorisCleVR: Good point. I based it on the "InverseBooleanToVisibilityConverter" that was previously defined as a static resource.
The inverse in this case is based on the inverting of the boolean value not the visibility. I am kind of used to it,
The converters I personally use have 4 instances defined: "HiddenInstance", "HiddenInvertInstance", "CollapsedInstance", "CollapsedInvertInstance". This since we also have cases in which we want to collapse things.

Would it be useful to also introduce these?
So then we would have a "HiddenInstance", "NotHiddenInstance", "CollapsedInstance", "NotCollapsedInstance"

Though I personally prefer the Invert one, but I am a bit biased in this case.

src/MaterialDesignThemes.Wpf/Converters/NullableToVisibilityConverter.cs
JorisCleVR: This will be changed to the same naming convention that we decide for the BooleanToVisibilityConverter.

src/MaterialDesignThemes.Wpf/Converters/TextFieldHintVisibilityConverter.cs
JorisCleVR: Agreed, Personally I would prefer to name "Instance" "StringIsEmptyInstance", so we then have a "StringIsEmptyInstance" and a "StringIsNotEmptyInstance". Then it is not implicit anymore what the "Instance" version means.

Furthermore what do you think about removing the "String" part, since the converter already states that it is for a "TextField" implicitly meaning it is a "String". Therefore having a "IsEmptyInstance" and a "IsNotEmptyInstance" would make more sense.

src/MaterialDesignThemes.Wpf/Converters/BooleanToVisibilityConverter.cs
Keboo: Yea I like that naming better, I am fine adding these ones in.

Keboo (approved):

theHitchenator:
While I appreciate the logic behind this change, it has the frustrating impact of preventing external code's access to some of the previously accessible instances of these converters due to their classes being marked as internal.
This would be fine except if someone has built their own templates (slight variations on MaterialDesign templates, bringing their own usercontrols in line with the MaterialDesign styles, etc.) using those resources, they now have to duplicate the classes for those converters. Not the end of the world, just mildly annoying.
I'm hoping I've missed an obvious provision for this eventuality.

JorisCleVR:
@theHitchenator That is indeed an oversight when creating this PR. 

I created a new PR to make all these converters public: https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/pull/3767

PR #3754, opened 2025-01-02
Added Attached Property "TextFieldAssist.ClearButtonSize"

### LINKED ISSUE #3751: The CloseCircle Icon of TextFieldAssist. HasClearButton is too small.
image

The software runs on a tablet and can be controlled by touching it with your hand. The clear button in the text box is too small for touch operation. Can you add a property to adjust the size of the clear button icon? Thanks!

corvinsz:
this PR adds a new Attached Property to the class `TextFieldAssist` called `ClearButtonSize`, it should fix #3751


Keboo (commented):
I like this change but can we apply the sizing to the button itself rather than to the pack icon. After that I think it is good.

corvinsz:
You are right, I changed it accordingly.
I also changed the `Height` & `Width` of the `PackIcon` to Auto, otherwise I won't stretch.

PR #3764, opened 2025-01-09
[SmartHint] Hint placement for "readonly" controls

### LINKED ISSUE #3762: Faulty Hint in ComboBox Styles
### Bug explanation

In the `Floating`, `Filled` and `Outlined` ComboBox Styles the AP `HintAssist.Hint` is being pushed right by the prefix content of the ComboBox.
In all other controls the Hint is aligned left:
image

### Version

5.2.0

nicolaihenriksen:
Fixes #3762 

Initially I thought this was just a fix for the `ComboBox` styles, but it turns out that it was actually a common issue when the control is "readonly" (e.g. `TextBox.IsReadOnly=true` or `ComboBox.IsEditable=False`).

While investigating the root cause, I stumpled upon 2 other issues that I fixed as well. So this PR covers:

1. Fixing a very weird issue in the `SmartHint` demo app page, where the initial calculation of hint would sit on top of the prefix text. I was not able to reproduce this outside of the `SmartHint` demo app which puzzles me quite a bit. To "fix" it, I added a hack to the demo app, where it simply overrides the prefix/suffix texts in a `Loaded` event handler on the demo app page itself; this seems to kick the calculation into gear. Another way to manually "fix" it is to simply clear the prefix text in the UI, and then set it to back to what it was (or anything non-emtpy basically).

2. The Material Design 2 spec is quite clear in the visualization of a `ComboBox` in various states. We did not follow that completely since we did not float the hint once the drop-down was visible. A small change in `SmartHint.cs` regarding `HintHost.IsKeyboardFocusWithin` fixes that issue. See the MD2 spec [here](https://m2.material.io/components/menus#exposed-dropdown-menu) (scroll down a small bit further to the "Behavior" section).

3. Fixes the actual issue reported in the linked issue above. For "readonly" controls, we have a "business rule" that the prefix/suffix texts are always visible. For editable controls, the prefix/suffix visibility depends on the value of `TextFieldAssist.PrefixTextVisibility` which defaults to `WhenFocusedOrNonEmpty`. In the "readonly" case, the fact that the prefix/suffix is always visible was not accounted for, this PR changes that.

### Before
image

image

### After
image

image

@corvinsz @Keboo Could you guys please check out this branch and give it a short spin? Try to play around with `Smart Hint` and `Fields line up` demo pages, the former gives you a lot of control over how the hint/prefix/suffix and control behavior in general. It also showcases the different alignments, as well as all the controls supporting the `SmartHint` control.

And a small note-to-self: Think twice before undertaking a major refactoring of something as complex as the `SmartHint` and the placement-hell that lies within...

### DIFF
--- a/src/MainDemo.Wpf/SmartHint.xaml.cs
+++ b/src/MainDemo.Wpf/SmartHint.xaml.cs
@@ -17,10 +17,24 @@ public partial class SmartHint : UserControl
     internal static void SetRichTextBoxText(DependencyObject element, object value) => element.SetValue(RichTextBoxTextProperty, value);
     internal static object GetRichTextBoxText(DependencyObject element) => element.GetValue(RichTextBoxTextProperty);
 
+    private SmartHintViewModel ViewModel { get; }
+
     public SmartHint()
     {
-        DataContext = new SmartHintViewModel();
+        DataContext = ViewModel = new SmartHintViewModel();
         InitializeComponent();
+
+        Loaded += SmartHint_Loaded;
+    }
+
+    private void SmartHint_Loaded(object sender, RoutedEventArgs e)
+    {
+        // HACK! For some strange reason, the calculation of the left margin for the hint is initially wrong if these values are set as default in the view model directly.
+        // Setting them here is a bit hacky, but it makes the demo page work, and I don't think this would be an issue in a real world application so I can live the hack.
+        // To see the issue in action: Simply comment out the 2 lines below, open the "Smart Hint" page and toggle the "IsReadOnly" checkbox in the "TextBox styles" section;
+        // that will place the hint on top of the prefix text for unknown reasons.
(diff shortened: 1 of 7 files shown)


corvinsz:
@nicolaihenriksen I have fiddled around with your branch and everything I have tested works fine.
I have no objections, good work!

While testing I have noticed a problem (#3766) with the Pre-/Suffix but that's probably for another PR, not for this one.

nicolaihenriksen:
> While testing I have noticed a problem (#3766) with the Pre-/Suffix but that's probably for another PR, not for this one.

Thank you for testing!! I have commented on the new issue; I think it is the default "read only" mode of the ComboBox that makes it give off the wrong impression. Let me know if my comment makes sense...

corvinsz:
Yes, your comment does make sense, my bad. I closed it. 😆

Keboo (approved):

Issue #3761, opened 2025-01-07
AutoSuggestBox has weird tab behavior


nicolaihenriksen:
### Bug explanation

When using the demo app, the `AutoSuggestBox` on the `Fields` demo page is behaving strangely. Basically I am experiencing 2 different issues, I'll describe them both here, then we can split them into separate issues if need be.

* When tabbing into the `AutoCompleteBox` using keyboard navigation (if put keyboard focus in a prior element, and hit tab until it gets keyboard focus), you are unable to tab out of the control. TAB does nothing, CTRL+TAB does nothing, and SHIFT+TAB does nothing.
* If the suggestion list of the `AutoCompleteBox` has been visible but has been closed again (I believe this is the prerequisite), the TAB key now "works", but instead of tabbing out of the control, it now selects an element from the (invisible!) suggestion list.

### Version

5.2.0

corvinsz:
fixed in #3765

corvinsz:
@sandfox922 maybe I am missunderstanding your "not select anything", but I can't reproduce it on `5.2.2-ci933`.
The behavior seems fine to me:

![Image](https://github.com/user-attachments/assets/059e255f-7afc-4adb-91ba-563d8c588191)

sandfox922:
When the suggestion list is displayed, if I intentionally deselect an item using the up/down arrow keys and then press TAB or ENTER, the input text disappears. 

![Image](https://github.com/user-attachments/assets/c3fe2f43-521c-427b-9310-bbebe21a8523)

corvinsz:
@sandfox922 I see. This however is different from what @nicolaihenriksen initially reported in this issue. It indeed seems like a bug though, could you please open a new issue for this so we can track it separately?

PR #3765, opened 2025-01-09
"fix" AutoSuggestBox has weird tab behavior

corvinsz:
potential fix for #3761 

From my manual testing this fixes both problems mentioned in the issue.

This is a first "working" version but probably not ready for prod yet.
The changes essentially only mark the `KeyEventArgs e` event as handled, if one of the `AutoSuggestBox` "specific" keys is pressed. Additionally the method `CommitValueSelection` now returns a `bool` to indicate whether an item was selected or not.

I think this is test-worthy and therefore WIP


Keboo (approved):

JeanPhilippeLux:
Hi, do you know when a new stable version with this fix will be release ?

Thank you.

PR #3767, opened 2025-01-10
Made all converter classes public

JorisCleVR:
https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/pull/3732 removed global converter resources.
Unfortunately some converters were internal instead of public and therefore became inaccessible.
This PR makes all converters public.


corvinsz:
I'm questioning if all converters should straight up be made `public`. This means we have to support and maintain them in the future.
Wouldn't it be better to keep `internal` ones `internal`, and only make the ones `public` that were `public` before?

JorisCleVR:
That is also a possibility. I get why you want to prevent maintaining them all.
Unfortunately the issue is that several internal classes were publicly exposed as a StaticResource. Which is the issue people who are wanting to use them are now running into.
I think most of the ones that were internal were publicly available.

If you can give me a list of the ones were not available as a public StaticResource I can keep them internal.

Keboo:
So I think making these public is probably fine. There is also an argument to be made, that since some people need to copy styles/templates into their own projects to make modifications having these also be internal makes the process more difficult as well. As a compromise, we added an Internal namespace and made the classes public. The idea here being that the library reserves the right to treat them as internal for the purpose of semantic versioning (there could be breaking API changes on minor versions), but the classes remain public to allow for people to continue to use them. 

Let me review these converters, but I think a compromise might simply be to create an Internal namespace within the converters namespace, and move some of the more purpose-built converters in there.

src/MaterialDesignThemes.Wpf/Converters/ClockItemIsCheckedConverter.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace MaterialDesignThemes.Wpf.Converters;
 
-internal class ClockItemIsCheckedConverter(Func<DateTime> currentTimeGetter, ClockDisplayMode displayMode, bool is24Hours) : IValueConverter
+public class ClockItemIsCheckedConverter(Func<DateTime> currentTimeGetter, ClockDisplayMode displayMode, bool is24Hours) : IValueConverter
```
Keboo, line 6: This one I think we can leave internal since its only usages were in code and not XAML.

src/MaterialDesignThemes.Wpf/Converters/DialogBackgroundBlurConverter.cs
```diff
@@ -8,7 +8,7 @@
 using System.Windows.Media.Effects;
 
 namespace MaterialDesignThemes.Wpf.Converters;
-internal sealed class DialogBackgroundBlurConverter : IMultiValueConverter
+public sealed class DialogBackgroundBlurConverter : IMultiValueConverter
```
Keboo, line 11: I propose moving this inside of a new `MaterialDesignThemes.Wpf.Converters.Internal` namespace.

src/MaterialDesignThemes.Wpf/Converters/DoubleToCornerRadiusConverter.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace MaterialDesignThemes.Wpf.Converters;
 
-internal class DoubleToCornerRadiusConverter : IValueConverter
+public class DoubleToCornerRadiusConverter : IValueConverter
```
Keboo, line 6: I propose moving this inside of a new `MaterialDesignThemes.Wpf.Converters.Internal` namespace.

src/MaterialDesignThemes.Wpf/Converters/DoubleToThicknessConverter.cs
```diff
@@ -3,8 +3,10 @@
 
 namespace MaterialDesignThemes.Wpf.Converters;
 
-internal class DoubleToThicknessConverter : IValueConverter
+public class DoubleToThicknessConverter : IValueConverter
```
Keboo, line 6: This one looks like it has no usages. Can we just remove it instead?

src/MaterialDesignThemes.Wpf/Converters/ExpanderRotateAngleConverter.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace MaterialDesignThemes.Wpf.Converters;
 
-internal class ExpanderRotateAngleConverter : IValueConverter
+public class ExpanderRotateAngleConverter : IValueConverter
```
Keboo, line 6: I propose moving this inside of a new `MaterialDesignThemes.Wpf.Converters.Internal` namespace.

src/MaterialDesignThemes.Wpf/Converters/FloatingHintOffsetCalculationConverter.cs
```diff
@@ -5,7 +5,7 @@
 namespace MaterialDesignThemes.Wpf.Converters;
 
 // TODO: This converter can be deleted when the new SmartHint approach from TextBox style is applied throughout.
-internal class FloatingHintOffsetCalculationConverter : IMultiValueConverter
+public class FloatingHintOffsetCalculationConverter : IMultiValueConverter
```
Keboo, line 8: I believe the TODO above this has already been done, so I think this converter can now be deleted.

src/MaterialDesignThemes.Wpf/Converters/FloatingHintTextBlockMarginConverter.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace MaterialDesignThemes.Wpf.Converters;
 
-internal class FloatingHintTextBlockMarginConverter : IMultiValueConverter
+public class FloatingHintTextBlockMarginConverter : IMultiValueConverter
```
Keboo, line 7: I propose moving this inside of a new `MaterialDesignThemes.Wpf.Converters.Internal` namespace.

src/MaterialDesignThemes.Wpf/Converters/FloatingHintTransformConverter.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace MaterialDesignThemes.Wpf.Converters;
 
-internal class FloatingHintTransformConverter : IMultiValueConverter
+public class FloatingHintTransformConverter : IMultiValueConverter
```
Keboo, line 7: It looks like the only remaining references to this are in tests, can we just remove it along with the tests?

src/MaterialDesignThemes.Wpf/Converters/SliderToolTipConverter.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace MaterialDesignThemes.Wpf.Converters;
 
-internal class SliderToolTipConverter : IMultiValueConverter
+public class SliderToolTipConverter : IMultiValueConverter
```
Keboo, line 6: I propose moving this inside of a new `MaterialDesignThemes.Wpf.Converters.Internal` namespace.

src/MaterialDesignThemes.Wpf/Converters/SliderValueLabelPositionConverter.cs
```diff
@@ -4,7 +4,7 @@
 namespace MaterialDesignThemes.Wpf.Converters;
 
 [ValueConversion(typeof(double), typeof(double), ParameterType = typeof(Orientation))]
-internal class SliderValueLabelPositionConverter : IValueConverter
+public class SliderValueLabelPositionConverter : IValueConverter
```
Keboo, line 7: I propose moving this inside of a new `MaterialDesignThemes.Wpf.Converters.Internal` namespace.

src/MaterialDesignThemes.Wpf/Converters/TextFieldClearButtonVisibilityConverter.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace MaterialDesignThemes.Wpf.Converters;
 
-internal class TextFieldClearButtonVisibilityConverter : IMultiValueConverter
+public class TextFieldClearButtonVisibilityConverter : IMultiValueConverter
```
Keboo, line 6: I propose moving this inside of a new `MaterialDesignThemes.Wpf.Converters.Internal` namespace.

src/MaterialDesignThemes.Wpf/Converters/TopThicknessConverter.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace MaterialDesignThemes.Wpf.Converters;
 
-internal class TopThicknessConverter : IValueConverter
+public class TopThicknessConverter : IValueConverter
```
Keboo, line 6: Since this only appears to be referenced by `ObsoleteConverters.xaml` can we leave it as internal as the only usage of it would be from someone needing those obsolete converters?

I would propose, that we also decorate this class with the `[Obsolete]` so it is clear it is no longer used directly by this library.

src/MaterialDesignThemes.Wpf/Converters/DoubleToThicknessConverter.cs
JorisCleVR: Indeed no references, so removed it

src/MaterialDesignThemes.Wpf/Converters/FloatingHintTransformConverter.cs
JorisCleVR: Indeed no more usages, so removed it

tests/MaterialDesignThemes.Wpf.Tests/Converters/SliderToolTipConverterTests.cs
```diff
@@ -11,7 +11,14 @@ public sealed class SliderToolTipConverterTests
...
     public void SliderConverterTest(object value)
     {
+
+/* Unmerged change from project 'MaterialDesignThemes.Wpf.Tests (net8.0-windows)'
+Before:
         Wpf.Converters.SliderToolTipConverter converter = new ();
+After:
+        SliderToolTipConverter converter = new ();
+*/
+        Wpf.Converters.Internal.SliderToolTipConverter converter = new ();
```
Keboo, line 21: Not sure I understand this comment. Is it still needed?

```suggestion
        Wpf.Converters.Internal.SliderToolTipConverter converter = new ();
```

Keboo (commented):
Thank you for doing that! It looks like two comments were missed (FloatingHintTextBlockMarginConverter, and FloatingHintOffsetCalculationConverter), and one additional comment.

src/MaterialDesignThemes.Wpf/Converters/FloatingHintOffsetCalculationConverter.cs
JorisCleVR: Removed

JorisCleVR:
> Thank you for doing that! It looks like two comments were missed (FloatingHintTextBlockMarginConverter, and FloatingHintOffsetCalculationConverter), and one additional comment.

Good find. I just found out that github can "hide" comments under a button. Therefore I overlooked them while doing these changes. I handled them now.

Keboo:
> @Keboo Good find. I just found out that github can "hide" comments under a button. Therefore I overlooked them while doing these changes. I handled them now.

Awesome. Thank you for the quick update!

Keboo (approved):

PR #3773, opened 2025-01-21
add RadioButton and CheckBox custom foreground

mzy666888:
add RadioButton and CheckBox custom foreground.


nicolaihenriksen:
@mzy666888 Just had a quick glance at the changes. I will let @Keboo decide if this should merge or not, but is there a particular reason you have changed the new default brushes such that they don't match what they were before? This seems like a visual breaking change.

Keboo:
Hi @mzy666888, thank you for this PR. Can you elaborate further on the problem you are trying to solve with this. It looks like both of those brushes could have been set inside of your app already. 

I am a little hesitant to merge what was two different brushes into a single one.

mzy666888:
thank you Keboo, like example, I just want to change the color for checkbox  Box foreground and  don't want to change the color for content foreground.

if leave it as it is, I will define custom style on my project.

PR #3774, opened 2025-01-21
MD3 ToggleButton - Add fallback value to binding to avoid division by NaN

### LINKED ISSUE #3740: MaterialDesignSwitchToggleButton:  'System.Windows.Media.Animation.DoubleAnimation' cannot use default origin value of 'NaN' on Window Exit
### Bug explanation

Moving to Material Design 3. Having a problem with `MaterialDesignSwitchToggleButton`.  The other ToggleButtons styles don't appear to have this exception. 

When closing the parent window after this style has been loaded, I get an exception. I am guessing the exit animation is causing this. Not sure if it's because there is no `from` value? 

https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/5df62fa8faa82b198320f764e6b0a7de4d2f49c7/src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.ToggleButton.xaml#L381

I am still researching this and have not created a side project to test this yet.  That will be next, but under pressure to finish this project, so will select a different style for the meantime.

Exception is crazy long, so will post the final part of it.

```
System.InvalidOperationException: 'System.Windows.Media.Animation.DoubleAnimation' cannot use default origin value of 'NaN'.
   at System.Windows.Media.Animation.DoubleAnimation.GetCurrentValueCore(Double defaultOriginValue, Double defaultDestinationValue, AnimationClock animationClock)
   at System.Windows.Media.Animation.DoubleAnimationBase.GetCurrentValue(Object defaultOriginValue, Object defaultDestinationValue, AnimationClock animationClock)
   at System.Windows.Media.Animation.AnimationStorage.GetCurrentPropertyValue(AnimationStorage storage, DependencyObject d, DependencyProperty dp, PropertyMetadata metadata, Object baseValue)
   at System.Windows.Media.Animation.AnimationStorage.EvaluateAnimatedValue(PropertyMetadata metadata, EffectiveValueEntry& entry)
   at System.Windows.UIElement.EvaluateAnimatedValueCore(DependencyProperty dp, PropertyMetadata metadata, EffectiveValueEntry& entry)
```

More to come - but opening this up just in case there is something obvious I may be missing.

Xaml usage:
```xaml
<ToggleButton Style="{DynamicResource MaterialDesignSwitchToggleButton}"
              ToolTip="Something"
              IsChecked="{Binding SomeBoolBinding, Mode=TwoWay}"
              Margin="2" />
```

### Version

5.1.0

nicolaihenriksen:
Fixes #3740 

The bindings changed in this PR are bound to a `Tag` property which becomes `null` when the control is unloaded. This results in an exception being thrown from the `DoubleAnimation` presumably because `DoubleAnimation.To` (of type `double`) being bound to `null` results in `double.NaN`.

To avoid the exception, this PR simply applies a fallback value of 0 to the binding, which effectively mitigates the issue.

I don't believe this should affect the visual appearance of the `ToggleButton` since the `Tag` binding will have a valid value as long as the element is loaded/visible.

Using Snoop, it is also possible to observe the `Tag` value changing from 16 to '' (null I suppose).

### DIFF
--- a/src/MaterialDesignThemes.Wpf/Themes/MaterialDesign3.ToggleButton.xaml
+++ b/src/MaterialDesignThemes.Wpf/Themes/MaterialDesign3.ToggleButton.xaml
@@ -377,7 +377,7 @@
                 </DoubleAnimation>
                 <DoubleAnimation Storyboard.TargetName="Thumb"
                                  Storyboard.TargetProperty="Width"
-                                 To="{Binding Tag, ElementName=SwitchGrid}"
+                                 To="{Binding Tag, ElementName=SwitchGrid, FallbackValue=0}"
                                  Duration="0:0:0.2">
                   <DoubleAnimation.EasingFunction>
                     <QuadraticEase EasingMode="EaseOut" />
@@ -385,7 +385,7 @@
                 </DoubleAnimation>
                 <DoubleAnimation Storyboard.TargetName="Thumb"
                                  Storyboard.TargetProperty="Height"
-                                 To="{Binding Tag, ElementName=SwitchGrid}"
+                                 To="{Binding Tag, ElementName=SwitchGrid, FallbackValue=0}"
                                  Duration="0:0:0.16">
                   <DoubleAnimation.EasingFunction>
                     <QuadraticEase EasingMode="EaseOut" />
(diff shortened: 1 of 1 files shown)


nicolaihenriksen:
This does not fix the issue, closing PR

PR #3781, opened 2025-01-30
Fix 3780 - NumericUpDown/DecimalUpDown display does not update with values ​​outside the range

### LINKED ISSUE #3780: NumericUpDown/DecimalUpDown display does not update with values ​​outside the range
### Bug explanation

If you set LostFocus Minimum/Maximum to NumericUpDown/DecimalUpDown and enter a number outside the range multiple times, and then click the LostFocus/IncreaseButton, the Value will be correct but the display will not be corrected to within the range.

ex) Minimum=-2.5, Maximum=2.5 value=0
  1st) Value:0/Input 20 -> lostFocus -> View:2.5/Value:2.5
  2nd) Value:2.5/Input 30 -> lostFocus -> View:30/Value:2.5
  3rd) Value:2.5/Input 30 -> IncreaseButton Click -> View:30/Value:2.5

Could it be because the SetCurrentValue of OnTextBoxFocusLost in UpDownBase.cs is not clamped at Minimum and Maximum?

### Version

5.2.1

corvinsz:
fixes #3780

I think my [comment](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/3780#issuecomment-2622861584) on the issue describe the fix/PR

### DIFF
--- a/src/MainDemo.Wpf/NumericUpDown.xaml
+++ b/src/MainDemo.Wpf/NumericUpDown.xaml
@@ -35,7 +35,11 @@
       <smtx:XamlDisplay Margin="10,5"
                   VerticalAlignment="Top"
                   UniqueKey="decimalUpDown_default">
-        <materialDesign:DecimalUpDown ValueStep="0.5" Minimum="-2.5" Maximum="2.5" />
+        <StackPanel>
+          <materialDesign:DecimalUpDown Value="{Binding MyDecimal, Mode=TwoWay}" ValueStep="0.5" Minimum="-2.5" Maximum="2.5" />
+          <TextBlock Text="{Binding MyDecimal, Mode=TwoWay}" />
+          <Button Content="a button to focus"/>
+        </StackPanel>
       </smtx:XamlDisplay>
 
       <smtx:XamlDisplay Margin="10,5"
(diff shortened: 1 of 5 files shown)


Keboo (approved):
Thank you for the fix here @corvinsz.
I added some UI tests as well.

corvinsz:
@Keboo I called `CoerceNumericValue` explicitly in this PR, because otherwise my value in the Viewmodel can go out of sync with the displayed value in the control.
This might not be the best fix, but I think the current behavior is not 100% desired either.

# Comparison
This example uses a `Maximum` of `2.5`.

Here is the current (main branch) behavior:

NumericUpDownPrev

The displayed value (`2,5`) does not match the bound value in my VM (`555`).

When calling `CoerceNumericValue` explicitly the behavior looks like this:

NumericUpDownPost


I think this has to do with the `OnValueChanged` not being invoked because the value is essentially not changing, it goes from `2.5` to `2.5` (which got "clamped down" from `555`)?

Let me know if I should create a separate issue for this.

Keboo:
@corvinsz thanks for the heads up. Yes, I think logging a new thing for it would be good. In general the DependencyProperty itself should be sending values through the coerce. I think making a UI test with the failure would be a good so that we can then hash on the problem. Let me know if you need any help setting this up.

nicolaihenriksen:
@Keboo I think @corvinsz is right. Probably would suffice to call the `Clamp` method before invoking `SetCurrentValue`.

I mentioned this a while back for this control in particular. Something seems buggy in WPF, such that if you bind the `Value` property to your VM and the user enters an out of bounds value, the setter in the VM will be invoked twice. First with the out of bounds value, and then immediately after with the coerced value. By clamping before calling `SetCurrentValue`, we avoid notifying the VM with an invalid (temporary) value.

I believe you will see the same clamping happening in the plus/minus button click handlers.

PR #3803, opened 2025-02-28
Fix Enhance Mouseover Visuals for TabControl

### LINKED ISSUE #3802: Enhance Mouseover Visuals for TabControl
Request: Improve the mouseover visual feedback for the TabControl. When hovering over an unselected tab, I propose the following enhancements:

- Highlight the tab to indicate it is interactive.
- Change the cursor to a pointer (hand icon) to further indicate clickability.

corvinsz:
fixes #3802

- Sets the `Cursor="Hand"` when hovering over a `TabItem`
- Added the same "hover effect" to `TabItem` which any normal `Button` has

## Current behavior
tabControlPre

## New behavior
tabControlPost

### DIFF
--- a/src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.TabControl.xaml
+++ b/src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.TabControl.xaml
@@ -291,7 +291,7 @@
                                Duration="0" />
             </Storyboard>
           </ControlTemplate.Resources>
-          <Grid x:Name="Root" Cursor="Hand">
+          <Grid x:Name="Root" Cursor="{Binding Path=(wpf:TabAssist.TabHeaderCursor), RelativeSource={RelativeSource AncestorType={x:Type TabControl}}}">
             <!-- This is the Header label ColorZone. -->
             <wpf:ColorZone x:Name="ColorZoneHeader"
                            HorizontalAlignment="Stretch"
--- a/src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.TabControl.xaml
+++ b/src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.TabControl.xaml
@@ -291,7 +291,7 @@
                                Duration="0" />
             </Storyboard>
           </ControlTemplate.Resources>
-          <Grid x:Name="Root">
+          <Grid x:Name="Root" Cursor="Hand">
             <!-- This is the Header label ColorZone. -->
             <wpf:ColorZone x:Name="ColorZoneHeader"
                            HorizontalAlignment="Stretch"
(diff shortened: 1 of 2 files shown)


jbooth88:
Adding a highlight to the tab header on mouse-hover makes sense. However changing the mouse cursor does not follow the WPF framework defaults for the for tab control. I also checked winforms and winUI, neither change the mouse cursor when hovering over a tab item (provided default styles are applied).

corvinsz:
I would argue WPF defaults don't matter.
The purpose of this library is to bring Material Design to WPF, and the MD specs suggest what is proposed in this PR (as far as i have read it).
https://m3.material.io/components/tabs/accessibility#1f5832f9-1898-4da1-bfe4-f40c84027f8a

src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.TabControl.xaml
```diff
@@ -291,7 +291,7 @@
                                Duration="0" />
             </Storyboard>
           </ControlTemplate.Resources>
-          <Grid x:Name="Root">
+          <Grid x:Name="Root" Cursor="Hand">
```
Keboo, line 294: Rather than hard code this value here, lets set it in a Style setter that way if consumers want to change it, they can.
See the `MaterialDesignRaisedButton` for an example of doing this.

corvinsz: @Keboo I intentionally placed the `Cursor="Hand"` property at that level. If I set it at the Style-level, the Cursor becomes a Hand when hovering over the content of the TabItem, which is not desired:
tabControlCursor


Maybe adding an attached property (e.g.: `TabControlAssist.TabHeaderCursor`) is the way to go? This way the cursor behaves the correct way, and consumers can change it, if they want.

Keboo: Ah yes, I had not considered that. Yes, let's go with the attached property approach.

corvinsz:
- I added a new attached property called `TabAssist.TabHeaderCursor`
- The new default `Cursor` is the `Hand`
- Consumers can force the old `Cursor` by setting the AP (e.g.: `materialDesign:TabAssist.TabHeaderCursor="Arrow"`)

nicolaihenriksen:
> I would argue WPF defaults don't matter. The purpose of this library is to bring Material Design to WPF, and the MD specs suggest what is proposed in this PR (as far as i have read it). https://m3.material.io/components/tabs/accessibility#1f5832f9-1898-4da1-bfe4-f40c84027f8a

I agree with @corvinsz. And often when I am in doubt when looking at the MD3 specs, I drop back to the MD2 specs where they have the interactive demos for many things. And that clearly shows the hand cursor on hover:
https://m2.material.io/components/tabs#interactive-demo

PR #3808, opened 2025-03-07
Updating tests to use xUnit v3 and MTP

Keboo:


corvinsz:
I think this can *easily* be fixed with a little regex magic?
Replace:
```
await\s+Task\.Delay\s*\(\s*([^)]*?)\s*\)\s*;
```
With
```
await Task.Delay($1, TestContext.Current.CancellationToken);
```
And don't forget to check `Use regular expressions`

Keboo:
Closing in favor of #3838

PR #3812, opened 2025-03-10
Inherit Windows accent color

### LINKED ISSUE #3811: [Feature] BundledTheme option to inherit Windows accent color automatically
`<BundledTheme>` currently provides an `Inherit` option for the `BaseTheme` attribute, which reads the Dark/Light preference set at the OS level. However, the `PrimaryColor` and `SecondaryColor` attributes don't provide such an option.

Add a new value `Inherit` to the `MaterialDesignColors.PrimaryColor` and `MaterialDesignColors.SecondaryColor` enumerations. Using this value will match the accent color configured in Windows Personalization settings.

See the "Accent Color" section on the [Personalize Your Colors in Windows](https://support.microsoft.com/en-us/windows/personalize-your-colors-in-windows-3290d30f-d064-5cfe-6470-2fe9c6533e37) page for details.

corvinsz:
fixes #3811
With this PR consumers can use the newly added `Inherit` member for the `PrimaryColor` and `SecondaryColor` enum.
Just like with `BaseTheme="Inherit"` inherit means "take the OS preferences".

This is largely based on how it is done in [ControlzEx](https://github.com/ControlzEx/ControlzEx/blob/48230bb023c588e1b7eb86ea83f7ddf7d25be735/src/ControlzEx/Theming/WindowsThemeHelper.cs#L53)

Notes:
In the case the value could not be read from the registry, or the conversion (registry value --> C# Color) is somehow failing, the code falls back to `SwatchHelper.Lookup.First().Value`. One alternative would be to throw an error instead, but I don't think such an "irrelevant" feature should throw an exception.

Example:
```diff

<materialDesign:BundledTheme BaseTheme="Inherit"
                             ColorAdjustment="{materialDesign:ColorAdjustment}"
-                            PrimaryColor="DeepPurple"
+                            PrimaryColor="Inherit"
-                            SecondaryColor="Lime" />
+                            SecondaryColor="Inherit" />
```

I also refactored the `BundledTheme` class slightly.

### DIFF
--- a/src/MaterialDesignThemes.Wpf/BundledTheme.cs
+++ b/src/MaterialDesignThemes.Wpf/BundledTheme.cs
@@ -1,4 +1,5 @@
-﻿using MaterialDesignColors;
+﻿using System.Windows.Media;
+using MaterialDesignColors;
 
 namespace MaterialDesignThemes.Wpf;
 
@@ -10,9 +11,8 @@ public class BundledTheme : ResourceDictionary, IMaterialDesignThemeDictionary
         get => _baseTheme;
         set
         {
-            if (_baseTheme != value)
+            if (SetField(ref _baseTheme, value))
             {
-                _baseTheme = value;
                 SetTheme();
             }
         }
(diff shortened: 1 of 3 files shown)


src/MaterialDesignThemes.Wpf/BundledTheme.cs
```diff
@@ -10,9 +11,8 @@ public BaseTheme? BaseTheme
         get => _baseTheme;
         set
         {
-            if (_baseTheme != value)
+            if (SetField(ref _baseTheme, value))
```
Keboo, line 14: This is great, can we also apply it to the other properties below as well?

src/MaterialDesignThemes.Wpf/BundledTheme.cs
```diff
@@ -62,17 +62,35 @@ public ColorAdjustment? ColorAdjustment
...
+                              ? (accentColor.Value ?? SwatchHelper.Lookup.First().Value)
+                              : SwatchHelper.Lookup[(MaterialDesignColor)secondaryColor];
+
+        Theme theme = Theme.Create(baseTheme, colorPrimary, colorSecondary);
+        theme.ColorAdjustment = ColorAdjustment;
+
+        ApplyTheme(theme);
+    }
+
+    protected bool SetField<T>(ref T field, T value)
```
Keboo, line 89: nit:
```suggestion
    protected static bool SetField<T>(ref T field, T value)
```

src/MaterialDesignThemes.Wpf/BundledTheme.cs
corvinsz: Thanks, I guess I overlooked the other properties.

Keboo (approved):

PR #3831, opened 2025-04-15
Fixed issue with margin of SplitButton not being applied as margin correctly

JorisCleVR:
When applying a margin to the SplitButton I got the following unexpected result:
image

With this PR the margins will be applied similar as they would for a normal button.


nicolaihenriksen (approved):
Was a little curious about some edge cases, but it does seem to work as intended for the ones I tested. Nice work.

Keboo (approved):

PR #3841, opened 2025-04-30
DialogHost API revamp

corvinsz:
Initially I "just" wanted to implement a fullscreen dialog, but I asked myself "why stop here?"

Here are a few thoughts/changes to explain this PR:
- To not clutter the existing `Show` methods even more I opted to implement a new class called `DialogOptions`, to have this option-based configuration-style
- `DialogOptions` has alot of `Previous<some property>` properties, to keep it as compatible with the old `Show` methods. The way this works is by storing the corresponding property before setting it to something different, then show the dialog, and then revert all the properties to the previously stored value. While the old and new API *could* be used interchangingly, it is probably a good idea to just stick with one of the two.
- I added a default CloseButton (`PART_CloseButton`) to the template
- I added a new interface `IDialogService`. We already have an abstraction for the Snackbar (`ISnackbarMessageQueue`), but not (yet) for the DialogHost
- Also I replaced alot of code with syntactic sugar or newer syntaxes

## There are 2 things I want to publicly address
1. What other `Show` methods should we add to the `IDialogService`?
2. I for the life of me can't figure out why the dialog is not sizing to fullscreen if I press the maximize button. I have been trying to figure this out for hours now without any luck:
fullscreenDialogProblem

There are still some ToDos left, but I wanted to share this publicly now so any ideas or recommendations can be incorporated.

If all of this is total nonsense, lmk 😆


corvinsz:
I'm closing this for now since this PR is horribly out of date anyway.
Additionally as long as we support the DialogHost Style using the Popup, we can not fully support some features I initially thought of implementing, like a fullscreen bool property.

"_The width and height of the new window can't be larger than 75 percent of the width or height of the screen._"
https://learn.microsoft.com/en-us/dotnet/desktop/wpf/controls/popup?view=netframeworkdesktop-4.8#popup-and-the-visual-tree

PR #3867, opened 2025-06-10
ToggleButton uses EllipseClipConverter

JorisCleVR:
ToggleButton now uses an EllipseClipConverter to create the Ellipse for clipping of button.

This fixes some binding errors like: System.Windows.Data Error: 2 : Cannot find governing FrameworkElement or FrameworkContentElement for target element. BindingExpression:Path=Width; DataItem=null; target element is 'EllipseGeometry' (HashCode=37186555); target property is 'RadiusX' (type 'Double')

### DIFF
--- a/src/MaterialDesignThemes.Wpf/Converters/EllipseClipConverter.cs
+++ b/src/MaterialDesignThemes.Wpf/Converters/EllipseClipConverter.cs
@@ -1,0 +1,41 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows.Data;
+using System.Windows.Media;
+
+namespace MaterialDesignThemes.Wpf.Converters
+{
+    public class EllipseClipConverter : IMultiValueConverter
+    {
+        public static readonly EllipseClipConverter Instance = new();
+
+        public object Convert(object[] values, Type targetType, object parameter, CultureInfo culture)
+        {
+            if (values is [double width, double height, ..])
+            {
+                if (width < 1.0 || height < 1.0)
+                {
+                    return Geometry.Empty;
+                }
+
+                Point center = new Point(width / 2.0, height / 2.0);
+                double radiusX = width / 2.0;
+                double radiusY = height / 2.0;
+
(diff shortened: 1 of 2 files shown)


src/MaterialDesignThemes.Wpf/Converters/EllipseClipConverter.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows.Data;
+using System.Windows.Media;
+
+namespace MaterialDesignThemes.Wpf.Converters
```
Keboo, line 10: nit: We have been favoring file-scoped namespaces

Keboo (approved):

src/MaterialDesignThemes.Wpf/Converters/EllipseClipConverter.cs
JorisCleVR: Indeed good to keep consistency in the code base: Fixed

corvinsz:
Not that we need it, and I'm just asking out of curiosity:
Do we in general not add tests for the converters?
Currently there are tests for only like 5 converters.

At work we always add tests for every converter to (obviously) test it, and also to show intent of what the converter is supposed to do. Sometimes it could be hard to fit a converters purpose into it's name. Especially because converters are this well encapsulated/standalone "component" it seems like a nobrainer to add tests for it.

nicolaihenriksen:
@corvinsz I don't believe we have a rule/guideline stating that you must or must not add tests for them. Personally I have added a few tests when the logic grows to be non-trivial.

However, I like your comment about the tests indicating intent, because naming is hard! and I have often found myself scratching my head looking a converter name trying to decipher what its purpose in life is.

So in short, I think you're more than welcome to add tests, especially fast running "unit tests". The more safety net, the better IMO.

corvinsz:
Neat. After @Keboo has successfully converted and merged all the tests to TUnit - and I can spare some time - I can take a look at adding tests for converters.

Keboo (approved):

Issue #3814, opened 2025-03-13
XAML Binding Failures when ProgressBar is styled with MaterialDesignCircularProgressBar in v5.2.1 (Was working in v5.1.0)


shochinhwa:
### Problem
Encountered multiple binding failures in XAML when ProgressBar is styled with MaterialDesignCircularProgressBar. There was no binding failures previously in v5.1.0 and it appears only after I upgraded it to the latest v5.2.1.

### Steps

1. Create a clean Example solution.
```
<Application
    x:Class="Example.App"
    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
    xmlns:materialDesign="http://materialdesigninxaml.net/winfx/xaml/themes"
    StartupUri="MainWindow.xaml">
    <Application.Resources>
        <ResourceDictionary>
            <ResourceDictionary.MergedDictionaries>
                <materialDesign:BundledTheme
                    BaseTheme="Light"
                    PrimaryColor="DeepPurple"
                    SecondaryColor="Lime" />
                <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesign3.Defaults.xaml" />
            </ResourceDictionary.MergedDictionaries>
        </ResourceDictionary>
    </Application.Resources>
</Application>
```

2. Add a ProgressBar with MaterialDesignCircularProgressBar style.

```
<Window
    x:Class="Example.MainWindow"
    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
    xmlns:circularProgressBar="clr-namespace:MaterialDesignThemes.Wpf.Converters.CircularProgressBar;assembly=MaterialDesignThemes.Wpf"
    xmlns:materialDesign="http://materialdesigninxaml.net/winfx/xaml/themes"
    Title="MainWindow"
    Width="800"
    Height="450"
    Style="{StaticResource MaterialDesignWindow}"
    mc:Ignorable="d">
    <Grid>
        <ProgressBar
            Width="50"
            Height="50"
            IsIndeterminate="True"
[...]

PR #3868, opened 2025-06-11
Used x:Reference in Button and ProgressBar ControlTemplates

JorisCleVR:
Used x:Reference in Button and ProgressBar ControlTemplates to fix some binding errors.

The issue is that ElementName does a Runtime lookup of the element. At some moments in time or on freezable objects this does not work, because they are not available in the visual tree. Therefore within a ControlTemplate x:Reference can be used to do a compile time / load time lookup.

Note that because the lookup is compile time / load time this also has a small performance gain during load of the element.


corvinsz:
This indeed fixes the binding failures.
Just for completeness: fixes #3814

PR #3873, opened 2025-06-17
Add AP to control TabControl Headers behavior

### LINKED ISSUE #3863: TabItem in two rows
### Bug explanation

When using TabPanel, the TabControl elements are not displayed in two rows when the window is scaled down, as it works in a standard WPF window.

Standart WPF window

Image

MaterialDesignInXaml

Image

### Version

5.2.1

corvinsz:
fixes #3863

This PR adds a new attached property (`materialDesign:TabAssist.HeaderBehavior`) to let consumers decide on how to display the headers of a `TabControl`.

Currently the headers are always contained inside of a ScrollViewer, which is not always ideal. Under certain circumstances the user is unable to click on "the next" tab, because it is simply not shown:

455977593-05159782-e2c0-48dc-96f7-40ac3e8f9c08

With the new AP TabControls can be forced to wrap their headers in a `WrapPanel`, e.g.: `materialDesign:TabAssist.HeaderBehavior="Wrapping"`.
There is basically now a trigger in the ControlTemplate which swaps out the header of the TabControl.

455977937-18020632-72d4-45b5-a850-e11e186f5825

## Note
As of right now the AP can't be changed at runtime. Apparently applying the template when the property changes isn't enough. The TabControl sometimes doesn't show it's content when selecting a Tab after changing the `materialDesign:TabAssist.HeaderBehavior`. I would argue having the need of changing this AP at runtime is rather rare.
```csharp
private static void OnHeaderBehaviorChanged(DependencyObject d, DependencyPropertyChangedEventArgs e)
{
    if (d is TabControl tabControl)
    {
        tabControl.ApplyTemplate();
    }
}
```

### DIFF
--- a/src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.TabControl.xaml
+++ b/src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.TabControl.xaml
@@ -31,57 +31,83 @@
         <ControlTemplate TargetType="{x:Type TabControl}">
           <DockPanel KeyboardNavigation.TabNavigation="Local">
             <wpf:ColorZone x:Name="PART_HeaderZone"
-                           VerticalAlignment="Stretch"
-                           Panel.ZIndex="1"
-                           wpf:ElevationAssist.Elevation="{TemplateBinding wpf:ElevationAssist.Elevation}"
-                           Background="{TemplateBinding wpf:ColorZoneAssist.Background}"
-                           DockPanel.Dock="Top"
-                           Focusable="False">
-              <ScrollViewer wpf:ScrollViewerAssist.BubbleVerticalScroll="True"
-                            wpf:ScrollViewerAssist.SupportHorizontalScroll="True"
+                         VerticalAlignment="Stretch"
+                         Panel.ZIndex="1"
+                         wpf:ElevationAssist.Elevation="{TemplateBinding wpf:ElevationAssist.Elevation}"
+                         Background="{TemplateBinding wpf:ColorZoneAssist.Background}"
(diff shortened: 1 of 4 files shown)


Keboo:
Soerry for the delay on this. I think a better implementation would be to create a secondary style and/or control template for the TabControl that implements it with the WrapPanel. The default WPF behavior contradicts the material design behavior, so we want the current (horizontal scroll behavior) to be the default, but in cases like this we have often put in secondary styles that people can set if they want a different behavior.

src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.TabControl.xaml
```diff
@@ -31,57 +31,83 @@
...
-                           Focusable="False">
-              <ScrollViewer wpf:ScrollViewerAssist.BubbleVerticalScroll="True"
-                            wpf:ScrollViewerAssist.SupportHorizontalScroll="True"
+                         VerticalAlignment="Stretch"
+                         Panel.ZIndex="1"
+                         wpf:ElevationAssist.Elevation="{TemplateBinding wpf:ElevationAssist.Elevation}"
+                         Background="{TemplateBinding wpf:ColorZoneAssist.Background}"
+                         DockPanel.Dock="Top"
+                         Focusable="False">
+              <Grid>
```
Keboo, line 40: Rather than this grid with both sets of controls. What if we apply a style trigger to the ColorZone and swap out its child content based on the AP?

corvinsz:
I'm going to create a new PR for this. The rebase was a little much

PR #3874, opened 2025-06-18
Evaluate TextBox.LineCount only once in UpdateAttachedProperties

greuelpirat:
I have a performance issue in `TextBox.OnTextChanged` handler in Version 5.2.1 with poor responsiveness. I tracked it down to TextBoxLineCountBehavior:

image

There has already been an improvement to queue the update with `Background`-priority. That is good.

But I also noticed, that `TextBox.LineCount` is evaluated twice. This PR reduces the calls to one, so it improves background performance even more.


Keboo (approved):
Awesome work on the analysis. Thank you.

PR #3877, opened 2025-06-23
implement TreeDataGrid

MohammadHadi2031:
Hi Keboo.

According to talk in the discord about adding columns to `TreeListView` I created this pull request. This is WIP and most of the task is done just the styles is left (light/dark theme, ...). for using this just run demo app and go to trees -> multi select tree view I have couple of questions:
1. Shall we show gripper (gridlines) for columns? in material design theme the data table column's gripper has no color.
2. In order to show columns for cells I used `GridViewRowPresenter` but I can't remove `internal:TreeListViewContentPresenter` because no children is shown. so the final code is as below. Is there better solution for this?
```
 <Grid Margin="0,0,0,0">
   <internal:TreeListViewContentPresenter x:Name="PART_ContentPresenter"
                                          Content="{TemplateBinding Content}"
                                          Visibility="Hidden" />

   <GridViewRowPresenter x:Name="PART_Header"
                         Columns="{Binding Columns, RelativeSource={RelativeSource AncestorType={x:Type wpf:TreeListView}}}"
                         Content="{TemplateBinding Content}" />
 </Grid>
```
3. For showing gridlines for each treelistview cells I created simple `TreeListViewCell` if it's not necessary to show gridlines we can remove this. 
```
 <Style TargetType="{x:Type wpf:TreeListViewCell}">
   <Setter Property="Template">
     <Setter.Value>
       <ControlTemplate TargetType="{x:Type wpf:TreeListViewCell}">
         <Border Background="{TemplateBinding Background}"
                 BorderBrush="{TemplateBinding BorderBrush}"
                 BorderThickness="{TemplateBinding BorderThickness}">
           <Grid>
             <Border Margin="0,0,-5,0"
                     Padding="0,3,0,3"
                     HorizontalAlignment="Stretch"
                     VerticalAlignment="Stretch"
                     BorderBrush="{Binding RelativeSource={RelativeSource AncestorType={x:Type wpf:TreeListViewItem}}, Path=IsSelected, Converter={StaticResource TreeListViewGridLineBrushConverter}}"
                     BorderThickness="0,0,1,0">
               <TextBlock Text="" />
             </Border>

             <ContentPresenter />
           </Grid>
         </Border>
       </ControlTemplate>
     </Setter.Value>
   </Setter>
 </Style>
```

4. I moved Toggle button to the first column data template this is a easy trick but maybe not perfect solution I want to ask you what could be better way to have toggle button on the first cell. here is the first column

```
<GridViewColumn Width="250"
                Header="Name">
  <GridViewColumn.CellTemplate>
    <DataTemplate>
      <materialDesign:TreeListViewCell>
          <StackPanel.Margin>
            <MultiBinding Converter="{StaticResource TreeListViewIndentConverter}">
              <Binding Path="LevelIndentSize"
                       RelativeSource="{RelativeSource FindAncestor, AncestorType={x:Type materialDesign:TreeListView}}" />
              <Binding Path="Level"
                       RelativeSource="{RelativeSource AncestorType={x:Type materialDesign:TreeListViewItem}}" />
            </MultiBinding>
          </StackPanel.Margin>

          <ToggleButton IsChecked="{Binding RelativeSource={RelativeSource AncestorType=materialDesign:TreeListViewItem, Mode=FindAncestor}, Path=IsExpanded}"
                        Style="{StaticResource MaterialDesignTreeListViewToggleButtonStyle}" />

          <TextBlock VerticalAlignment="Center"
                     Text="{Binding Name, Mode=OneTime}" />
        </StackPanel>
      </materialDesign:TreeListViewCell>
    </DataTemplate>
  </GridViewColumn.CellTemplate>
</GridViewColumn>
```

5. Can I create new control called `TreeDataGrid`?

feel free to mention me if there is any error. here is my view
image

### DIFF
--- a/tests/MaterialDesignThemes.UITests/WPF/TimePickers/TimePickerTests.cs
+++ b/tests/MaterialDesignThemes.UITests/WPF/TimePickers/TimePickerTests.cs
@@ -528,12 +528,15 @@ public class TimePickerTests : TestBase
         await timePickerTextBox.SendKeyboardInput($"invalid time");
         await Assert.That(await timePickerTextBox.GetText()).IsEqualTo("invalid time");
 
-        // Act
-        await clearButton.LeftClick();
-        await Task.Delay(50, TestContext.Current!.CancellationToken);
+        await Wait.For(async () =>
+        {
+            // Act
+            await clearButton.LeftClick();
+            await Task.Delay(50, TestContext.Current!.CancellationToken);
 
-        // Assert
-        await Assert.That(await timePickerTextBox.GetText()).IsNull();
+            // Assert
+            await Assert.That(await timePickerTextBox.GetText()).IsNull();
+        });
 
         recorder.Success();
     }


Keboo:
Just replying back to the questions.

1. I think let's skip on the separator line. I think having it similar to the DataGrid is probably most common.
2. This is a bit interesting and will require some more thought. Initially I think it might be fine, but I am interested in testing out some cases with template bindings as that was the intent for in the other presenter.
3. Yea I think we can drop it (as per 1)
4. This is more interesting, I would really want to keep the toggle outside of the columns.
5. I am bit hesitant to create a brand-new control. I would be more open to a new control template over the existing one though. It seems like that might be the cleaner approach.

I think we might need to pivot on the implementation a bit here. It would probably be better to look more at implementing it view the GridView similar to how the WPF base [ListView control works](https://github.com/dotnet/wpf/blob/28843aafcedc8299336ae470c868f842a4ca52b4/src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Fluent/Styles/ListView.xaml#L77). This likely will be a more complicated path, but I think it will be the better approach.

MohammadHadi2031:
Thanks a lot for suggesting `GridView` it got much easier but still toggle button is the issue. So what I did was to see how WPF is working with GridView and I found out that for having columns it uses [style for the ScrollViewer ](https://github.com/dotnet/wpf/blob/28843aafcedc8299336ae470c868f842a4ca52b4/src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Fluent/Styles/GridView.xaml#L211) and for having cells in rows it uses `GridViewRowPresenter` too. So we have it columns and cells. the only thing is to support both `TreeListView.View == null` and `TreeListView.View != null` I've done it with a simple `DataTrigger` just like how WPF does. 

```
<DataTrigger Binding="{Binding View, RelativeSource={RelativeSource AncestorType=wpf:TreeListView}, Converter={StaticResource ViewIsGridViewConverter}}" Value="True">
    <Setter TargetName="PART_ContentPresenter" Property="Visibility" Value="Hidden" />
    <Setter TargetName="PART_GridViewRowPresenterr" Property="Visibility" Value="Visible" />
</DataTrigger>

```

image


The only problem is the `ToggleButton` outside the TreeListViewItem causes tests to fail. I'm working on it

MohammadHadi2031:
Hi. The toggle button is now automatically added to the first column. I commented out the assignment of `TreeListView.View` to make the tests pass.

MohammadHadi2031:
I think we can remove [this comment ](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/blob/fad77d5b0d7c6c89af6f32a358827c944c96c331/src/MainDemo.Wpf/Trees.xaml#L234) after merging #3875

src/MainDemo.Wpf/Trees.xaml
```diff
@@ -231,19 +230,41 @@
...
+             https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/3640#issuecomment-2274086113
+             https://learn.microsoft.com/dotnet/desktop/wpf/advanced/optimizing-performance-controls?view=netframeworkdesktop-4.8&WT.mc_id=DT-MVP-5003472
+             -->
             <materialDesign:TreeListView.ItemContainerStyle>
               <Style TargetType="materialDesign:TreeListViewItem" BasedOn="{StaticResource {x:Type materialDesign:TreeListViewItem}}">
                 <Setter Property="IsExpanded" Value="{Binding IsExpanded}" />
               </Style>
             </materialDesign:TreeListView.ItemContainerStyle>
+
+            <!--<materialDesign:TreeListView.View>
```
Keboo, line 247: For the demo app, it would be great if we just included a second instance of this TreeListView and enabled the columns for it.

src/MaterialDesignThemes.Wpf/TreeListView.cs
```diff
@@ -259,4 +262,68 @@ internal void MoveSelectionToParent(TreeListViewItem item)
...
+    {
+        base.OnPropertyChanged(e);
+
+        if (e.Property == ViewProperty)
+        {
+            if (View is GridView gridView)
+            {
+                AddToggleButtonToFirstColumn(gridView);
+            }
+        }
```
Keboo, line 276:
```suggestion
        if (e.Property == ViewProperty && View is GridView gridView)
        {
            AddToggleButtonToFirstColumn(gridView);
        }
```

src/MaterialDesignThemes.Wpf/TreeListView.cs
```diff
@@ -259,4 +262,68 @@ internal void MoveSelectionToParent(TreeListViewItem item)
...
+            RelativeSource = new RelativeSource(RelativeSourceMode.FindAncestor, typeof(TreeListView), 1)
+        });
+        marginMultiBinding.Bindings.Add(new Binding(TreeListViewItem.LevelProperty.Name)
+        {
+            RelativeSource = new RelativeSource(RelativeSourceMode.FindAncestor, typeof(TreeListViewItem), 1)
+        });
+        stackPanelFactory.SetBinding(StackPanel.MarginProperty, marginMultiBinding);
+
+        var toggleButtonFactory = new FrameworkElementFactory(typeof(ToggleButton));
+        toggleButtonFactory.SetValue(ToggleButton.StyleProperty, Application.Current.Resources["MaterialDesignTreeListViewToggleButtonStyle"]);
```
Keboo, line 310:
```suggestion
        stackPanelFactory.SetBinding(MarginProperty, marginMultiBinding);

        var toggleButtonFactory = new FrameworkElementFactory(typeof(ToggleButton));
        toggleButtonFactory.SetValue(StyleProperty, FindResource("MaterialDesignTreeListViewToggleButtonStyle"));
```

src/MaterialDesignThemes.Wpf/TreeListView.cs
```diff
@@ -259,4 +262,68 @@ internal void MoveSelectionToParent(TreeListViewItem item)
...
+    }
+
+    private DataTemplate CreateToggleButtonTemplate(DataTemplate originalTemplate)
+    {
+        var template = new DataTemplate();
+
+        var stackPanelFactory = new FrameworkElementFactory(typeof(StackPanel));
+        stackPanelFactory.SetValue(StackPanel.OrientationProperty, Orientation.Horizontal);
+
+        var marginMultiBinding = new MultiBinding
```
Keboo, line 295: I think these margins are a bit low. I think we want there to be a larger separation between the toggle button and the content (closer to what it is without the GridView). The height of the individual rows when using the GridView is also a bit larger than without. Ideally we would want these to be pretty close to the same default sizes.

With GridView

Image

Without GridView:
Image

Keboo (commented):
This is looking really good! thank you!

MohammadHadi2031:
Done

JeanPhilippeLux:
Is it possible to test this new control? I took the last version of the code but it's not in the demo app.

thnak you.

MohammadHadi2031:
Yes you can checkout this branch and test it

JeanPhilippeLux:
> Yes you can checkout this branch and test it

strange, I download your branch and when I run the demo, I don't see the new control. Perhaps it's not integrated in the demo application?

thank you.

MohammadHadi2031:
So you `git clone ....` then `git checkout hadi/treedatagrid` then run demo app but It's not there?

> > Yes you can checkout this branch and test it
> 
> strange, I download your branch and when I run the demo, I don't see the new control. Perhaps it's not integrated in the demo application?
> 
> thank you.

JeanPhilippeLux:
> So you `git clone ....` then `git checkout hadi/treedatagrid` then run demo app but It's not there?
> 
> > > Yes you can checkout this branch and test it
> > 
> > 
> > strange, I download your branch and when I run the demo, I don't see the new control. Perhaps it's not integrated in the demo application?
> > thank you.

Yes. I open it in Visual Studio. :(

Will make a new try.

PR #3885, opened 2025-07-16
Fix: Add TemplateBindings to BorderBrush and BorderThickness

### LINKED ISSUE #3883: SnackBar BorderThickness and BorderBrush not working
### Bug explanation

When setting BorderThickness and BorderBrush on a SnackBar, these settings aren't reflected on the template as the border doesn't have these properties template bound.

Suggested solution is to implement the following lines to the SnackBar Style, in the Border object:

BorderBrush="{TemplateBinding BorderBrush}"
BorderThickness="{TemplateBinding BorderThickness}"

### Version

5.2.1

HedpE-:
Add TemplateBindings to BorderBrush and BorderThickness to SnackBar Style.

Fixes #3883

### DIFF
--- a/.github/workflows/get_versions.yml
+++ b/.github/workflows/get_versions.yml
@@ -23,9 +23,9 @@ jobs:
   getting_versions:
     env:
       #Update these base version numbers
-      mdix-version: "5.2.2"
-      mdix-colors-version: "5.2.2"
-      mdix-mahapps-version: "5.2.2"
+      mdix-version: "5.3.0"
+      mdix-colors-version: "5.3.0"
+      mdix-mahapps-version: "5.3.0"
     name: Set version numbers
     runs-on: ubuntu-latest
     defaults:


nicolaihenriksen:
@HedpE- Thank you for the PR. The changes look fine to me! One thought though. In many cases, when I am manipulating the `BorderBrush/BorderThickness` of a bordered element, I also want/need control over the `CornerRadius`. We could consider adding a `SnackBar.CornerRadius` dependeny property of type `Thickness`.

The default style would then need an additional setter:
```xaml
<Setter Property="CornerRadius" Value="3" />
```
And the value set in the `ContentBorder` should be changed to a template binding:
```xaml
<Border x:Name="ContentBorder"
  ...
  CornerRadius="{TemplateBinding CornerRadius}"
  ...
</Border>
```

HedpE-:
Yes I agree, it's a fiine addition, should I do another PR for the CornerRadius?

nicolaihenriksen:
> Yes I agree, it's a fiine addition, should I do another PR for the CornerRadius?

Adding it here as you did is fine I believe 👍

PR #3893, opened 2025-07-30
Update README.md

jahansubbirci:


Keboo:
Thank you for this, however, for now I want to keep the README using the bundled theme. This other possible configuration is covered in the demo app

PR #3913, opened 2025-09-03
Fix3904

### LINKED ISSUE #3904: NVDA does not read ComboBox names with complex Headers in MaterialDesignFloatingHintComboBox and MaterialDesignFilledComboBox
### Bug explanation

When using ComboBox controls with MaterialDesignFloatingHintComboBox or MaterialDesignFilledComboBox styles, NVDA does not read the control's name correctly if the Header is complex (e.g., a StackPanel or DockPanel). This makes the controls inaccessible to screen reader users, as NVDA skips or fails to announce a meaningful name.
Steps to Reproduce:
1. Use a ComboBox with MaterialDesignFloatingHintComboBox or MaterialDesignFilledComboBox style.
2. Set a complex Header (e.g., a DockPanel with another ComboBox or a StackPanel with a TextBlock and icon).
3. Set materialDesign:HintAssist.Hint to a descriptive string (e.g., "Routing").
4. Run the app and navigate to the ComboBox with NVDA.
Expected Behavior:
NVDA should announce a meaningful name for the ComboBox, ideally using HintAssist.Hint.
Actual Behavior:
NVDA does not announce a meaningful name, reducing accessibility.
Proposed Solution:
Add a Setter to the MaterialDesignFloatingHintComboBox and MaterialDesignFilledComboBox styles in MaterialDesignTheme.ComboBox.xaml to set AutomationProperties.Name based on HintAssist.Hint:
<Style x:Key="MaterialDesignFloatingHintComboBox" TargetType="{x:Type ComboBox}" BasedOn="{StaticResource MaterialDesignComboBox}">
    <Setter Property="AutomationProperties.Name" Value="{Binding RelativeSource={RelativeSource Self}, Path=(materialDesign:HintAssist.Hint), FallbackValue={x:Null}}" />
    
</Style>
<Style x:Key="MaterialDesignFilledComboBox" TargetType="{x:Type ComboBox}" BasedOn="{StaticResource MaterialDesignComboBox}">
    <Setter Property="AutomationProperties.Name" Value="{Binding RelativeSource={RelativeSource Self}, Path=(materialDesign:HintAssist.Hint), FallbackValue={x:Null}}" />
    
</Style>
Benefits:
• Enhances accessibility for screen reader users.
• Uses existing HintAssist.Hint for consistency.
• No visual changes to the ComboBox.
• Removes need for manual AutomationProperties.Name in XAML.
Code Sample:
<ComboBox
    x:Name="cmbRoutings"
    MaxWidth="300"
    materialDesign:HintAssist.Hint="Routing"
    AutomationProperties.Name="Routing"
    Style="{StaticResource MaterialDesignFilledComboBox}" />

Additional Notes:
• Currently, I must manually set AutomationProperties.Name for each ComboBox, which is repetitive.

### Version

5.2.1

corvinsz:
fixes #3904

This PR sets the default value of the `AutomationProperties.Name` to the `HintAssist.Hint` of the respective input control, so that a screen reader picks up the hint-text by default.
The effected controls are essentially all the input controls:
- `TextBox`
- `PasswordBox`
- `ComboBox`
- `DatePicker`
- `TimePicker`
- `NumericUpDown`
- `AutoSuggestBox`

I successfully tested the changes with the [NVDA](https://www.nvaccess.org/download/) screen reader.

The issuer is talking about supporting "complex headers" too, but do we really want to support something like this?
In my opinion, we shouldn't, see my [comment](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/3904#issuecomment-3225135283).

### DIFF
--- a/src/MainDemo.Wpf/Home.xaml
+++ b/src/MainDemo.Wpf/Home.xaml
@@ -54,11 +54,6 @@
             <TextBlock Margin="8,0,0,0" Text="EXPLORE" />
           </StackPanel>
         </Button>
-
-        <ComboBox materialDesign:HintAssist.Hint="Routing"
-                  Style="{StaticResource MaterialDesignFilledComboBox}"
-                  Width="400"/>
-        
       </StackPanel>
     </Grid>
 
--- a/src/MainDemo.Wpf/Home.xaml
+++ b/src/MainDemo.Wpf/Home.xaml
@@ -54,6 +54,11 @@
             <TextBlock Margin="8,0,0,0" Text="EXPLORE" />
           </StackPanel>
         </Button>
+
+        <ComboBox materialDesign:HintAssist.Hint="Routing"
+                  Style="{StaticResource MaterialDesignFilledComboBox}"
+                  Width="400"/>
+        
       </StackPanel>
     </Grid>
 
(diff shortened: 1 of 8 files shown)


Keboo (approved):

Keboo:
I agree. We can't handle complex headers as there are too many edge cases to try and figure out. As nice as it would be, I think leaving those in the hands of the application author is best.

PR #3915, opened 2025-09-04
Set SmartHint.IsContentNullOrEmpty on initial load

### LINKED ISSUE #3914: TextBox clear button visible even when no text
### Bug explanation

When TextBox visibility is initially set `Collapsed` and `TextFieldAssist.HasClearButton = true` and TextBox visibility changes to `Visible`, clear button is visible inside the TextBox even when there no text. See attached video.

```
<Window x:Class="TestingMaterialDesignXaml.MainWindow"
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
        xmlns:materialDesign="http://materialdesigninxaml.net/winfx/xaml/themes"
        mc:Ignorable="d"
        Title="MainWindow" Height="450" Width="800"
        Style="{StaticResource MaterialDesignWindow}">
        <StackPanel VerticalAlignment="Center">
            <TextBox x:Name="TextBox" 
                     Width="200"
                     Margin="10"
                     Visibility="Collapsed"/>
            <Button Width="200"
                    Content="Hide/Show"
                    Click="ButtonBase_OnClick"/>
        </StackPanel>
</Window>
```

```
using System.Windows;

namespace TestingMaterialDesignXaml
{
    /// <summary>
    /// Interaction logic for MainWindow.xaml
    /// </summary>
    public partial class MainWindow
    {
        public MainWindow()
        {
            InitializeComponent();
        }

        private void ButtonBase_OnClick(object sender, RoutedEventArgs e)
        {
            TextBox.Visibility = TextBox.Visibility == Visibility.Collapsed ? Visibility.Visible : Visibility.Collapsed;
        }
    }
}
```

https://github.com/user-attachments/assets/d40dfcbb-2220-4b4a-80ea-a2923804d29c

### Version

5.2.1

corvinsz:
fixes #3914

As far as I can tell, the root cause of this issue is that the `SmartHint.IsContentNullOrEmpty` is not updated when the visibility changes, and therefore set to `false` (even though it should be `true`).
Setting the `IsContentNullOrEmpty` initially (in the `RefreshState` method) fixes this issue.

485673542-c0fd7ea8-2535-49e3-ab85-10fb743c3e0a

### DIFF
--- a/src/MaterialDesignThemes.Wpf/SmartHint.cs
+++ b/src/MaterialDesignThemes.Wpf/SmartHint.cs
@@ -253,6 +253,8 @@ public class SmartHint : Control
         if (proxy is null) return;
         if (!proxy.IsVisible) return;
 
+        IsContentNullOrEmpty = proxy.IsEmpty();
+
         var action = new Action(() =>
         {
             string state = string.Empty;
--- a/tests/MaterialDesignThemes.UITests/WPF/TextBoxes/TextBoxTests.cs
+++ b/tests/MaterialDesignThemes.UITests/WPF/TextBoxes/TextBoxTests.cs
@@ -585,6 +585,33 @@ public class TextBoxTests : TestBase
 
         recorder.Success();
     }
+
+    [Test]
+    [Description("Issue 3914")]
+    public async Task TextBox_ClearButtonRemainsHidden_WhenInitiallyCollapsedAndMadeVisible()
+    {
+        await using var recorder = new TestRecorder(App);
+
+        var grid = await LoadXaml<Grid>($"""
+            <Grid Margin="30">
+                <TextBox Visibility="Collapsed"
+                         VerticalAlignment="Center"
+                         materialDesign:TextFieldAssist.HasClearButton="True">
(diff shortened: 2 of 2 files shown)


Keboo:
Change looks good. Would you mind adding a UI test for this case as well?

corvinsz:
I added a UI test for this.

DoLovya:
Hello, I think your modification should be placed in the action section. Could you please have this bug fixed by me? I want to join some open-source communities to learn from them.

corvinsz:
I reopened this so the tests run on CI once more. So far I couldn't reproduce the failing test (`OnDatePicker_WithClearButton_ClearsSelectedUncommittedText`) on my machine, and I also don't think it's a flaky test because I have ran the test like 20 times now without it failing once.

PR #3919, opened 2025-09-09
Add hover halo for `CheckBox` and `RadioButton`

corvinsz:
The [MD2](https://m2.material.io/components/checkboxes#behavior) (& MD3) specs suggest there is a hovering-effect for the `CheckBox` and `RadioButton` when hovering directly over the "rectangular" part of the control.
Additionally the cursor should be the hand.

### Before
image

### After
image

(The above screenshots also apply to the `RadioButton`)


nicolaihenriksen:
@corvinsz Looks good. Can you post screengrabs of the light theme as well just for reference?

corvinsz:
@nicolaihenriksen sure, here you go:
image

Keboo (approved):

PR #3920, opened 2025-09-09
Porting material-color-utilities to dotnet #2475

MovGP0:
This is a port of the [material-color-utilities](https://github.com/material-foundation/material-color-utilities) (MCU) to C#.

- This code can be used for automatic MaterialDesign3 color scheme generation from a selected color.
- Unit tests are ported from the dart version of the library
- See [m3.material.io: Color system](https://m3.material.io/styles/color/system/how-the-system-works#75d38b0f-75e2-4e5e-b2a9-b05514c420d1) for details
- Implemented Shims to work with `net462` and `net8.0-windows` targets

### DIFF
--- a/tests/MaterialColorUtilities.Tests/ContrastConstraintTests.cs
+++ b/tests/MaterialColorUtilities.Tests/ContrastConstraintTests.cs
@@ -334,7 +334,7 @@ private static readonly double ContrastTolerance = 0.05;
 
     [Test]
     [DisplayName("scheme_correctness_test: all variants & contrast levels satisfy constraints")]
-    public async Task Scheme_Correctness_All()
+    public void Scheme_Correctness_All()
     {
         var constraints = BuildConstraints();
 


MovGP0:
The library creates colors according to this system:
image

Keboo:
Wow @MovGP0 this looks really cool. It will take me some time to get it reviewed. But thank you for it.

MovGP0:
@Keboo I was trying to keep the implementation as close as possible to the original implementation.

However, this has also has some drawback we should consider:
1. colors are represented as signed 32bit integers, instead of `System.Windows.Color`. While a conversion is easily possible, the question is if we only should use a conversion at the border of the library, or also replace `int` with `Color` for internal usage and break with the original source.
2. the unit tests don't have a 100% coverage and have multiple asserts per test. While this makes it aligned wtih the original source code, I'd consider it bad practice to structure unit tests like this.
3. the `CS0618` (`[Obsolete]`) warning is on purpose, since this is how it's in the original code. The question is if this code should be removed or the warning suppressed. Currently it breaks the build.

Keboo:
> @Keboo I was trying to keep the implementation as close as possible to the original implementation.
> 
> However, this has also has some drawback we should consider:
> 
> 1. colors are represented as signed 32bit integers, instead of `System.Windows.Color`. While a conversion is easily possible, the question is if we only should use a conversion at the border of the library, or also replace `int` with `Color` for internal usage and break with the original source.
> 2. the unit tests don't have a 100% coverage and have multiple asserts per test. While this makes it aligned wtih the original source code, I'd consider it bad practice to structure unit tests like this.
> 3. the `CS0618` (`[Obsolete]`) warning is on purpose, since this is how it's in the original code. The question is if this code should be removed or the warning suppressed. Currently it breaks the build.

1. I think at the border of the library using Color over int will make it much more usable.
2. I am fine with the tests having multiple asserts. Usually, the guidance I use is having tests assert on one "logical" thing, even if that means multiple assert statement. And I am happy to make exceptions in certain cases.
3. I pushed a commit to suppress the warning inline for now.


I think I am going to set this to merge. Since it is not currently being consumed or published I think it will then allow for easier collaboration on it moving forward. I think getting the int -> Color bit will be helpful and samples in the demo app.

PR #3923, opened 2025-09-13
Fix dialog losing focus when changing WindowState

### LINKED ISSUE #3434: DialogHost accessibility issue
### Bug explanation

The DialogHost seems to lost the focus when the window is minimized. When you restore it after, the focus goes to the main view even if the dialog host is always opened.

Steps to reproduce : 
1. Open a DialogHost which contains some controls
2.  Use the tab button to navigate on the DialogHost controls => The focus is on the DialogHost and its controls.
3. Minimize the app with the DialogHost opened
4. Maximize the app -> The DialogHost is always opened
5. Use the tab button again to get the focus => The DialogHost has not the focus, it is on the main View.
6. Close the DialogHost and reopened it
7. Use the tab button => The focus is on the DialogHost again

### Version

4.9.0

corvinsz:
fixes #3434 

The fix is basically to listen for the `StateChanged` event of the parent `Window`.
If the window is restored from the minimized state, we focus the previously focused element, which got stored in a field `_lastFocusedDialogElement` when the window got minimized.

I want to point out the artificial delay I added, that we have in other places in the DialogHost class too. I just couldn't get it to work without the (50ms) delay, as described in my comment above the "hacky" code.
```csharp
...
// Kinda hacky, but without a delay the focus doesn't always get set correctly because the Focus() method fires too early
Task.Delay(50).ContinueWith(_ => this.Dispatcher.BeginInvoke(DispatcherPriority.Background, new Action(() =>
...
```

---
<a href="https://gitkraken.dev/review/github/MaterialDesignInXAML/MaterialDesignInXamlToolkit/pull/3923">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://gitkraken.dev/images/figures/gitkraken-review-badge-dark.svg">
    Open with GitKraken
  </picture>
</a>

### DIFF
--- a/src/MaterialDesignThemes.Wpf/DialogHost.cs
+++ b/src/MaterialDesignThemes.Wpf/DialogHost.cs
@@ -64,6 +64,8 @@ public class DialogHost : ContentControl
     private DialogClosingEventHandler? _attachedDialogClosingEventHandler;
     private DialogClosedEventHandler? _attachedDialogClosedEventHandler;
     private IInputElement? _restoreFocusDialogClose;
+    private IInputElement? _lastFocusedDialogElement;
+    private WindowState _previousWindowState;
     private Action? _currentSnackbarMessageQueueUnPauseAction;
 
     static DialogHost()
@@ -370,6 +372,7 @@ public class DialogHost : ContentControl
 
         dialogHost.CurrentSession = new DialogSession(dialogHost);
         var window = Window.GetWindow(dialogHost);
+        dialogHost.ListenForWindowStateChanged(window);
         if (!dialogHost.IsRestoreFocusDisabled)
         {
             dialogHost._restoreFocusDialogClose = window != null ? FocusManager.GetFocusedElement(window) : null;
@@ -395,7 +398,8 @@ public class DialogHost : ContentControl
 
         //https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/187
         //totally not happy about this, but on immediate validation we can get some weird looking stuff...give WPF a kick to refresh...
-        Task.Delay(300).ContinueWith(t => dialogHost.Dispatcher.BeginInvoke(DispatcherPriority.Background, new Action(() => {
+        Task.Delay(300).ContinueWith(t => dialogHost.Dispatcher.BeginInvoke(DispatcherPriority.Background, new Action(() =>
+        {
             CommandManager.InvalidateRequerySuggested();
             //Delay focusing the popup until after the animation has some time, Issue #2912
             UIElement? child = dialogHost.FocusPopup();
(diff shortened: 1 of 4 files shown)


corvinsz:
It seems like there is a problem with running the tests via github actions. This PR just had another failing test (`SplitButton_WithButtonInPopup_CanBeInvoked`). When reran, the test passed.
I'm not sure if this is a TUnit or github actions problem?

I had the same issue in #3915

src/MaterialDesignThemes.Wpf/DialogHost.cs
```diff
@@ -405,6 +409,50 @@ private static void IsOpenPropertyChangedCallback(DependencyObject dependencyObj
         })));
     }
 
+    
+    private void ListenForWindowStateChanged(Window? window)
+    {
+        window ??= Window.GetWindow(this);
```
Keboo, line 415: Not sure we need the `??=` as the only caller used this same call to try and get the window, so if that is null, this will very likely return null a second time.
```suggestion
```

src/MaterialDesignThemes.Wpf/DialogHost.cs
```diff
@@ -405,6 +409,50 @@ private static void IsOpenPropertyChangedCallback(DependencyObject dependencyObj
...
+
+        if (window is not null)
+        {
+            window.StateChanged += Window_StateChanged;
```
Keboo, line 419: We want to make sure we don't register twice.
```suggestion
            window.StateChanged -= Window_StateChanged;
            window.StateChanged += Window_StateChanged;
```

src/MaterialDesignThemes.Wpf/DialogHost.cs
```diff
@@ -405,6 +409,50 @@ private static void IsOpenPropertyChangedCallback(DependencyObject dependencyObj
...
+        // We only need to focus anything manually if the window changes state from Minimized --> (Normal or Maximized)
+        // Going from Normal --> Maximized (and vice versa) is fine since the focus is already kept correctly
+        if (IsWindowRestoredFromMinimized() && IsLastFocusedDialogElementFocusable())
+        {
+            // Kinda hacky, but without a delay the focus doesn't always get set correctly because the Focus() method fires too early
+            Task.Delay(50).ContinueWith(_ => this.Dispatcher.BeginInvoke(DispatcherPriority.Background, new Action(() =>
+            {
+                _lastFocusedDialogElement!.Focus();
+            })));
+        }
```
Keboo, line 447: Because Focusable and Visible can change I would move this check inside of the Task.Delay.


```suggestion
        if (IsWindowRestoredFromMinimized())
        {
            // Kinda hacky, but without a delay the focus doesn't always get set correctly because the Focus() method fires too early
            Task.Delay(50).ContinueWith(_ => this.Dispatcher.BeginInvoke(DispatcherPriority.Background, new Action(() =>
            {
                if (IsLastFocusedDialogElementFocusable())
                {
                    _lastFocusedDialogElement!.Focus();
                }
            })));
        }
```

corvinsz (commented):
Discussion on how to get the last focused element

corvinsz:
I also adjusted the test to test against both DialogHost styles. I would say this problem is worth testing for both styles.

Keboo:
So the issue with the test should be resolved if you rebase on the latest main. I believe I fixed that test last week.

On the note of focus, I think we want to use the Keyboard Focused element rather than the logically focused element. The problem with the current events, is that Windows will have removed the keyboard focused element by the time those events are raised. After a little testing I _think_ that we can watch one of the WND_PROC messages to know when a minimize is about to occur, and capture the element with keyboard focus then.

```csharp
private void ListenForWindowStateChanged(Window? window)
{
    if (window is not null)
    {
        if (PresentationSource.FromVisual(window) is HwndSource source)
        {
            HwndSourceHook hook = Hook;
            source.AddHook(hook);
            //TODO: Need to unregister hooks as well, and ensure we don't double register
        }

        IntPtr Hook(IntPtr hwnd, int msg, IntPtr wParam, IntPtr lParam, ref bool handled)
        {
            //https://learn.microsoft.com/en-us/windows/win32/menurc/wm-syscommand
            const int WM_SYSCOMMAND = 0x0112;
            const int SC_MINIMIZE = 0xf020;
            switch (msg)
            {
                case WM_SYSCOMMAND:
                    if (wParam.ToInt64() == SC_MINIMIZE && //Minimize
                        _popupContentControl?.IsKeyboardFocusWithin == true) //Only persistent the one with keyboard focus
                    {
                        var element = Keyboard.FocusedElement;
                        //TODO: Perist keyboard focused element
                    }
                    break;
            }
            return IntPtr.Zero;
        }
    }
}
```

This code still has the TODOs that would need to be addressed, but this way we can rely on looking at keyboard focus to see which element we care about.

corvinsz:
I implemented the changes as you suggested. From my manual testing it works fine, however the UI test I added is failing.
I haven't yet figured out why it fails.
grafik


**Edit:** I'm starting to think you can't properly debug the tests. If I place a breakpoint in the [switch statement](https://github.com/corvinsz/MaterialDesignInXamlToolkit/blob/34082b4928b5955fcd150b7f51f2985072b8bd83/src/MaterialDesignThemes.Wpf/DialogHost.cs#L438-L439) inside of the `Hook` method, it is never hit. When testing manually it is being hit.

**Edit 2:** To my understanding when setting the `WindowState` programmatically WPF does not send a `WM_SYSCOMMAND` message, but rather it directly calls the Win32 API internally. A hybrid approach (listening for `WM_SYSCOMMAND` and `Window.StateChanged` event) might be the solution here, but I have yet to figure out a working way of doing so.

PR #3935, opened 2025-09-29
Fix White rectangle when disabling Calendar

### LINKED ISSUE #3932: White rectangle when disabling `Calendar`
### Bug explanation

When setting `IsEnabled="False"` on a `Calendar`, a white rectangle is shown:

Image

The red border is due to Snoop, ignore that 😄 

### Version

latest main (5.3.0-ci1132)

corvinsz:
fixes #3932

# Visual breaking change!

The `IsEnabled` of all childs of the Calendar was already inherited correctly.
I added a visual state to grey out the `MaterialDesignCalendarDayButton` to indicate the disabled state.

I removed the rectangle (`PART_DisabledVisual`) which was previously overlaying the Calendar when it was disabled.

## Notes
The MD2 & 3 specs do not specify looks for a disabled calendar, but I find the current behavior very unsuitable.
When no official spec is given, my next best bet is always to get inspired by other UI libraries. 
In React MUI it's done like one would expect, simply grey out all of the Buttons of the Calendar.
https://mui.com/x/react-date-pickers/date-calendar/#form-props

## Before
495171361-0a8af7c2-7cb9-4de4-ac83-26fe57c8aa81

## After
495171386-41708ff0-f5ab-45b4-ba73-c42db1231ec3

### DIFF
--- a/src/MainDemo.Wpf/Pickers.xaml
+++ b/src/MainDemo.Wpf/Pickers.xaml
@@ -441,7 +441,7 @@
                    Margin="0 0 0 8" />
 
         <smtx:XamlDisplay UniqueKey="calendar_1">
-          <Calendar />
+          <Calendar IsEnabled="{Binding DataContext.ControlsEnabled, RelativeSource={RelativeSource FindAncestor, AncestorType=Window}}" />
         </smtx:XamlDisplay>
       </StackPanel>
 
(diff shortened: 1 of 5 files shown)


Keboo:
One small thing, I think we want to also change the opacity on the year text as well.
image

corvinsz:
Done!

PR #3949, opened 2025-10-11
Implementing conversions from int to System.Windows.Media.Color for MaterialColorUtilities

MovGP0:
- Implemented conversion methods for colors given as `int` to `System.Windows.Media.Color`
- Added unit tests for the conversion methods and for showcasing schema generation.

**Notes:**
- `System.Windows.Media.Color` is not available when targeting Non-WPF (WinForms, MAUI, MacOS, Linux, etc.) plattforms. E.g. the following is not supported:
    - MAUI (`Microsoft.Maui.Graphics.Color`)
    - WinForms (`System.Drawing.Color`)
    - [Splat](https://github.com/reactiveui/splat) (`Splat.SplatColor`)
- The current methods do not easily support the scenario that the user wants to pick secondary/tertiary/error/panel colors manually. We may want to implement custom factories for this purpose.
- I've added a dependency to [Shouldly](https://docs.shouldly.org/) to implement multi-asserts properly

### DIFF
--- a/src/MaterialDesign3.MaterialColorUtilities/DynamicColor/DynamicColor.cs
+++ b/src/MaterialDesign3.MaterialColorUtilities/DynamicColor/DynamicColor.cs
@@ -140,10 +140,8 @@ public sealed class DynamicColor
         return (argb & 0x00ffffff) | (alpha << 24);
     }
 
-#if WPF
     public System.Windows.Media.Color GetColor(DynamicScheme scheme)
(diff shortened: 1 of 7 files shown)


MovGP0:
ℹ️ The follwoing changes where ammended:
- Added a factory method for creating dynamic color schemes with colors that can be overriden by the user

Keboo (approved):

MovGP0:
ℹ️ The follwoing changes where ammended:

- removed conditional WPF checks.

PR #3957, opened 2025-10-31
Add null assertion in BundledThemeTests

BenjaminMichaelis:
Added a null assertion to the bundled theme test.


BenjaminMichaelis:
Just proving that the current logger is not actually being used - you can remove the whole thing and summary doesn't change as shown in another pr.

Issue #3959, opened 2025-10-31
Better chip background color in dark theme


database64128:
**Is your feature request related to a problem? Please describe.**

With the light theme, the chips look fine:

<img width="1707" height="1528" alt="chips with light theme" src="https://github.com/user-attachments/assets/eb5180a1-bd6d-41f5-806a-7151424de4d1" />

With the dark theme, non-outlined chips have this weird blueish background:

<img width="1707" height="1528" alt="chips with dark theme" src="https://github.com/user-attachments/assets/06fadda7-c6b0-47a1-b4c8-9ae9c2e3b24c" />

The background dates back to #189.

**Describe the solution you'd like**

Change `MaterialDesign.Brush.Chip.Background`'s dark theme color from `#FF2E3C43` to `White50`.

<img width="1707" height="1528" alt="chips in dark theme with updated background color" src="https://github.com/user-attachments/assets/6f5a5347-6aa6-46d2-9d56-019202c66344" />

**Describe alternatives you've considered**

MD2 does not appear to have guidelines for the dark variant [0] [1].

MD3 uses a tinted primary color [2].

**Additional context**

[0] https://m2.material.io/components/chips
[1] https://m2.material.io/design/color/dark-theme.html
[2] https://m3.material.io/components/chips/specs

PR #3960, opened 2025-10-31
Improve chip background colors

database64128:
Closes https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/3959


Keboo:
Thank you for the excellent bug writeup and fix

PR #3998, opened 2026-01-13
Introduce hyperlink base style disconnected from enforced font sizes

jbooth88:
This pull request introduces a base style for Hyperlink that lets font size and font weight properties bubble in from the parent in the visual tree. Users are currently limited to select a size from the list of available preset styles. If they want a custom font size they have to override in their app.

I am not sure a unique style key is even necessary for each of the various sizes and weights (body, title, etc..) , as they all appear to share the same values. But removing the hyperlink styles would be a breaking change (and the end result is the same).


jbooth88:
Closed because it included a commit from months ago, will re-upload soon

PR #4001, opened 2026-01-15
Remove apparently pointless Margin necessitating an extra Border to handle hit testing

LyndonGingerich:
Non-rhetorical question: Why would a dialog need margin?

This border was added in a12e00b8da09bdc8ec773c63938cd21e59c9a80d: "Clicking on dialog shadow now dismisses dialog (#2290)". If you remove it and turn on `CloseOnClickAway`, the `Margin` leaves a dead zone 35 pixels wide around the dialog. But why can't we just remove the `Margin` instead of adding an extra click handler behind it?

### DIFF
--- a/src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.DialogHost.xaml
+++ b/src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.DialogHost.xaml
@@ -155,13 +155,6 @@
                          PlacementTarget="{Binding ElementName=DialogHostRoot, Mode=OneWay}"
                          Style="{TemplateBinding PopupStyle}">
               <Grid>
-                <Border Background="Transparent" IsHitTestVisible="{TemplateBinding CloseOnClickAway}">
-                  <Border.InputBindings>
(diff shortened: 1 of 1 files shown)


LyndonGingerich:
Considering that there is a specific `DialogMargin` property, there is probably a good reason for it. I'll be quite content to learn that reason.

LyndonGingerich:
Oh that's a user-visible property. Never mind.

I would still be grateful to know what people use this for, particularly since it is set to 35 by default rather than 0.

Keboo:
Hi @LyndonGingerich 
The reason for this comes because we have two different DialogHost styles, the default which puts the dialog in a popup control and an embedded style that simply puts the dialog in a card within the same viaaul tree. The advantage of the popup is that it allows the popup to exceed the size of the parent container, however there are times where it is more advantageous to have the dialog embedded within its parent.

The margin here then means something slightly different. For the popup case it is often less impactful, though it does prevent the dialog from being shoved flush to the edge of the screen. For the embedded one, it is there 

As for the default value of 35, this actually pre-dates my time on the library so I am not actually sure why that value was selected. The MD3 spec [lists 56dp as the recommendation](https://m3.material.io/components/dialogs/guidelines), but that seems a little too large.

PR #4011, opened 2026-02-20
Add support for various properties to PopupBox

### LINKED ISSUE #4010: Visual PopupBox enhancement
### Bug explanation

Currently the `materialDesign:PopupBox`:
- does not have any hover effect
- ripple effect extends beyond the control
- does not support setting any of the following properties:
  - `Background`
  - `BorderBrush`
  - `BorderThickness`
  - `materialDesign:ButtonAssist.CornerRadius`

https://github.com/user-attachments/assets/77529016-8943-4e2d-89d9-a144c4706af5

### Version

5.3.0

corvinsz:
fixes #4010

fixes everything mentioned in the linked issue.
The changes are largely based on how it's done in the other button styles.

https://github.com/user-attachments/assets/48e1a531-dd78-4160-a07d-e0f63d2cb03e

### DIFF
--- a/tests/MaterialDesignThemes.UITests/Samples/PopupBox/PopupBoxWithTemplateSelector.xaml
+++ b/tests/MaterialDesignThemes.UITests/Samples/PopupBox/PopupBoxWithTemplateSelector.xaml
@@ -8,11 +8,11 @@
              d:DesignHeight="450" d:DesignWidth="800">
   <UserControl.Resources>
     <DataTemplate x:Key="OddTemplate">
-      <Border Background="Red" Width="10" Height="10" />
+      <Border x:Name="ContentBorder" Background="Red" Width="10" Height="10" />
     </DataTemplate>
 
     <DataTemplate x:Key="EvenTemplate">
-      <Border Background="Blue" Width="10" Height="10" />
+      <Border x:Name="ContentBorder" Background="Blue" Width="10" Height="10" />
     </DataTemplate>
 
     <local:ColorTemplateSelector x:Key="ColorTemplateSelector"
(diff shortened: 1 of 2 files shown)


corvinsz:
The test `PopupBox_WithContentTemplateSelector_ChangesContent` is failing even on the `main` branch, so I doubt it has anything to do with the changes I have done?

For some reason the background color is `null`.
Could this possibly be a XAMLTest issue?
image

nicolaihenriksen:
> The test PopupBox_WithContentTemplateSelector_ChangesContent is failing even on the main branch

@corvinsz I doubt that it fails on the main branch, because that would fail the pipeline. However, note that the `await Wait.For()` is there for resilience (indicating that the assertion may fail a number of times before passing). So simply stepping through with the debugger attached, could give you some wrong results. In this particular case, it is probably also very much timing related as I assume the `await button.LeftClick()` is intended to open the `Popup` which requires a little time/animation.

Try to set a break point on the `await Wait.For()` line and ensure the "attach debugger to remote process" is appropriately set allowing you to interact with the test window. Then use Snoop to dig out the actual value when the popup is open.

nicolaihenriksen:
@corvinsz Looking at your changes, I believe the issue is that the test simply looks up a border inside the `Popup` using this code:
```xaml
var border = await popupBox.GetElement<Border>();
```

And since you have introduced a `Border` element further up in the visual tree (in the control template), it now grabs the wrong border. Perhaps the best approach is to appropriately name the `Border` in `PopupBoxWithTemplateSelector` and then look it up by name instead of just grabbing the first border it finds.

corvinsz:
@nicolaihenriksen I see, thank you for the detailed explanation. I was not aware of the fact that you can't easily debug the `await Wait.For()`.
I went with your proposed solution to just reference the `Border` by name.

Keboo (approved):

PR #4015, opened 2026-03-05
ScrollViewerAssist: Avoid TargetInvocationException on WinForms application shutdown

### LINKED ISSUE #3439: Using a TabControl triggers a TargetInvocationException at App shutdown, when the WPF Control is hosted in a Winform Elementhost
### Bug explanation

To reproduce this bug, create a Winform app (tested in .NET 8), add an Elementhost (add from code behind since it is not in the Toolbox anymore...) and host a WPF UserControl which contains a TabControl. This will display fine until you close the app. During shutdown, the TargetInvocationException will be thrown. The call stack is hardly helpful since these are shutdown calls. Other controls are apparently not affected and work fine.

Setting the TabControl Style to x:Null will not trigger the exception anymore, so it is definitely something in the TabControl style.

The "source" object of the exception being found in the WeakEventManager.cs seems to be a Scrollviewer whose parent is a "ColorZone". Hope this hint helps.

### Version

4.9.0

nicolaihenriksen:
Fixes #3439 

I have pushed a branch called `reproduceIssue3439` which has a WinForms demo application showcasing the issue and the fix. The last commit on the branch, is the same commit (content) as the fix in this PR. Simply revert that if you want to see the error occurring on application shutdown.

In a use case where a `TabControl` is hosted in an `ElementHost` inside of a WinForms application, the handlers were not removed, and thus an exception was thrown during application shutdown.

This PR registers a `HwndSourceHook` which acts on the WM_DESTROY and WM_NCDESTROY messages in order to clean up the handlers (and unregister the hook to avoid memory leaks)

### DIFF
--- a/src/MaterialDesignThemes.Wpf/ScrollViewerAssist.cs
+++ b/src/MaterialDesignThemes.Wpf/ScrollViewerAssist.cs
@@ -114,6 +114,12 @@ public static class ScrollViewerAssist
             }
         }
 
+        static void RemoveHandlers(ScrollViewer scrollViewer)
+        {
+            WeakEventManager<ScrollViewer, RoutedEventArgs>.RemoveHandler(scrollViewer, nameof(ScrollViewer.Loaded), OnLoaded);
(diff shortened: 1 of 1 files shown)


Keboo (approved):
Reviewed on stream if you are interested. Lgtm.

PR #4023, opened 2026-03-25
Postpone DialogHost.Show() task completion until closing dialog animation has completed

### LINKED ISSUE #4017: DialogHost.Show() completes before closing animation finishes with DialogHost.Close()
**Is your feature request related to a problem? Please describe.**
When closing a dialog using DialogHost.Close(...) (For my the most natural way to close a Dialog if you need do this for code, i don't understand the first way in wiki) after await DialogHost.Show(...), the dialog closing animation can't start if the next line of code blocks UI thread, 1 - 2 ms lines of code after .Close() actually blocks UI Thread and the closing animation doesn't start it feels unnatural until something is implemented that frees the UI thread, a combination of blocking and non-blocking UI Thread can "lag" the close animation. Feels more natural in my opinion just wait the close animation internally.

```cs
await DialogHost.Show(dialog);
await Task.Delay(300); //Wait close animation for continue something, I don't know the exact time of animation but works for now
```

**Describe the solution you'd like**
Ideally, DialogHost.Show(...) should only complete after the dialog has fully closed and the closing animation has finished.

This would allow developers to write code like this without needing manual delays:

In other words, the dialog lifecycle awaited by Show() would include the closing animation timing internally.

**Describe alternatives you've considered**
The current workaround is to manually add a delay after Show() like i said above:

```cs
await DialogHost.Show(dialog);
await Task.Delay(300);
```

Another option is to move work to pool threads with Task.Run.... , but in my opinion, just wait animation closes is more natural.

**Additional context**
No more context i guess.

nicolaihenriksen:
Fixes #4017 

As mentioned in the issue, the `Task` returned by `DialogHost.Show()` completes BEFORE the closing animation has completed. This can result in the dialog "being stuck" on screen in case the UI thread is being used for other work immediately after the `await DialogHost.Show()` call.

This PR attempts to handle this very specific case by simply postponing the task completion with 400ms which ensures the closing animation has completed. To see the issue reproduced, you can simply revert the last commit on the branch.

Currently only creating this as a draft, because I am not sure I like this solution to the problem. But on the other hand I struggle to see any straight-forward alternatives. @Keboo if you have some input, it would be very much appreciated.

Like the OP, I have also had to add a custom delay in some cases after my dialog closes, just to be sure the dialog has disappeared off screen before I do other work.

UPDATE: There is a test failing, which is definitely related. The test attempts to access `dialogHost.IsOpen` after the the dialog closes, and now the test is on a different thread which doesn't own the `dialogHost`. Perhaps this can be mitigated using `STAThreadExecutor`, but it points to something being wrong in my implementation IMO.

Lastly, before this can be merged (if we get to that point), the sample project reproducing the issue needs to be removed again.

### DIFF
--- a/MaterialDesignToolkit.Full.slnx
+++ b/MaterialDesignToolkit.Full.slnx
@@ -6,6 +6,7 @@
     <Platform Name="x86" />
   </Configurations>
   <Folder Name="/Demos/">
+    <Project Path="src/Issue4017/Issue4017.csproj" />
     <Project Path="src/MahMaterialDragablzMashUp/MahAppsDragablzDemo.csproj" />
     <Project Path="src/MainDemo.Wpf/MaterialDesignDemo.csproj" />
     <Project Path="src/MaterialDesign3.Demo.Wpf/MaterialDesign3Demo.csproj" />
--- a/src/Issue4017/App.xaml
+++ b/src/Issue4017/App.xaml
@@ -1,0 +1,18 @@
+﻿<Application x:Class="Issue4017.App"
+             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+             xmlns:local="clr-namespace:Issue4017"
+             xmlns:materialDesign="http://materialdesigninxaml.net/winfx/xaml/themes"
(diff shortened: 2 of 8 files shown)


nicolaihenriksen:
Closing this as I don't like my implementation, and the failing test points to it being wrong.

PR #4034, opened 2026-04-04
Implementing DynamicColorTool which is showcasing dynamic color palette generation

MovGP0:
This PR introduces a new tool that generates dynamic Material 3 color palettes.

Existing code is mostly unchanged, so the existing color logic is uneffected.


Keboo (approved):
Made a few minor tweaks to it. Thank you for all the work!

PR #4042, opened 2026-04-15
Add DP `MatchSizeWith` to `PackIcon`

### LINKED ISSUE #1216: PackIcon styles matching text styles (width/height)
I was thinking that creating styles for PackIcon that makes the icon match in size for the style of texts could remove some of the trial-n-error of creating height, width on each icon.

Suppose I have 
```
    <materialDesign:PackIcon Kind="Warning"/>
    <TextBlock Style="{StaticResource MaterialDesignTitleTextBlock}" Margin="8 0 0 0">STOP!</TextBlock>
</StackPanel>
```
Then to match the size of the textblock we will have to add `Height="27" Width="27"` to the PackIcon.

I imagine the style could simply be named `MaterialDesignTitlePackIcon`. Something like this _might_ work:
```
<Style x:Key="MaterialDesignTitlePackIcon" BasedOn="{StaticResource {x:Type local:PackIcon}}">
	<Setter Property="Height" Value="27"/>
	<Setter Property="Width" Value="27"/>
</Style>
```

corvinsz:
Possibly fixes #1216

This adds a new dependency property `MatchSizeWith` to the `PackIcon`, so the icon can be linked to another `FrameworkElement` to automatically adapt it's height. This is useful to automatically size an icon which belongs to some text.
As @Keboo mentioned in the linked issue, there are some trade-offs to be made, which I don't think we can fully eliminate.

See MD spec here: https://m3.material.io/styles/icons/applying-icons (_Using Material Symbols with typography_)
image

The icons `Height` and `Width` are bound to the referenced FrameworkElement's `ActualHeight` to keep it rectangular.

Example usage:
```xaml
  <materialDesign:PackIcon Kind="Person"
                           MatchSizeWith="{Binding ElementName=tbTitle}" />
  <TextBlock x:Name="tbTitle"
             Style="{StaticResource MaterialDesignHeadline2TextBlock}"
             Text="Edit Person" />
</StackPanel>
```

If we actually want something like this, and it merges, I will add a follow-up PR for more documentation (incl. MD3 example).


mzy666888:
good work

Keboo:
I like the change but perhaps a different name for the DP. Thoughts on `ScaleToSizeOf`?

Keboo (approved):

corvinsz:
Sounds good. I renamed the property from `MatchSizeWith` to `ScaleToSizeOf` and also adjusted the tests name.

PR #4050, opened 2026-05-13
Position popups below input fields

corvinsz:
Is there any reason the `Placement` is set to `Custom` if it blocks visibility of the selected time/date?
image

MD2 specs:
image


Keboo (approved):

Keboo:
I think I fixed this in my own PR cleaning up the callback as well.  #4052

PR #4054, opened 2026-05-26
Support `CornerRadius` in `Clock` control

corvinsz:
The `Clock` exposes a `CornerRadius` property but does not actually use it in it's template.

The converter `FixedBottomCornerRadiusConverter` is there to apply the `CornerRadius` to only the top left & top right corners of the "header":
image

### DIFF
--- a/src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Clock.xaml
+++ b/src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Clock.xaml
@@ -3,10 +3,7 @@
                     xmlns:converters="clr-namespace:MaterialDesignThemes.Wpf.Converters"
                     xmlns:wpf="clr-namespace:MaterialDesignThemes.Wpf">
 
-  <converters:CornerRadiusCloneConverter x:Key="FixedBottomCornerRadiusConverter"
-                                         FixedBottomLeft="0"
-                                         FixedBottomRight="0" />
-
+  
   <Style x:Key="MaterialDesignClockItemThumb" TargetType="{x:Type Thumb}">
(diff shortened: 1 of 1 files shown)


src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Clock.xaml
```diff
@@ -3,6 +3,10 @@
                     xmlns:converters="clr-namespace:MaterialDesignThemes.Wpf.Converters"
                     xmlns:wpf="clr-namespace:MaterialDesignThemes.Wpf">
 
+  <converters:CornerRadiusCloneConverter x:Key="FixedBottomCornerRadiusConverter"
```
Keboo: Lets nest this down inside of the Style resources rather than at the top of this ResourceDictionary. This ends up exposing it in consumer's XAML when they reference this dictionary.

nicolaihenriksen (approved):
LGTM. Perhaps consider an explicit default value for the CornerRadius in the (base) style? If I am not mistaken, the old code would use "2", where the new code will use "0" unless you explicitly set it to something else?

corvinsz:
@nicolaihenriksen you are indeed correct, thanks for the heads up.
For some reasons a few styles already had a `CornerRadius` of 8 applied to them. I just left them as is.

src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.Clock.xaml
corvinsz: Done

Keboo (approved):

PR #4072, opened 2026-06-17
Add Segmented list box style

corvinsz:
adds a new style `SegmentedListBox` to have this "switch" style selection of n items:
image

### DIFF
--- a/src/MainDemo.Wpf/Toggles.xaml
+++ b/src/MainDemo.Wpf/Toggles.xaml
@@ -452,6 +452,65 @@
       </ListBox>
     </smtx:XamlDisplay>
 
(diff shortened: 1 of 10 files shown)


Keboo:
Push some changes to it. Overall looks good and I think I am good merging it.

One thought I had was on this style not supporting scrolling/virtualization. But I am not sure if anyone would use it and expect that behavior.

corvinsz:
I considered scrolling and virtualization, but intentionally left them out (maybe I should have mentioned this). I think this style is mainly suited for a small number of options rather than large datasets. We can always add support for them later if a need arises.

Your changes LGTM, happy to merge.

PR #4081, opened 2026-08-13
Allow triple left click on a  TextBox to select all text

JLdgu:
Triple clicking in a textbox to select all text is a common action on Windows applications

### DIFF
--- a/src/MaterialDesignThemes.Wpf/TextFieldAssist.cs
+++ b/src/MaterialDesignThemes.Wpf/TextFieldAssist.cs
@@ -183,7 +183,7 @@ public static class TextFieldAssist
         object sender,
         MouseButtonEventArgs e)
     {
-        if (e.ClickCount == 3 && sender is TextBox textBox)
+        if (e.ClickCount == 3 && sender is TextBoxBase textBox)
         {
             textBox.SelectAll();
             e.Handled = true;
--- a/src/MaterialDesignThemes.Wpf/TextFieldAssist.cs
+++ b/src/MaterialDesignThemes.Wpf/TextFieldAssist.cs
@@ -151,6 +151,45 @@ public static class TextFieldAssist
     public static bool GetIncludeSpellingSuggestions(TextBoxBase element) => (bool)element.GetValue(IncludeSpellingSuggestionsProperty);
 
     /// <summary>
+    /// Allow triple click to select all text in a TextBox.
+    /// </summary>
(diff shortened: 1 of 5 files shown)


src/MaterialDesignThemes.Wpf/TextFieldAssist.cs
```diff
@@ -150,6 +150,45 @@ public static class TextFieldAssist
...
+        (bool)obj.GetValue(SelectAllOnTripleClickProperty);
+
+    public static void SetSelectAllOnTripleClick(DependencyObject obj, bool value) =>
+        obj.SetValue(SelectAllOnTripleClickProperty, value);
+
+    private static void OnSelectAllOnTripleClickChanged(
+        DependencyObject d,
+        DependencyPropertyChangedEventArgs e)
+    {
+        if (d is TextBox textBox)
```
corvinsz:
```suggestion
        if (d is TextBoxBase textBox)
```

src/MaterialDesignThemes.Wpf/TextFieldAssist.cs
```diff
@@ -150,6 +150,45 @@ public static class TextFieldAssist
...
+            else
+                textBox.PreviewMouseLeftButtonDown -= TextBox_PreviewMouseLeftButtonDown;
+        }
+    }
+
+    private static void TextBox_PreviewMouseLeftButtonDown(
+        object sender,
+        MouseButtonEventArgs e)
+    {
+        if (e.ClickCount == 3 && sender is TextBox textBox)
```
corvinsz:
```suggestion
        if (e.ClickCount == 3 && sender is TextBoxBase textBox)
```

corvinsz (changes requested):
I few things I noticed:
1. We should type check against `TextBoxBase` so this behavior also works for the `RichTextBox`
2. Maybe an example dedicated to this attached property with a small description would be helpful, especially because this behavior is not something consumers can "just see". Please don't forget the MD3 demo app 😄 
3. In my opinion the test you added should also test the negative (`materialDesign:TextFieldAssist.SelectAllOnTripleClick="False"`) behavior to prevent any regressions.

corvinsz (approved):
LGTM!

The tests could be made data-driven, but I’m totally fine with the current approach as well.

Keboo (approved):
This is great, thank you! For some reason I thought this was already the default. I tweaked it slightly so that it will be on by default for anyone using the library but let people opt out with the AP.

PR #4083, opened 2026-08-21
Allow for waiting for DialogHost animations.

Keboo:
Potential Fix for #4017

Introduces `WaitForOpened` and `WaitForClosed` methods on `DialogHost` to enable asynchronous waiting for the completion of the dialog's visual state transitions. This provides more precise control for executing logic precisely after the dialog has fully opened or closed, rather than just when events are dispatched.

Outstanding items:
1. Are the WaitFor* methods named properly?
2. Tests, need to have them. Will be impossible to test race conditions reliably, but happy path should be there.
3. Should the CancelationToken int he WaitFor* methods be required rather than optional? 

Updates the `DialogSession` to expose the parent `DialogHost` instance, facilitating access to these new awaitable methods.

### DIFF
--- a/src/MaterialDesignThemes.Wpf/DialogHost.cs
+++ b/src/MaterialDesignThemes.Wpf/DialogHost.cs
@@ -1,9 +1,12 @@
+using System.Diagnostics;
 using System.Runtime.InteropServices;
 using System.Security;
+using System.Threading;
 using System.Windows.Data;
 using System.Windows.Interop;
 using System.Windows.Media;
 using System.Windows.Threading;
+using MaterialDesignThemes.Wpf.Internal;
 
 namespace MaterialDesignThemes.Wpf;
 
@@ -31,13 +34,18 @@ public enum DialogHostOpenDialogCommandDataContextSource
 [TemplatePart(Name = PopupPartName, Type = typeof(Popup))]
 [TemplatePart(Name = PopupPartName, Type = typeof(ContentControl))]
 [TemplatePart(Name = ContentCoverGridName, Type = typeof(Grid))]
-[TemplateVisualState(GroupName = "PopupStates", Name = OpenStateName)]
-[TemplateVisualState(GroupName = "PopupStates", Name = ClosedStateName)]
+[TemplatePart(Name = RootContentPartName, Type = typeof(FrameworkElement))]
+[TemplateVisualState(GroupName = VisualStateGroupName, Name = OpenStateName)]
+[TemplateVisualState(GroupName = VisualStateGroupName, Name = ClosedStateName)]
 public class DialogHost : ContentControl
 {
+    public const string VisualStateGroupName = "PopupStates";
+
     public const string PopupPartName = "PART_Popup";
     public const string PopupContentPartName = "PART_PopupContentElement";
     public const string ContentCoverGridName = "PART_ContentCoverGrid";
+    public const string RootContentPartName = "PART_DialogHostRoot";
+
     public const string OpenStateName = "Open";
     public const string ClosedStateName = "Closed";
 
@@ -56,6 +64,9 @@ public class DialogHost : ContentControl
     private DialogClosingEventHandler? _asyncShowClosingEventHandler;
     private DialogClosedEventHandler? _asyncShowClosedEventHandler;
     private TaskCompletionSource<object?>? _dialogTaskCompletionSource;
+
+    private VisualStateMonitor? _visualStateMonitor;
+
 
     private Popup? _popup;
     private ContentControl? _popupContentControl;
@@ -395,7 +406,8 @@ public class DialogHost : ContentControl
 
         //https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/187
         //totally not happy about this, but on immediate validation we can get some weird looking stuff...give WPF a kick to refresh...
-        Task.Delay(300).ContinueWith(t => dialogHost.Dispatcher.BeginInvoke(DispatcherPriority.Background, new Action(() => {
+        Task.Delay(300).ContinueWith(t => dialogHost.Dispatcher.BeginInvoke(DispatcherPriority.Background, new Action(() =>
+        {
             CommandManager.InvalidateRequerySuggested();
             //Delay focusing the popup until after the animation has some time, Issue #2912
             UIElement? child = dialogHost.FocusPopup();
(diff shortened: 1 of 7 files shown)


src/MaterialDesignThemes.Wpf/Internal/VisualStateMonitor.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        _visualStateGroup.CurrentStateChanged += stateChanged;
+
+        currentState = _visualStateGroup.CurrentState.Name;
+        if (currentState == state)
+        {
+            _visualStateGroup.CurrentStateChanged -= stateChanged;
+
+            return Task.CompletedTask;
+        }
```
nicolaihenriksen, line 41: Don't you already have this guard in line 17-18? Is there a scenario where the current state can change between line 18 and here?

src/MaterialDesignThemes.Wpf/DialogHost.cs
```diff
@@ -628,9 +640,26 @@ public override void OnApplyTemplate()
...
+            if (stateNames.Contains(OpenStateName) && stateNames.Contains(ClosedStateName))
+            {
+                _visualStateMonitor = new(stateGroup);
+            }
+        }
         base.OnApplyTemplate();
     }
 
+    public Task WaitForOpened(CancellationToken cancellationToken = default)
+        => _visualStateMonitor?.WaitForState(OpenStateName, cancellationToken)
```
nicolaihenriksen, line 689: Regarding whether cancellationToken should be required: How about setting a reasonable self-cancelling token (say 5 seconds) if one is not supplied by the caller? I guess what you're trying to avoid by making it mandatory is the infinite hang waiting for a state change that never occurs? A reasonable default token would do that too I guess.

My assumption here is that the caller would only call these methods right before opening/closing the `DialogHost` and therefore the timeout should just be slightly longer than the duration of the animation. I hope my assumption is correct.

nicolaihenriksen (approved):
Only some minor comments. Looks good to me though.

src/MaterialDesignThemes.Wpf/Internal/VisualStateMonitor.cs
nicolaihenriksen: After watching your latest stream I now understand the multi-threading concern 👍 

My head hurts a little bit when trying to mentally understand if this actually removes the race-condition. My gut feeling says that you have indeed decreased the "window of opportunity" for the race-condition, but I am unsure that it is completely removed. If you're certain that it removes it, I would love to hear the reasoning as I am failing to convince myself 😆

src/MaterialDesignThemes.Wpf/Internal/VisualStateMonitor.cs
```diff
@@ -0,0 +1,45 @@
...
+
+    public Task WaitForState(string state, CancellationToken cancellationToken)
+    {
+        string currentState = _visualStateGroup.CurrentState.Name;
+        if (currentState == state) return Task.CompletedTask;
+
+        TaskCompletionSource<string> tcs = new();
+        cancellationToken.Register(() => tcs.TrySetCanceled());
+
+        EventHandler<VisualStateChangedEventArgs> stateChanged = null!;
```
corvinsz:
```suggestion
EventHandler<VisualStateChangedEventArgs> stateChanged = null!;
cancellationToken.Register(() =>
{
    _visualStateGroup.CurrentStateChanged -= stateChanged;
    tcs.TrySetCanceled(cancellationToken);
});
```

1. When cancelling via the `CancellationToken` we should unsubscribe the from the event.
2. Pass the `CancellationToken` to the `TrySetCanceled()` method.

corvinsz (commented):
Looks great already, good idea.

> 1. Are the WaitFor* methods named properly?

LGTM
> 2. Tests, need to have them. Will be impossible to test race conditions reliably, but happy path should be there.

Agree

> 3. Should the CancelationToken int he WaitFor* methods be required rather than optional?

Besides what @nicolaihenriksen already mentioned (using a default timeout if none is supplied), what about adding an overload that takes a `TimeSpan`?
```csharp
public Task WaitForClosed(TimeSpan timeout)
{
    using CancellationTokenSource cts = new(timeout);
    return WaitForClosed(cts.Token);
}
```

4. ~Shouldn't the WaitFor* methods be static, or am I misunderstanding the design of the public API?~
Edit: 
I overlooked your example and now understand how you intended the API: register for `DialogClosingEventHandler` and call `await eventArgs.Session.DialogHost.WaitForClosed();` in the handler.
To improve discoverability (and maybe developer experience?), thoughts on adding static methods for WaitFor* like so?
```csharp
public static Task WaitForOpened(object? dialogIdentifier = null, CancellationToken cancellationToken = default)
{
    var instance = GetInstance(dialogIdentifier);
    return instance.WaitForOpened(cancellationToken);
}

public static Task WaitForClosed(object? dialogIdentifier = null, CancellationToken cancellationToken = default)
{
    var instance = GetInstance(dialogIdentifier);
    return instance.WaitForClosed(cancellationToken);
}

// usage:
await DialogHost.Show(...);
await DialogHost.WaitForClosed();
```

src/MaterialDesignThemes.Wpf/DialogHost.cs
```diff
@@ -628,9 +640,26 @@ public override void OnApplyTemplate()
 
         VisualStateManager.GoToState(this, GetStateName(), false);
 
+        if (GetTemplateChild(RootContentPartName) is FrameworkElement root &&
+            VisualStateManager.GetVisualStateGroups(root) is [VisualStateGroup stateGroup, ..])
+        {
+            var stateNames = stateGroup.States.OfType<VisualStateGroup>().Select(x => x.Name).ToList();
```
corvinsz:
```suggestion
            var stateNames = stateGroup.States.OfType<VisualState>().Select(x => x.Name).ToList();
```

src/MaterialDesignThemes.Wpf/DialogHost.cs
Keboo: Mostly agree, though this could easily be called before a DialogHost.Show so I am not sure there is a reasonable timeout.

src/MaterialDesignThemes.Wpf/Internal/VisualStateMonitor.cs
Keboo: 1. Yes, great catch.
2. Done

src/MaterialDesignThemes.Wpf/Internal/VisualStateMonitor.cs
Keboo: I _think_ it is good.... probably.

PR #4085, opened 2026-08-24
I aimed for .cs files. Not enough experienced for .xaml files.

Markot4:
Two things I have to write before I specify what I did:

1. I apologize if I broke any rules when it comes to contribution.

2. I did use GitHub CoPilot in certain cases since I do not want to mess up the source code even though I forked it.

Anyways: here is what I did:

1. I fixed the invalid cast error in src/MahMaterialDragablzMashUp/DialogsViewModel.cs:

* Replaced the direct cast ((MainWindow)Application.Current.MainWindow) with a safe lookup:
 * var leftFlyout = Application.Current?.MainWindow?.FindName("LeftFlyout") as Flyout;
 * Toggle leftFlyout.IsOpen if found.
 * This avoids the compile-time CS0030 error (Window -> MainWindow) and works regardless of XAML-generated partial class details.

I find it funny how this removed 4 errors and many warnings.

2. I made the two changes to address the warnings in DialogsViewModel.cs:
 * Marked DialogDictionary as readonly.
 * Changed ProgressDialog from async void to async Task.

3. I updated the property setter in MahVieModel.cs to throw an ArgumentOutOfRangeException instead of System.Exception, which static analyzers (S112) flag as a poor practice. Property setters should signal invalid argument values with a more specific exception type.

I replaced throw new Exception("Value must be positive") with throw new System.ArgumentOutOfRangeException(nameof(UpDownValue), value, "Value must be non-negative"), which provides a specific exception type, the offending parameter name, and the actual value.

4. Since there weren't any other .cs files, I decided to edit DialogsViewModel.cs:

* Changed InputDialog signature to: private async Task InputDialog() and used await DialogCoordinator.Instance.ShowInputAsync(...).

* Changed the command registrations in the constructor to async lambdas that await the methods:
 * ShowInputDialogCommand = new AnotherCommandImplementation(async _ => await InputDialog());
 * ShowProgressDialogCommand = new AnotherCommandImplementation(async _ => await ProgressDialog());

These make the async calls observed (awaited) and remove CS4014 warnings.

I have no intention of editing .xaml files since I have only studied XML, but not quite enough to understand what I can do.
If there is anything else I have to do, make sure to comment and specify eveything else I can fix. Or even add.

Edit: Oh, wait. I forgot I needed to close VS after installing one of NuGet packages... I will have to commit again just in case.

Edit edit: I saw errors for xaml files just to disappear. what?

### DIFF
--- a/src/MahMaterialDragablzMashUp/DialogsViewModel.cs
+++ b/src/MahMaterialDragablzMashUp/DialogsViewModel.cs
@@ -11,18 +11,18 @@ public class DialogsViewModel
 
     public ICommand ShowLeftFlyoutCommand { get; }
 
-    private ResourceDictionary DialogDictionary = new ResourceDictionary() { Source = new Uri("pack://application:,,,/MaterialDesignThemes.MahApps;component/Themes/MaterialDesignTheme.MahApps.Dialogs.xaml") };
+    private readonly ResourceDictionary DialogDictionary = new ResourceDictionary() { Source = new Uri("pack://application:,,,/MaterialDesignThemes.MahApps;component/Themes/MaterialDesignTheme.MahApps.Dialogs.xaml") };
 
     public DialogsViewModel()
     {
-        ShowInputDialogCommand = new AnotherCommandImplementation(_ => InputDialog());
-        ShowProgressDialogCommand = new AnotherCommandImplementation(_ => ProgressDialog());
+        ShowInputDialogCommand = new AnotherCommandImplementation(async _ => await InputDialog());
+        ShowProgressDialogCommand = new AnotherCommandImplementation(async _ => await ProgressDialog());
         ShowLeftFlyoutCommand = new AnotherCommandImplementation(_ => ShowLeftFlyout());
     }
 
     public Flyout? LeftFlyout { get; set; }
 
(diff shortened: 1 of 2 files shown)


corvinsz:
@Markot4 thank you for your contribution.

A more descriptive PR title would be helpful. 😉

Regarding the changes themselves: if the goal is to refactor the `MahAppsDragablzDemo` application (which I'm not sure is actively maintained anymore), it may be worth considering a broader refactor of the entire class and adopting `CommunityToolkit.Mvvm` where appropriate. It's one of the most widely used MVVM frameworks in the .NET ecosystem and would provide a more consistent approach throughout the code.

If your primary motivation is to use dialogs and drawers (flyouts) with data binding support, you may also want to consider using this library (`MaterialDesignInXaml`), which provides built-in support for these scenarios. I'm not particularly familiar with the MahApps controls, as I haven't worked with them extensively.

Oh and also:
> Anyways: here is what I did:
    I fixed the invalid cast error in src/MahMaterialDragablzMashUp/DialogsViewModel.cs:

I'm not sure which error. The app compiles and runs fine for me 🤔

Keboo (approved):
Thank you for this.

PR #4091, opened 2026-09-08
fix(combobox): allow dropdown toggle when clicking prefix or suffix text

### LINKED ISSUE #4090: ComboBox dropdown is not toggled when clicking PrefixText or SuffixText
### Bug explanation

Clicking on the `TextFieldAssist.PrefixText` (or `Suffix`) does not open or close the dropdown of a `ComboBox`.
This is quite annoying for users when using long prefix or suffix texts.

### Version

5.3.2

corvinsz:
fixes #4090

Sets `IsHitTestVisible="False"` on the internal `PrefixTextBlock` and `SuffixTextBlock` so mouse events are handled by the underlying `ComboBox`.

I tested clicking both the Prefix and Suffix in the same test rather than using two separate tests (or making the test data-driven) to avoid the overhead of starting the app multiple times.

@Keboo I'm not 100% sure whether the `Task.Delay(50)` is appropriate here, or if this should instead use a `Wait.For(...)` call. There is a test above that uses the same 50ms delay, so I followed that approach.

### DIFF
--- a/src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.ComboBox.xaml
+++ b/src/MaterialDesignThemes.Wpf/Themes/MaterialDesignTheme.ComboBox.xaml
@@ -350,6 +350,7 @@
                          Grid.Column="1"
                          Margin="0,0,2,0"
                          FontSize="{TemplateBinding FontSize}"
+                         IsHitTestVisible="False"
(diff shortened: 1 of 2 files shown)


Keboo (approved):
LGTM. Delay seems fine here

